PR #17, opened 2018-08-21
Support cacerts rule

Globegitter:
This is done not so cleanly but does not seem much worse than the rest - happy to improve if any further suggestions.


aiuto:
Friendly ping. I don't know enough about this to review it myself.

PR #21, opened 2018-12-21
Support applying patches for deb_packages

scele:


MarkusTeufelberger:
Great idea, please also document the parameters.

scele:
> Great idea, please also document the parameters.

Done, documentation copy-pasted from [here](https://github.com/bazelbuild/bazel/blob/master/tools/build_defs/repo/http.bzl#L252)!

MarkusTeufelberger:
Thanks, I don't have merge rights though - pinging @philwo who as kind enough to merge last time.

Globegitter:
@MarkusTeufelberger if you get merge rights would you be happy to keep on maintaining this repository? I am sure we could arrange that.

MarkusTeufelberger:
A few existing mainline rules should also be migrated to and maintained in this repository (https://docs.bazel.build/versions/master/be/pkg.html) and I doubt that I would be the best person to maintain them in my free time.

Unless there's a group of co-maintainers, I wouldn't want to do this, especially since I currently don't use these rules.

Globegitter:
That is fair, I would in theory be happy to but if anything just have capacity to look at incoming PRs and possibly respond to issues. But no active coding etc.

philwo:
@aehlig Could you give this a quick look when you have time? :)

If it looks good to you, I'm happy to merge!

MarkusTeufelberger:
Documentation in https://github.com/bazelbuild/rules_pkg/blob/master/deb_packages/README.md is still missing btw.

aehlig:
Also, while technically the addition of patches looks fine, I'm a bit surprised. If I understand correctly, the external repository created contains just a bunch of `.deb` files, which is not the easiest format to patch. So I guess, it is about the programatic transformations of that directory, or which use case do you have in mind?

scele:
> Also, while technically the addition of patches looks fine, I'm a bit surprised. If I understand correctly, the external repository created contains just a bunch of .deb files, which is not the easiest format to patch. 

@aehlig, very good point... Looking at our downstream code, we have another patch applied, which adds `build_file / build_file_content` attributes to `deb_packages`, and providing those will cause the deb package to be extracted. In our use cases, we'd provide `cc_library` or `py_library` targets to expose the (prebuilt) contents of the deb file to the bazel build.

Currently rules_pkg focuses on providing deb packages just for the purposes of assembling docker images, not for extracting the debs and consuming the contents. Would you be interested in contributions also for the latter use case? If not, I don't think this PR makes sense alone, so I could just close it.

aiuto:
Hello all. I'm coming in late here. Should I (or @aehlig) merge this or do we NOT merge based on the last comment.

scele:
Closing.

Issue #24, opened 2019-05-16
EOF error


Globegitter:
I am running the `bazel run //:update_deb_packages` script which used to work fine until a few months ago and now it is just erroring out with the unhelpful error:

```
2019/05/16 11:04:55 EOF
```

Suspicions that I have: Either the script expects different version of `buildifier` or `buildozer` then I now have installed. Or something changed about our `BUILD.bazel` files that the script can not handle correctly anymore.

Globegitter:
Ok, the error is coming from the line `openpgp.CheckArmoredDetachedSignature(keyring, check, sig)`

PR #25, opened 2019-05-16
Check for 200 status code when downloading file.

Globegitter:
While this does not fix https://github.com/bazelbuild/rules_pkg/issues/24 it at least shows the correct error message.


Globegitter:
This does actually fix #24 - the problem was that one of the URLs got removed so things broke down. With this change it will give a better error message and also correctly loop through all mirrors provided.

PR #26, opened 2019-06-05
Start adding rules to create packages to rules_pkg.

aiuto:
- Add the plan to the README
- Update deb_packages dependencies so it at least builds again.
- Refactor so that deb_packages has its own WORKSPACE

This is a minimal change to allow us to begin moving packaging
rules from Bazel to this repository.


aiuto:
Tracking issue: https://github.com/bazelbuild/bazel/issues/8489

deb_packages/WORKSPACE
```diff
@@ -154,10 +161,18 @@ http_file(
 )
 
 # Docker rules.
-git_repository(
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+#git_repository(
```
MarkusTeufelberger: These comments can be probably removed

README.md
```diff
@@ -26,10 +26,22 @@ bundled software packages from various places.
 
 The initial scope is currently targeting `Debian` style distributions, because the Bazel
 [docker rules](https://github.com/bazelbuild/rules_docker) allow optional `deb` and `tar`
-files to be included in their container images.
+files to be included in their container images.  That work was done in December 2017.
 
-Currently these rules are focused on retrieving already built packages from repositories,
-not creating them as artifacts (e.g. like the Bazel
 [packaging rules](https://docs.bazel.build/versions/master/be/pkg.html)).
```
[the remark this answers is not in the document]
aiuto: done

deb_packages/WORKSPACE
aiuto: done

aiuto:
WORKSPACE files are still a bit of a mess.
For now I think we have to end up with a WORKSPACE for each major slice.
What I am doing with the movement though is to make the WORKSPACE for the *source* of rules_pkg follow a deps.bzl pattern, where it brings in the same deps as would an external user.  I think that will give us better fidelity with their experience.

PR #27, opened 2019-06-06
Initial migration of //tools/build_defs/pkg to rules_pkg repo

aiuto:
This change copies the code from bazel/tools/build_defs/pkg out to the rules_pkg repo.
To make the rules run decoupled from Bazel I did some minimal work.

- Move to ABSL flags.
- Add deps.bzl.
- Fix some docs. There is certainly more to do.
- Fix the tests to use the runfiles library rather than ad hoc methods

Note that the tests are now all in a distinct folder from the BUILD file
which contains the tools needed to create packages. This is an experiment
in packaging techniques for rules. The idea is that most users should be
able to import a "thin" version of a rule set.  That would only include
the files needed to use the rule, but not those needed to test and or
package the rules. That would currently be all the *files* in pkg but
none of the folders.

A "thicker" version might include the tests, but at the cost of perhaps
making your workspace deps resolver bring in more things.

The "thickest" version would include all the code needed to repackage
and redistribute the rule set. That would be using the full source
distribution.

I may abandon this experiment and fold tests back together with the
sources, but doing so will not impact users of the tools, so that
would be transparent and harmless. So please indulge me for now.


aiuto:
Suggest for review.
Check out this repo and bazel side by side
Do diff -r on the various parts. It should mostly be cosmetic.

aiuto:
You are welcome. I have a badge for long CL descriptions.

PR #33, opened 2019-06-07
Quick and dirty rules to help make a release.

aiuto:
- Create the tarball - which is the subset of the repo needed to *use* the rules.
- Smoke test the tarball
- Provide a tool to print the WORKSPACE entry with the correct url and
  sha256, so you can do the release notes.

The code is ugly and has a bunch of TODOs for later. This was good enough to create the packaging for release 0.1.0

### DIFF
--- a/pkg/distro/BUILD
+++ b/pkg/distro/BUILD
@@ -1,0 +1,68 @@
+package(
+    default_visibility = ["//visibility:private"],
+)
+
+load("@rules_pkg//:pkg.bzl", "pkg_tar")
+
+VERSION = "0.1.0"
+
+pkg_tar(
+    name = "rules_pkg-%s" % VERSION,
+    srcs = [
+        "@rules_pkg//:standard_package",
+    ],
+    extension = "tar.gz",
+    mode = "0644",
+    owner = "0.0",
+    package_dir = ".",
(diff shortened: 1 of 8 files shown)


pkg/distro/BUILD
```diff
@@ -0,0 +1,68 @@
...
+VERSION = "0.1.0"
+
+pkg_tar(
+    name = "rules_pkg-%s" % VERSION,
+    srcs = [
+        "@rules_pkg//:standard_package",
+    ],
+    extension = "tar.gz",
+    mode = "0644",
+    owner = "0.0",
```
katre: This is pretty opaque. What does it mean?

aiuto: Is it opaque because you are unfamiliar with RPM packaging or for different reasons?
I would do different comments depending on the reason.

katre:
It's the "0.0" specifically, it not any style of Unix file owner syntax I
am familiar with.

aiuto:
Ah... that is stupidity of the way pkg_tar works. They made the syntax owner.group. That has no precedent and seems to provide no win over a clearer syntax. 
Either way, the comment now says what it does, which should clarify/

PR #34, opened 2019-06-07
Wrapped dependency repository rules with the _maybe macro

apt-itude:
My organization uses Artifactory to proxy/cache all of our build dependencies, and it's hard rule that everything must be sourced this way. To accomplish this with Bazel, we need to be able to override each  repository rule with an equivalent rule that points to Artifactory instead of the upstream source. 

While we could simply omit the `rules_pkg_dependencies()` call and manually override these dependencies, we actually have a tool that will inspect all repo rules and automatically generate the equivalent Artifactory rules. The use of `_maybe` allows that to work.

### DIFF
--- a/pkg/deps.bzl
+++ b/pkg/deps.bzl
@@ -8,23 +8,25 @@ def _maybe(repo, name, **kwargs):
 
 def rules_pkg_dependencies():
     # Needed for helper tools
-    http_archive(
+    _maybe(
+        http_archive,
         name = "abseil_py",
         urls = [
             "https://github.com/abseil/abseil-py/archive/pypi-v0.7.1.tar.gz",
-      ],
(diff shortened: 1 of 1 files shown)


apt-itude:
Gah I always forget to set my CLA-approved email for a new repo

philwo:
Looks fine, I'll leave it to @aiuto to merge this, as he's maintaining the repository.

aiuto:
My bad. I was away for the last 2 weeks and totally missed this.

I made the same change in https://github.com/bazelbuild/rules_pkg/commit/55afe43d430dba09270eeedcf2e0ce3374c02897#diff-b1ec5372e7a6138f4df358e8a2b164d8

So I don't think there is any need for this one.

apt-itude:
No worries! Glad the change got made either way

PR #43, opened 2019-06-14
Federation exploration

aiuto:
"fix" visibility to ":archive". It is a workaround rather than a real fix.
Bazal does not have a way to specify "@<the project that this BUILD or
.bzl file comes from>".


aiuto:
That was an "oops" merge. It should have gone into aiuto/rules_pkg. 
I think the key is to never use the github UI.

PR #44, opened 2019-06-14
Change owners so we spamming everyone while we are in rapid developem…

aiuto:
…ent.

The bazel team members are churning this right now. We can select our
own reviewers.


aiuto:
The old PR may have gotten trashed because .... reasons.

Issue #35, opened 2019-06-10
pkg_deb flag decoding breaks for python3.


aiuto:
From https://github.com/bazelbuild/bazel/issues/7370
@jwnimmer-tri 

If I'm understanding my build errors correctly, this change re-broke the py2/3 compatibility on make_deb.py that #6818 had previously fixed.

```
  File ".../bazel-out/k8-opt/bin/external/bazel_tools/tools/build_defs/pkg/make_deb.runfiles/bazel_tools/tools/build_defs/pkg/make_deb.py", line 301, in GetFlagValue
    flagvalue = flagvalue.decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

It's possible that the toolchain changes in 0.25.0 are also at fault and/or the only problem, but with the binary tagged PY2AND3, the code doesn't look correct to me on first read.

qzmfranklin:
I'd believe that this is identical:
https://github.com/bazelbuild/bazel/issues/8465

Would like to see this fixed.

aiuto:
Yes. We are going to fix it in this repository, not in bazelbuild/bazel.
I am very happy to accept a patch if you want to send one.

qzmfranklin:
Let me do that. We have been seeing this issue multiple times.  Motivated enough to do something.

qzmfranklin:
Though, the BUILD file and the soruce file stipulates that the `make_deb` target be run under python3.

```bzl
py_binary(
    name = "make_deb",
    srcs = ["make_deb.py"],
    python_version = "PY2",
    srcs_version = "PY2AND3",
    visibility = ["//visibility:public"],
    deps = [
        ":archive",
        "@abseil_py//absl/flags",
    ],
)
```

Commenting out the `python_version = "PY2"` line, run the tests got:
```text
13:58 $ bazel test ...
INFO: Analyzed 34 targets (15 packages loaded, 375 targets configured).
INFO: Found 29 targets and 5 test targets...
FAIL: //distro:packaging_test (see /home/zhongming/.cache/bazel/_bazel_zhongming/dcb9e5cca53c9c59e32893c1321b9420/execroot/rules_pkg/bazel-out/k8-fastbuild/testlogs/distro/packaging_test/test.log)
[...]

PR #47, opened 2019-07-03
Update make_deb.py to work for Python2 or 3

aiuto:
Generally being more careful about unicode handling.
Some very weird magic with pre-processing argv to undo undesired interpreter behavior.

This leaves the code in a state where someone could use either interpreter.
Next step it to figure out how to select the python version to match whatever toolchain we have.

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -41,10 +41,13 @@ py_binary(
     ],
 )
 
+
+# NOTE: This works for PY2 and PY3.
+# TODO: Figure out how to use "whatever version" we have on the machine.
 py_binary(
     name = "make_deb",
     srcs = ["make_deb.py"],
-    python_version = "PY3",
+    python_version = "PY2",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -44,7 +44,7 @@ py_binary(
 py_binary(
     name = "make_deb",
     srcs = ["make_deb.py"],
-    python_version = "PY2",
+    python_version = "PY3",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -42,11 +42,22 @@ py_binary(
 )
 
 
-# NOTE: This works for PY2 and PY3.
-# TODO: Figure out how to use "whatever version" we have on the machine.
 py_binary(
     name = "make_deb",
     srcs = ["make_deb.py"],
+    python_version = "PY3",
+    srcs_version = "PY3",
+    visibility = ["//visibility:public"],
+    deps = [
+        ":archive",
+        "@abseil_py//absl/flags",
+    ],
+)
+
+py_binary(
+    name = "make_deb_py2",
+    srcs = ["make_deb.py"],
+    main = "make_deb.py",
     python_version = "PY2",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -46,7 +46,6 @@ py_binary(
     name = "make_deb",
     srcs = ["make_deb.py"],
     python_version = "PY3",
-    srcs_version = "PY3",
     visibility = ["//visibility:public"],
     deps = [
         ":archive",
--- a/pkg/make_deb.py
+++ b/pkg/make_deb.py
@@ -380,4 +380,10 @@ def main(unused_argv):
 if __name__ == '__main__':
   MakeGflags()
   FLAGS = flags.FLAGS
+  # sys.argv is not actually proper str types on Unix with Python3
+  # The bytes of the arg are each directly transcribed to the characters of
+  # the str. It is actually more complex than that, as described in the docs.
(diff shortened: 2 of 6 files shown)


aiuto:
Mostly fixes #35 
@qzmfranklin, @jwnimmer-tri  Do you want to patch this in, change the target to PY3 and see if it works for you?

brandjon:
> Next step it to figure out how to select the python version to match whatever toolchain we have.

#8713. I think the easiest thing is to make it `python_version = "PY3"` since that's preferred, but keep it compatible with PY2 (enforced via a `python_version = "PY2"` `py_test`). In the host config you don't control its real Python version anyway.

If a user doesn't have Python 3 installed but wants to run targets marked PY3 that are in reality compatible with both Python 2 and 3, they should use the non-strict toolchain (implemented in #8547).

pkg/make_deb.py
```diff
@@ -303,15 +303,30 @@ def CreateChanges(output,
           'Checksums-Sha256',
           '\n' + ' '.join([checksums['sha256'], debsize, deb_basename]))
   ])
-  with open(output, 'w') as changes_fh:
+  with open(output, 'wb') as changes_fh:
     changes_fh.write(changesdata.encode('utf-8'))
 
 
 def GetFlagValue(flagvalue, strip=True):
```
brandjon: Nit: The name's a bit vague. Maybe rename to NormalizeUnicode or NormalizeString or something? The docstring should clarify what specific kind of normalization we're talking about.

aiuto: I want to keep the name of GetFlagValue just to limit the number of changed lines.
But I did expand the docstring and substantially re-wrote it to do the fsencoding stuff here, so that we do not (incorrectly) turn a utf-8 path in @filename args to an improper path.

aiuto:
Also changed the default version to py3 and added a py2 test.

pkg/BUILD
```diff
@@ -42,11 +42,22 @@ py_binary(
 )
 
 
-# NOTE: This works for PY2 and PY3.
-# TODO: Figure out how to use "whatever version" we have on the machine.
 py_binary(
     name = "make_deb",
     srcs = ["make_deb.py"],
+    python_version = "PY3",
+    srcs_version = "PY3",
```
brandjon: I'd actually leave out `srcs_version` entirely and let it default to `PY2AND3`. The source code is compatible with both versions, after all. It's only the target that says to use one or the other.

If this were a `py_library` target, `PY2AND3` would be necessary in order for the target to be usable as a dependency in binaries of both versions. It doesn't matter for `py_binary` except for documentation / clarity.

Actually, setting it to the default `PY2AND3` value explicitly could even be helpful, in that it replaces the "This works for both PY2 and PY3" comment. Same principle for the `make_deb_py2` target.

pkg/make_deb.py
```diff
@@ -386,4 +400,5 @@ def main(unused_argv):
   # https://docs.python.org/3/library/sys.html#sys.argv
   # https://docs.python.org/3/library/os.html#os.fsencode
   # https://www.python.org/dev/peps/pep-0383/
-  main(FLAGS([os.fsencode(arg).decode('utf-8') for arg in sys.argv]))
+  # main(FLAGS([os.fsencode(arg).decode('utf-8') for arg in sys.argv]))
```
brandjon: Stale line. Also should the above comment be deleted now that the logic's moved to `GetFlagValue`?

pkg/make_deb.py
```diff
@@ -308,26 +308,40 @@ def CreateChanges(output,
...
+    # assertion: py2: flagvalue is unicode
+    # assertion: py3: flagvalue is str, but in weird format
     if flagvalue[0] == '@':
+      # Subtle: We do not want to re-encode the value here, because it
+      # is encoded in the right format for file open operations.
       with open(flagvalue[1:], 'rb') as f:
         flagvalue = f.read().decode('utf-8')
+    else:
+      # convert fs specific encoding back to proper unicode.
+      flagvalue = os.fsencode(flagvalue).decode('utf-8')
```
brandjon: `fsencode()` will return the original bytes. If we're on Windows, what encoding will those bytes be? If they're not utf-8, doesn't that mean we'll be mis-decoding `flagvalue`?

Another way to put it: `sys.argv` gives us a sequence of unicode code points as specified by the system. On windows this can actually be interesting characters (>255), while on linux this is always characters in range 0-255. This is because the use of utf-8 in linux is a convention, not a built-in feature understood by the OS. Therefore, unconditionally decoding the bytes as utf-8 is (I think?) incorrect on systems that follow a different convention.

pkg/BUILD
aiuto: Done.

aiuto:
> Mostly fixes #35
> @qzmfranklin, @jwnimmer-tri Do you want to patch this in, change the target to PY3 and see if it works for you?

I added changed to PY3 as the default and added a py2 test, so this is less important. I think I will merge and we'll see how that goes.

jwnimmer-tri:
Thanks for your work on this!  Apologies for the slow reply, I have been on vacation.

I tested the latest master here 96e34abe84353387fd378fba48be88836f6c3f2d against my tree, and I can confirm that the `decode()`-related problems I saw in bazelbuild/bazel/issues/7370 are resolved.

PR #57, opened 2019-07-15
Allow supplying package_dir of pkg_tar from a file

vmax:
- This PR allows to provide `package_dir_file` to `pkg_tar` — similarly to how `version`/`version_file` [and others] work in `pkg_deb` rule
- `GetFlagValue` implementation was copy-pasted from `make_rpm.py`. IMO, a good refactor here would be to separate it out to `common.py`, wrap it in `py_library` and put it in `deps` of `make_deb`, `make_rpm` and `build_tar`. Let me know if it's needed — I can submit a follow-up PR.

### DIFF
--- a/pkg/build_tar.py
+++ b/pkg/build_tar.py
@@ -23,6 +23,8 @@ import tempfile
 from rules_pkg import archive
 from absl import flags
 
+from helpers import GetFlagValue
+
 flags.DEFINE_string('output', None, 'The output file, mandatory')
 flags.mark_flag_as_required('output')
 
@@ -88,16 +90,6 @@ flags.DEFINE_string('root_directory', './',
                     'Default root directory is named "."')
 
 FLAGS = flags.FLAGS
-
-
-def GetFlagValue(flagvalue, strip=True):
-    if flagvalue:
-        if flagvalue[0] == '@':
-            with open(flagvalue[1:], 'r') as f:
-                flagvalue = f.read()
-        if strip:
-            return flagvalue.strip()
-    return flagvalue
 
 
 class TarFile(object):
--- a/pkg/build_tar.py
+++ b/pkg/build_tar.py
@@ -90,6 +90,16 @@ flags.DEFINE_string('root_directory', './',
 FLAGS = flags.FLAGS
 
 
+def GetFlagValue(flagvalue, strip=True):
+    if flagvalue:
+        if flagvalue[0] == '@':
+            with open(flagvalue[1:], 'r') as f:
+                flagvalue = f.read()
+        if strip:
+            return flagvalue.strip()
+    return flagvalue
+
+
 class TarFile(object):
   """A class to generates a TAR file."""
 
@@ -335,7 +345,7 @@ def main(unused_argv):
       ids_map[f] = (int(user), int(group))
 
   # Add objects to the tar file
-  with TarFile(FLAGS.output, FLAGS.directory, FLAGS.compression,
+  with TarFile(FLAGS.output, GetFlagValue(FLAGS.directory), FLAGS.compression,
                FLAGS.root_directory, FLAGS.mtime) as output:
 
     def file_attributes(filename):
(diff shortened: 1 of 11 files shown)


pkg/build_tar.py
```diff
@@ -90,6 +90,16 @@
 FLAGS = flags.FLAGS
 
 
+def GetFlagValue(flagvalue, strip=True):
```
aiuto: If  you are going to bring this in, please take the version from make_deb.py, which is PY2 and PY3 compatible. https://github.com/bazelbuild/rules_pkg/blob/96e34abe84353387fd378fba48be88836f6c3f2d/pkg/make_deb.py#L310

aiuto: Maybe extracting this into a library first is the right approach.  The helpers already depend on another library (archive), so structurally it is just another file and an import.

vmax:
@aiuto could I have another round of review, please?

aiuto:
Sorry for the delay. I was fighting a fire this week.

This is great.
Do you promise to follow up with a test for the new flag?

vmax:
@aiuto is the test we need here an integration test that would verify that archives built using `package_dir` and `package_dir_file` are identical?

aiuto:
Sure. That would be sufficient.

vmax:
@aiuto added an integration test here: https://github.com/bazelbuild/rules_pkg/pull/57/commits/cc4a38475db73d307c4197c3b1f4655def61b7c7

pkg/tests/BUILD
```diff
@@ -12,9 +12,35 @@ genrule(
     cmd = "for i in $(OUTS); do echo 1 >$$i; done",
 )
 
+pkg_tar(
+    name = "test-tar-package-dir",
```
aiuto: Please use _ instead of - as a separator to be consistent with the other code.

pkg/tests/BUILD
```diff
@@ -12,9 +12,35 @@ genrule(
...
+    srcs = [
+        ":etc/nsswitch.conf",
+    ],
+    package_dir = "/package"
+)
+
+genrule(
```
aiuto: Why don't you just check in the file  'testdata/package_dir.txt' with the content '/package'.  That will avoid Windows compatibility problems with the genrule.  Then you also do not need to include this in the archive_testdata rule, because the glob will get it.

pkg/tests/archive_test.py
```diff
@@ -339,5 +339,24 @@ def testChangingRootDirectory(self):
     ]
     self.assertTarFileContent(self.tempfile, content)
 
+  def testPackageDirFileAttribute(self):
+      """
```
aiuto: Multiline docstrings should have a one line summary at the top.

https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

aiuto:
@vmax: Would you be interested in being listed as a collaborator for
bazelbuild/rules_pkg? That way I can send some code reviews to you.

pkg/tests/BUILD
vmax: @aiuto good point! addressed in https://github.com/bazelbuild/rules_pkg/pull/57/commits/fa82eea821e9cadd65cc994215e8c7fdb08c5284

pkg/tests/BUILD
vmax: @aiuto good point! addressed in https://github.com/bazelbuild/rules_pkg/pull/57/commits/fa82eea821e9cadd65cc994215e8c7fdb08c5284
I haven't really considered that Windows is incompatible with UNIX `echo`

pkg/tests/archive_test.py
vmax: @aiuto addressed in https://github.com/bazelbuild/rules_pkg/pull/57/commits/fa82eea821e9cadd65cc994215e8c7fdb08c5284

vmax:
> @vmax: Would you be interested in being listed as a collaborator for bazelbuild/rules_pkg? That way I can send some code reviews to you.

@aiuto thanks, this would be an honor for me

PR #62, opened 2019-07-23
Remove outputs list from pkg_deb and pkg_tar

aiuto:
Replace rules with wrapper macros.
Add OutputGroup

Bump version to 0.2.1. Compatible change, no new features.

Required for https://github.com/bazelbuild/bazel/issues/7977

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -15,7 +15,7 @@
 
 load(":path.bzl", "compute_data_path", "dest_path")
 
-version = "0.2.0"
+version = "0.2.1"
 
 # Filetype to restrict inputs
 tar_filetype = [".tar", ".tar.gz", ".tgz", ".tar.xz", ".tar.bz2"]
@@ -128,6 +128,7 @@ def _pkg_tar_impl(ctx):
         },
         use_default_shell_env = True,
     )
+    return OutputGroupInfo(out=[ctx.outputs.out]);
 
 def _pkg_deb_impl(ctx):
     """The implementation for the pkg_deb rule."""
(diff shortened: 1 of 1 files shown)


pkg/pkg.bzl
```diff
@@ -239,9 +240,15 @@ def _pkg_deb_impl(ctx):
...
+    output_groups = {"out": [ctx.outputs.out]}
+    if hasattr(ctx.outputs, "deb"):
+        output_groups["deb"] = [ctx.outputs.deb]
+    if hasattr(ctx.outputs, "changes"):
+        output_groups["changes"] = [ctx.outputs.changes]
+    return OutputGroupInfo(**output_groups)
 
 # A rule for creating a tar file, see README.md
-_real_pkg_tar = rule(
+pkg_tar_impl = rule(
```
vmax: @aiuto previously this was private - do we want to export it?

pkg/pkg.bzl
```diff
@@ -287,10 +295,14 @@ def pkg_tar(**kwargs):
...
-    _real_pkg_tar(**kwargs)
+    extension = kwargs.get("extension") or "tar"
+    pkg_tar_impl(
+        out = kwargs["name"] + "." + extension,
+        **kwargs
+    )
 
 # A rule for creating a deb file, see README.md
-pkg_deb = rule(
+pkg_deb_impl = rule(
```
vmax: @aiuto as we have public `pkg_deb`, should this one be exported too?

aiuto: Sigh.  No. 
The problem is that with the removal of `outputs` you can no longer default an output file name. So you need a macro. You can not use the real implementation directly or you won't set all the mandatory outputs.

Issue #30, opened 2019-06-06
Improve pkg_zip


aiuto:
Some requirements:
- portable to *nix & Windows without forking the system zip
- ability to repath files to where you want them
- ability to set permissions per file
- tests that run on all platforms

j3parker:
Is a python implementation along the lines of `pkg_tar` acceptable?

aiuto:
Yes. We should aim for cross OS portability first, and then we can iterate towards faster.
Are you thinking of taking this on? It would be great to have a design proposal for how to handle re-pathing files and permissions. Ideally, the same mechanism should eventually work for pkg_{tar, deb, rpm, ...} too.

On Monday I can extract the docs for the rules Google uses internally for mapping files into packages into a doc to use as an example of requirements. The actual rules we use have sort of grown into a mashup of competing implementations and ideas, so I would not want to bring them forward directly. The do, however, capture a lot of the corner cases.

j3parker:
Ok. I'd be interesting in taking it on if that's helpful (I saw the "help wanted" tag). Portability is important for me because my company would need it to run on Windows and Linux :) (and I have a Mac).

A design proposal sounds great. Would you like to see that done via https://github.com/bazelbuild/proposals ? Or another way? You mentioned the proposal RE: repathing/permissions. Would the proposal be specific to that or would it be a proposal for `pkg_zip`, including those aspects? Should the proposal consider pkg_{tar, deb,rpm, ...} as well in that context?

I understand what you mean by repathing and permissions at a high level, but any extra docs/context you could provide would be great, especially for corner cases. As an outsider, nothing comes to mind ;)

Something personally I'd want to nail down early is the contract for the rule in detail. That could be done in a design proposal and/or in a small PR (with `pkg_zip = rule( ... )`  (basically, the attrs) + an impl that is just `# TODO` to a branch. If you like getting things in `master` then maybe it could initially be `_pkg_zip`). I could also paste a gist to this issue instead.

More generally I'd be interested in doing it in a sequence of smaller PRs (either into master or a branch). The benefits to this would be it'd be easier for someone else to pick up the work if I couldn't finish it and (I find) it's easier to review. If you'd rather have one PR with the full implementation (after a design proposal), though, that's fine too.

aiuto:
On Sat, Jul 20, 2019 at 12:07 PM Jacob Parker <notifications@github.com>
wrote:

> Ok. I'd be interesting in taking it on if that's helpful (I saw the "help
[...]

Issue #49, opened 2019-07-09
replace absl.flags with argparse in the tools (make_dep.py, make_rpm.py, ...)


aiuto:
That removes a dependency on abseil and six. 
It will also make the code easier for most developers to understand.

PR #65, opened 2019-07-27
Add a basic `pkg_zip` rule

j3parker:
This is a basic implementation of `pkg_zip` for #30 without path mapping.


pkg/BUILD
```diff
@@ -42,6 +42,15 @@ py_binary(
     ],
 )
 
+py_binary(
+    name = "build_zip",
+    srcs = ["build_zip.py"],
+    visibility = ["//visibility:public"],
+    deps = [
+        ":archive",
+        "@abseil_py//absl/flags",
```
j3parker: I saw in #49 that you want to get rid of this dependency. I wrote this mirroring the `pkg_tar` code. Would you prefer it using abseil for now or immediately going to `optparse`?

pkg/README.md
```diff
@@ -333,6 +333,74 @@ Creates a tar file from a list of inputs.
...
+      <td>
+        <code>String, default to 'zip'</code>
+        <p>
+            The extension for the resulting zipfile. The output
+            file will be '<i>name</i>.<i>extension</i>'.
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>strip_prefix</code></td>
```
j3parker: I wasn't sure if we wanted this attr on this rule. We talked about opening a proposal for the more flexible `pkg_directory`/etc. rules in #30 and those would cover this functionality. Is it still useful to have it on the root of the rule for simple cases?

pkg/build_tar.py
```diff
@@ -23,7 +23,7 @@
 from rules_pkg import archive
 from absl import flags
 
-from helpers import GetFlagValue
+from helpers import GetFlagValue, UnquoteAndSplit
```
j3parker: I moved this function out of `build_tar.py` to `helpers.py` so it could be used by `build_zip.py` as well. I renamed it to match the other function in that library, let me know if that doesn't make sense.

pkg/build_zip.py
```diff
@@ -0,0 +1,57 @@
...
+def main(unused_argv):
+  with ZipFile(FLAGS.output, 'w') as zip:
+    for f in FLAGS.file:
+      (src_path, dst_path) = UnquoteAndSplit(f, '=')
+
+      # remove leading /'s: the zip format spec says paths should never have a
+      # leading slash, but Python will happily do this. The built-in zip tool
+      # in Windows will complain that such a zip file is invalid.
+      if dst_path[0] == '/':
+          dst_path = dst_path[1:]
```
j3parker: Maybe this is weird to have in `build_zip.py`. I'm open to alternatives. My thought was that (1) its definitely a compatibility problem to have leading slashes, but (2) thats a zip specific thing, and (3) doing it in Starlark would make this binary less flexible/would either be the same hack or more work (not sharing as much code with `pkg_tar`.

pkg/build_zip.py
```diff
@@ -0,0 +1,57 @@
...
+from zipfile import ZipFile, ZipInfo, ZIP_DEFLATED
+
+flags.DEFINE_string('output', None, 'The output file, mandatory')
+flags.mark_flag_as_required('output')
+
+flags.DEFINE_multi_string('file', [], 'A file to add to the zip')
+
+FLAGS = flags.FLAGS
+
+def main(unused_argv):
```
j3parker: I didn't go all out making this a class like `pkg_tar` (I assume it's like that for testing/reusing the impl for other `pkg_*` formats?) let me know if you'd like to see anything like this.

pkg/build_zip.py
```diff
@@ -0,0 +1,57 @@
...
+      # leading slash, but Python will happily do this. The built-in zip tool
+      # in Windows will complain that such a zip file is invalid.
+      if dst_path[0] == '/':
+          dst_path = dst_path[1:]
+
+      entry_info = ZipInfo(
+        filename = dst_path,
+        # choose something deterministic.
+        # this is the earliest date possible in the basic zip file format.
+        date_time = (1980, 1, 1, 0, 0, 0),
```
j3parker: Do you want this configurable, like `pkg_tar`?

pkg/build_zip.py
```diff
@@ -0,0 +1,57 @@
...
+      )
+
+      entry_info.compress_type = ZIP_DEFLATED
```
j3parker: The docs say not every distribution of Python has to support this, but every one I tested did. The other options are [`ZIP_STORED`](https://docs.python.org/3/library/zipfile.html#zipfile.ZIP_STORED) (no compression) and two other compression formats which are apparently not well-supported by tools.

I didn't make an attr for compression level (ala #60). Do you want it in this PR or a future one?

pkg/tests/build_test.sh
```diff
@@ -55,6 +55,15 @@ function get_tar_permission() {
   tar tvf "${test_data}" | fgrep "00 $file" | cut -d " " -f 1
 }
 
+function get_zip_listing() {
+  local input=$1
+  local test_data="${TEST_DATA_DIR}/${input}"
+  # Remove table headers, summary and filter to only the columns we care about
+  unzip -v "${test_data}" 2>/dev/null \
+    | awk 'NR>3 && $1 == "--------" {exit} NR>3 {print $1, $2, $5, $6, $7, $8}' \
+    | sed 's:01-01-1980:1980-01-01:g' # Normalize dates
```
j3parker: I'm open to alternatives :) My Mac did this differently from my Windows/Linux builds. I haven't investigated it. 🤷‍♂

j3parker:
Ok, so manually testing this on Windows, Mac, and Linux - it all works. The automated tests run and work on Linux and Mac.

A few tests from the master branch fail on Mac (I haven't investigated but will + will open an issue). A lot more fail on Windows because the `genrule` that generates test files fails (I think due to path separator stuff? Not sure yet).

What I found would work is replacing the `genrule` with `write_file` rules (from skylib). Does that sound cool? I didn't add that in this PR as its a bit noisy but I could open up another PR?

pkg/tests/build_test.sh
vmax: @j3parker consider using `zipinfo` with `-T` switch; from `zipinfo`'s help: `-T  print file times in sortable decimal format`. I've checked in on macOS and Linux and it produces consistent datetime (`19800101.000000`)

aiuto: My preference is no more shell tests.

This could be a distinct python test that used python's zlib apis to unzip and see that the archive structure matched what you expect.

pkg/README.md
aiuto: Let's try leaving it out for now and adding as requested.

pkg/build_tar.py
aiuto: That makes sense.
Could you do this in a distinct refactoring PR, so we can start adding unit tests for it.

It should also have a meaningful name. unquote_and_split its not what it really does. That implies we unquote args.  It is simply splitting on the separator and allowing \= in the string. It is also a mess of unexpected semantics. It looks like   na\xffme=value would result in 'nax25me' value, instead of the expected 'na%me', 'value' or the (also correct) 'na\xffme', 'value'

How about   SplitNameValuePairAtSeparator(arg, sep='=')

pkg/README.md
```diff
@@ -333,6 +333,74 @@ Creates a tar file from a list of inputs.
...
+    <tr>
+      <td><code>srcs</code></td>
+      <td>
+        <code>List of files, optional</code>
+        <p>File to add to the layer.</p>
+        <p>
+          A list of files that should be included in the archive.
+        </p>
+      </td>
+    </tr>
```
aiuto: This needs package_dir.

pkg/build_zip.py
aiuto: Yes. Otherwise you'll have epoch problems for different systems.

pkg/pkg.bzl
```diff
@@ -379,3 +379,50 @@ def pkg_deb(name, package, **kwargs):
...
+        "srcs": attr.label_list(allow_files = True),
+        # Implicit dependencies.
+        "build_zip": attr.label(
+            default = Label("@rules_pkg//:build_zip"),
+            cfg = "host",
+            executable = True,
+            allow_files = True,
+        ),
+    },
+    outputs = {
```
aiuto: Outputs is going away. This needs to have an explicit attr.output() and return an OutputGroupInfo. See https://github.com/bazelbuild/rules_pkg/commit/30dd90482fd2387d9641eaca37b5c303b2e3fb36
for an example of the change

pkg/BUILD
aiuto: Go to optparse. It means less change in the future.

pkg/build_zip.py
aiuto: We can do this in a followup. There is a lot of other scaffolding to do first.

pkg/build_zip.py
```diff
@@ -0,0 +1,57 @@
...
+
+      # Unfortunately, the only way to choose the files date with zipfile is
```
aiuto: There is no need for the comment. Why you think about it, we must read the file into this processes memory at some point so we can compress it. So, nothing is 'unfortunate'.

j3parker: Ok, I was guessing that the other method (the one that takes a path) went through native C code and that might ultimately be better but that's total speculation (I don't know much about Python, didn't check and definitely didn't measure). I'll remove the comment!

pkg/build_tar.py
aiuto: Thanks for pulling that out.

j3parker: NP! I'm going to rebase this and incorporate the feedback soon (lots of great feedback, thanks)

j3parker:
One thing I forgot to ask about: `pkg_tar` has a `deps` attr which can point to other tar files that will be extracted and added to it.

Do we want something like that for `pkg_zip`? A follow-up/if someone asks?

pkg/build_zip.py
aiuto: We will probably be there at some point. But you can figure out the API when needed rather than desigining in a vacuum.

pkg/build_zip.py
aiuto: This seems reasonable for now.

j3parker:
I'll have something up by tomorrow night, sorry for the radio silence!

aiuto:
No problem at all. Thank you for taking on the project. It's just that I
had to make a new release and was wondering if I could include that. No
worries. There will be point releases St least once a week for the next few
months. There is always another bus.

pkg/BUILD
j3parker: Done!

pkg/README.md
j3parker: Done

pkg/build_zip.py
j3parker: I added a `timestamp` attr with some logic. It doesn't match `pkg_tar` exactly (which has `mtime`/`portable_mtime` attrs) let me know if you want to see something else here.

pkg/README.md
j3parker: I removed it

pkg/tests/build_test.sh
j3parker: I've removed the shell tests

j3parker:
I owe you some tests. I'm going to do them in Python; I'll push them tomorrow.

pkg/pkg.bzl
j3parker: Opps I forgot incorporate all the changes into that last PR, one sec.

j3parker: There we go

pkg/build_zip.py
```diff
@@ -0,0 +1,112 @@
...
+
+from datetime import datetime
+from helpers import SplitNameValuePairAtSeparator
+from optparse import OptionParser
+from zipfile import ZipFile, ZipInfo, ZIP_DEFLATED
+
+ZIP_EPOCH = 315532800
+
+def get_options_parser():
+  parser = OptionParser()
```
aiuto: parser = argparse.ArgumentParser(fromfile_prefix_chars='@')

You fromfile_prefix_chars='@' for args.use_param_file to work in the pkg_zip rule.
https://docs.python.org/2/library/argparse.html#fromfile-prefix-chars

pkg/build_zip.py
```diff
@@ -0,0 +1,112 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""This tool builds zip files from a list of inputs."""
+
+from datetime import datetime
+from helpers import SplitNameValuePairAtSeparator
+from optparse import OptionParser
```
aiuto: Please use argparse. optparse is deprecated.

pkg/pkg.bzl
```diff
@@ -379,3 +379,57 @@ def pkg_deb(name, package, **kwargs):
...
+def _pkg_zip_impl(ctx):
+    args = ctx.actions.args()
+
+    for f in ctx.files.srcs:
+        args.add("-f", "%s=%s" % (_quote(f.path), dest_path(f, strip_prefix = None)))
+
+    args.add("-o", ctx.outputs.out.path)
+    args.add("-d", ctx.attr.package_dir)
+    args.add("-t", ctx.attr.timestamp)
+    args.use_param_file("%s")
```
aiuto: As written, this does not work if there are a large number of files. If it does spill args to a file build_zip will get invoked with just the name of the temp file. And then it will be sad....  

If we were using absl flags, then typically the format would be 
   args.use_param_file("--flagfile=%s")
because absl flags treats flagfile specially (see: https://abseil.io/docs/python/guides/flags)

But argparse does not do this magically, it uses fromfile_prefix_chars to do indirection or args to a file. So
     args.set_param_file_format("multiline")
     args.use_param_file("@%s")
FWIW, use_param_file is underdocumented. It needs an example of how it must actually be used.

j3parker: Yeoch, sorry. I forgot I had configured this!

pkg/pkg.bzl
```diff
@@ -379,3 +379,64 @@ def pkg_deb(name, package, **kwargs):
...
+    args.add("-d", ctx.attr.package_dir)
+    args.add("-t", ctx.attr.timestamp)
+
+    args.add_all(
+        ctx.files.srcs,
+        map_each = _format_zip_file_arg,
+    )
+
+    args.set_param_file_format("multiline")
+    args.use_param_file("@%s")
```
j3parker: I manually tested this (on Linux/Mac/Windows) by setting `use_always = True` and running the tests...
Do you want to see a test for this? Maybe we could generate a really big zip file. Or maybe use this undocumented option: https://github.com/bazelbuild/bazel/blob/3a879281507497dfbad1dda0051368a36f6e0e2b/src/main/java/com/google/devtools/build/lib/analysis/config/CoreOptions.java#L119

pkg/tests/BUILD
```diff
@@ -28,6 +28,82 @@ pkg_tar(
...
+    srcs = [
+        "zip_test.py",
+    ],
+    data = [
+        "test_zip_empty.zip",
+        "test_zip_basic.zip",
+        "test_zip_package_dir0.zip",
+        "test_zip_timestamp.zip",
+
+        # these could be replaced with diff_test() rules (from skylib)
```
j3parker: I didn't want to try bringing skylib in inside this PR (especially since it seems like you're up to a lot of stuff around pacakging/federation/bringing this into Bazel core etc.) Let me know what you think.

pkg/tests/zip_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+HELLO_CRC = 2069210904
+LOREM_CRC = 2178844372
+
+class ZipTest(unittest.TestCase):
+    def get_test_zip(self, zipfile):
+        """Get the file path to a generated zip in the runfiles."""
+
+        return self.data_files.Rlocation(
+            "rules_pkg/tests/" + zipfile
```
j3parker: The archive tests used `os.join` which was admirable but didn't work out well on Windows: we really just want forward slashes here.

pkg/pkg.bzl
aiuto: We can add that later.

pkg/tests/BUILD
aiuto: I am not sure diff_test is right either. What you want is a test like this

zip_content_test(
    name = "test_zip_package_dir0_test"
    zip_file = "test_zip_package_dir0.zip"
    expected = [
          "rwxr-xr-x foo",
           "...."
    ]
)

Rather than diffing the content directly, diff the semantic meaning of the content.
But that is worth doing for all the packaging things at once. If we can get a stanard comparison format, the same diff and reporting tool could be used for any of them.

aiuto:
> Ok, so manually testing this on Windows, Mac, and Linux - it all works. The automated tests run and work on Linux and Mac.
> 
> A few tests from the master branch fail on Mac (I haven't investigated but will + will open an issue). A lot more fail on Windows because the `genrule` that generates test files fails (I think due to path separator stuff? Not sure yet).
 
The genrule probably fails because it requires a shell.

> What I found would work is replacing the `genrule` with `write_file` rules (from skylib). Does that sound cool? I didn't add that in this PR as its a bit noisy but I could open up another PR?

That would probably be a better test, but it also adds an extra dependency. This can wait until another PR, as all of the testing needs to be made portable and more useful.

aiuto:
This is still coming up with one change requested, but I can not find it.
I think this is really close. I just want to give it one last careful pass, but I am tied up for a day or two with a few urgent things. I'll try to get back to you for an approve and merge by end of week.

pkg/tests/BUILD
j3parker: Ah, I was thinking `diff_test` because I'm nervous about cross-machine/cross-OS determinism. The zip format has some degrees of freedom where these kind of differences could theoretically sneak in. Checking in a (small) pre-generated zip could guard against that.

`zip_content_test` (etc.) seems cool. I wonder if `pkg_library` etc. could be used for `expected`?

j3parker:
Cool! No worries.

aiuto:
Can you rebase off of master so the change is just pkg_tar and not all the other changes too.

pkg/tests/zip_test.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+
+# These could be replaced with diff_test() rules (from skylib)
+class ZipEquivalency(ZipTest):
+    """Check that some generated zip files are equivalent to each-other."""
+
+    def assertFilesEqual(self, actual, expected):
+        """Given the names of some generated zips, assert that their contents
+           are equal."""
```
aiuto: Does this fit in 80 characters?  You can omit "generated" if not, since only one is generated and the other is golden.

pkg/tests/zip_test.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertZipFileContent(
+            "test_zip_package_dir0.zip",
+            [
+                {"filename": "abc/def/hello.txt", "crc": HELLO_CRC},
+                {"filename": "abc/def/loremipsum.txt", "crc": LOREM_CRC},
+            ],
+        )
+
+
+# These could be replaced with diff_test() rules (from skylib)
```
aiuto: NO!!!! diff_test generates bash or cmd scripts depending on platform. You are better off with filecmp.

pkg/tests/zip_test.py
```diff
@@ -0,0 +1,130 @@
...
+        Args:
+            zipfile: the test-package-relative path to a zip file to test.
+            content: an array of dictionaries containing a filename and crc
+                     key, and optionally a timestamp key.
+        """
+        with ZipFile(self.get_test_zip(zipfile)) as f:
+            infos = f.infolist()
+            self.assertEqual(len(infos), len(content))
+
+            for info, expected in zip(infos, content):
```
aiuto: This is going to be sort of brittle, because the content must match the expetected in order. And sometimes we want to just know that the times are correct and others that the names are right. It is fine for now, but what may work more flexibly is something like

                 expect_crc = expected.get('crc')
                 if expect_crc:
                      self.assertEqual(info.CRC, expect_crc)

So any of the fields is optional for the test.   But we can think about that later.

pkg/build_zip.py
```diff
@@ -0,0 +1,115 @@
...
+import argparse
+from datetime import datetime
+from helpers import SplitNameValuePairAtSeparator
+from zipfile import ZipFile, ZipInfo, ZIP_DEFLATED
+
+ZIP_EPOCH = 315532800
+
+def _get_argument_parser():
+  parser = argparse.ArgumentParser(
+    description = 'create a zip file',
```
aiuto: When you can fit on one line, please do so. See the style for the other tools. 
That applies to throughout.

pkg/build_zip.py
```diff
@@ -0,0 +1,115 @@
...
+  dst_path = _remove_leading_slash(dst_path)
+
+  return dst_path
+
+def parse_date(ts):
+  ts = datetime.utcfromtimestamp(ts)
+  return (ts.year, ts.month, ts.day, ts.hour, ts.minute, ts.second)
+
+def main(args):
+  package_dir = _remove_trailing_slash(args.directory)
```
aiuto: package_dir = args.directory.rstrip('/')

pkg/tests/BUILD
aiuto: > The zip format has some degrees of freedom where these kind of differences could theoretically sneak in.

We want to work that out of the process. Or, at least, to make sure that any bits which are different are no-ops. 
In any case, diff_test does not seem suited for binary comparisons

Yannic:
What's the status on this? I'd like to use this in https://github.com/google/closure-compiler/issues/3461

j3parker:
Ah, I forgot about this. I'll address @aiuto 's latest feedback today (beyond that I can not say).

pkg/tests/zip_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+    def setUp(self):
+        self.data_files = runfiles.Create()
+
+class ZipContentsCase(ZipTest):
+    """Use zipfile to check the contents of some generated zip files."""
+
+    def assertZipFileContent(self, zipfile, content):
+        """Assert that zipfile contains exactly the entries described by
+           content.
```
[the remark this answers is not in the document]
j3parker: Ok, I had to reword it a bit more to fit in 80, but it is done.

pkg/tests/zip_test.py
j3parker: I reworded this.

pkg/tests/zip_test.py
j3parker: Ok, I removed the comment.

pkg/build_zip.py
j3parker: Sure, sorry. In the latest push I did my best to match `build_tar.py`.

pkg/build_zip.py
j3parker: Neat. There was some more stuff like this in the file and I ended up simplifying the path manipulation a bit.

pkg/tests/zip_test.py
j3parker: Ok

statik:
Thanks for making this! I am excited to use the new `pkg_zip` rule in a project. I hope a new release of rules_pkg will be available soon now that this is merged.

aiuto:
I'm sorry for the delay.   pkg_zip is now in a release.
https://github.com/bazelbuild/rules_pkg/releases/tag/0.2.3

Much thanks to Jacob Parker for doing this work.

j3parker:
@statik glad to hear! Thanks for doing the release, @aiuto!

Keep in mind that this rule is pretty limited: if you want to create a zip that has a folder structure other than "all the files in the same (configurable) directory" it is not yet possible to do. There is a bigger-picture issue to create a reusable set of rules for describing rich folder structures that has been discussed in #30. I'll post an "update" there.

PR #67, opened 2019-07-30
unquote_and_split refactor

j3parker:
This sequence of changes moves the function to a helper file (to be shared with a zip tool) and renames the function + a parameter for clarity. I've added tests for the pre-existing (peculiar?) behaviour but otherwise haven't changed it.

### DIFF
--- a/pkg/tests/helpers_test.py
+++ b/pkg/tests/helpers_test.py
@@ -17,14 +17,14 @@ import tempfile
 import helpers
 
 
-class HelpersTestCase(unittest.TestCase):
-    def test_getFlagValue_nonStripped(self):
+class GetFlagValueTestCase(unittest.TestCase):
+    def testNonStripped(self):
         self.assertEqual(helpers.GetFlagValue('value ', strip=False), 'value ')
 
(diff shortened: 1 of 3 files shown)


pkg/tests/helpers_test.py
```diff
@@ -17,29 +17,80 @@
...
+    def testDoesntUnescapeVal(self):
+        key, val = helpers.SplitNameValuePairAtSeparator('abc=x\\=yz\\', '=')
+
+        self.assertEqual(key, 'abc')
+
+        # the val doesn't get unescaped at all
+        self.assertEqual(val, 'x\\=yz\\')
+
+    def testUnescapesNonsepCharsToo(self):
+        key, val = helpers.SplitNameValuePairAtSeparator('na\\xffme=value', '=')
```
aiuto:
```suggestion
        key, val = helpers.SplitNameValuePairAtSeparator('na\\xffme=value',
                                                                                             '=')
```

aiuto: Fooey. github won't let me merge the suggestion into the commit.
I'll merge now and the next person to touch this can fix the formatting.

PR #73, opened 2019-08-09
Convert pkg_tar and pkg_deb helpers to use argparse rather than Abseil flags

aiuto:
Part of #49

### DIFF
--- a/pkg/build_tar.py
+++ b/pkg/build_tar.py
@@ -225,7 +225,7 @@ def main():
                       help='A debian package to add to the layer')
   parser.add_argument(
     '--link', action='append',
-    help='Add a symlink a inside the layer ponting to b if a:b is specified')
+    help='Add a symlink a inside the layer pointing to b if a:b is specified')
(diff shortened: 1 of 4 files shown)


pkg/build_tar.py
```diff
@@ -262,47 +196,110 @@ def add_deb(self, deb):
...
+                      help='An empty dir to add to the layer.')
+  parser.add_argument('--empty_root_dir', action='append',
+                      help='An empty dir to add to the layer.')
+  parser.add_argument('--tar', action='append',
+                      help='A tar file to add to the layer')
+  parser.add_argument('--deb', action='append',
+                      help='A debian package to add to the layer')
+  parser.add_argument(
+    '--link', action='append',
+    help='Add a symlink a inside the layer ponting to b if a:b is specified')
```
vmax: @aiuto typo
```suggestion
    help='Add a symlink a inside the layer pointing to b if a:b is specified')
```

PR #75, opened 2019-08-09
Convert make_rpm to argparse and add an initial test for it.

### LINKED ISSUE #49: replace absl.flags with argparse in the tools (make_dep.py, make_rpm.py, ...)

aiuto:
Fixes #49 

There are two things going on here.
1. Convert make_rpm.py to use argparse
2. Add a minimal test to show it working
3. Drive by fixe to make_rpm.py because of a BUILDROOT issue.

While testing, I think I discovered a latent bug in the tool, w.r.t. BUILDROOT, so I had to fix that to make any pkg_rpm() instance build.  Regardless of that issue, the change to argparse is what this PR is about. If it turns out there is still flakiness w.r.t. BUILDROOT, we can fix that as a distinct issue.

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -17,6 +17,7 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+import argparse
 import contextlib
 import fileinput
 import os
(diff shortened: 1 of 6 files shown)


pkg/make_rpm.py
```diff
@@ -225,17 +209,18 @@ def SetupWorkdir(self, spec_file, original_dir):
   def CallRpmBuild(self, dirname):
     """Call rpmbuild with the correct arguments."""
 
+    buildroot = os.path.join(dirname, RpmBuilder.BUILD_DIR)
```
aiuto: This is the only interesting part of the change. It changes --builtroot from <tmpdir>/BUILDROOT to <tmpdir>/BUILD. Without that, it was not finding any of my sources.

pkg/make_rpm.py
```diff
@@ -281,16 +266,39 @@ def Build(self, spec_file, out_file):
...
+  parser.add_argument('--spec_file', required=True,
+                      help='The file containing the RPM specification.')
+  parser.add_argument('--out_file', required=True,
+                      help='The destination to save the resulting RPM file to.')
+  parser.add_argument('--rpmbuild', help='Path to rpmbuild executable.')
+  parser.add_argument('--debug', action='store_true', default=False,
+                      help='Print debug messages.')
+  parser.add_argument('files', nargs='*')
+
+  options = parser.parse_args()
```
brandjon: This form uses `sys.argv`, in which case the `argv` param is unused. I'd change this to pass in `argv`. Then `main()` can in principle be called programmatically by another module.

aiuto: Done

PR #78, opened 2019-08-15
Add overrides for setup fuctions to print_rel_notes

aiuto:
Add capability to set setup file name, deps method and toolchains method.


aiuto:
Fixes #78

PR #88, opened 2019-09-18
Relnotes: Support projects without deps.bzl file

fweikert:
The WORKSPACE stanza of the release notes should not always contain instructions to load a deps.bzl file, since projects may not have such a file. This behavior can be controlled by setting the has_deps_file parameter of the print_rel_notes macro.
It defaults to True for backwards compatibility.

I'd like to modify the Python scripts in a future commit to improve the handling of flags, and to support WORKSPACE stanzas that refer to the Bazel federation.

### DIFF
--- a/pkg/releasing/defs.bzl
+++ b/pkg/releasing/defs.bzl
@@ -1,6 +1,6 @@
 
 
-def print_rel_notes(name, repo, version, outs=None):
+def print_rel_notes(name, repo, version, outs=None, has_deps_file=True):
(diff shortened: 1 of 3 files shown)


aiuto:
I'm going to get 2.3 out the door to add pkg_zip, then we should combine ideas on this.
I was doing some related work in this branch.
https://github.com/aiuto/rules_pkg/tree/depmethod

fweikert:
Yeah, I also forgot about PR #78. I'll update my code.

fweikert:
This PR has been replaced by #93

PR #92, opened 2019-09-24
Closes #83. Keep owners while merging `pkg_tar`s into `pkg_tar`s

### LINKED ISSUE #83: File owners are losed when pkg_tar merged into other pkg_tar via `deps` attribute
`pkg_tar` with files owned by a particular owner loses this information if is used in other `pkg_tar` as dependency.
Erasing of owners occurs [here](https://github.com/bazelbuild/rules_pkg/blob/master/pkg/archive.py#L388)
What is the reason of the existence of this `numeric` argument?

ppodolsky:


ppodolsky:
@googlebot I signed it!

aiuto:
Related to #83.   I would like to discuss the change in the context of that.

There are tradeoffs between using numeric and textual owners which we need to be clear about. This change seems to shift the behavior from numeric to textual, but only when rebundling a tar file, which seems dubious. We are still using numeric in the other cases.

This will also be a breaking change, so if we decide to keep it we'll need to gate it with another rule attribute.

aiuto:
I should have been clearer in my comment from Sept. 25. The change, as is, can't go in.
- it will change existing behavior dramatically for some users so the choice of numeric id or not has to be an option to pkg_tar.
- there need to be tests around it.

aiuto:
Closing for staleness.  If you still want to try get this feature, please open an issue first where we can work through the design.

PR #93, opened 2019-09-26
WORKSPACE stanza: Do not include deps method by default

fweikert:
Previously the WORKSPACE stanza would include a deps method with a default name if no explicit value was set by the rule.
This caused problems for projects that don't require any dependencies to be loaded.

This commit ensures that the WORKSPACE stanza only contains a load statement for the setup file if the caller sets an explicit deps or toolchains method.

This PR is an improved version of #88.


fweikert:
Friendly ping

PR #94, opened 2019-09-30
[rfc] pkg_rpm: add $(location)-style .spec file substitution

beasleyr-vmw:
pkg_rpm allows users, in their .spec files, to refer to `data` files by those
files' basenames.  This can be unwieldy when working with large sets of files,
multiple files from distinct packages with the same basename, or files whose
names aren't known in advance.

This commit adds support for `$(location :label)`-style substitution in .spec
files.  `pkg_rpm` supports both `$(location)` and `$(locations)` for single
and multiple file attributes, respectively.  Package-relative and absolute
labels are both supported.

TODO: add test cases, documentation; prolly collapse this `for` loop with the one preceding it

### DIFF
--- a/pkg/rpm.bzl
+++ b/pkg/rpm.bzl
@@ -57,6 +57,26 @@ def _pkg_rpm_impl(ctx):
     for data_file in ctx.files.data:
         key = "{%s}" % data_file.basename
         substitutions[key] = data_file.path
+
+    # Supplement basename substitutions with label substitutions
+    # (ex: `{$(location :foo)}`).
+    for data in ctx.attr.data:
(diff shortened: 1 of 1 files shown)


aiuto:
Can you add some documentation and tests cases first. I can't tell from this change if this is a useful new feature ore a future can of worms.

beasleyr-vmw:
Sorry about the slow response 

> Can you add some documentation and tests cases first. I can't tell from this change if this is a useful new feature ore a future can of worms.

That's reasonable. I'm going to close this for now while kicking it around some more internally to see how we actually like it.

Issue #96, opened 2019-09-30
pkg_zip drops execute attribute when zipping executable file


statik:
I am using `pkg_zip` to zip up a `go_binary` generated using bazel_rules_go. The go_binary has the execute bit set, but when I unpack the contents of the generated zip file, the program does not have the execute bit set.

The goal I'm trying to accomplish is to deliver an artifact that can run in AWS Lambda, and the lambda environment is expecting the execute bit to be set.

I'm happy to try and contribute a fix for this, figured opening an issue was a good first step.

j3parker:
Ouch, that's unfortunate.

It looks like [ZipInfo.external_attrs](https://docs.python.org/3/library/zipfile.html#zipfile.ZipInfo.external_attr) might be the ticket. See [4.4.2 and 4.4.15 of this document](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT), maybe, for some more details.

j3parker:
Ah you beat me to it :) Thanks.

aiuto:
Solutions that look at the file system mode (like https://github.com/kindlyops/rules_pkg/commit/4a8c3e79f6cc83c10bab6357bfcce6be695d925e) do not work in general for Bazel, especially when remote build and caching are involved.  

Modes we want in the .tar (or .zip, .rpm, .msi, ...) have to be specified in BUILD files in some way.
Have you tried a **mode** attribute in your BUILD rules?

statik:
Thanks! I will give adding a mode attribute a try. 

Note that pkg_tar attempts to look at the mode of the source file as well, I wonder if that is a bug? https://github.com/bazelbuild/rules_pkg/blob/b8d6ea0a5465973ce0970f6e063dfebea473732c/pkg/build_tar.py#L69-L71

statik:
I've redone #97 to use a mode attribute in the BUILD rule as suggested, it seems to work well. Thanks!

scorsi:
Hope it's resolved I'm facing some trouble while migrating to Bazel build. :)

statik:
Hi @scorsi sorry you are hitting this. I don't know why #97 hasn't been merged yet. I've had to stop using rules_pkg because of this and #104.

As an alternative, this genrule works for zipping up a binary to deploy to AWS lambda
```
genrule(
    name = "lambda_deploy",
    srcs = [":lambda_binary"],
    outs = ["lambda_deploy.zip"],
[...]

PR #97, opened 2019-09-30
Control mode on zip inputs (#96)

statik:
This seems to fix the issue. Tried adding a test, I'm happy to improve/polish based on your feedback.


pkg/build_zip.py
```diff
@@ -65,6 +66,12 @@ def main(args):
       dst_path = _combine_paths(args.directory, dst_path)
 
       entry_info = ZipInfo(filename=dst_path, date_time=ts)
+      st = os.stat(src_path)
```
aiuto: This is not a general solution. See the comment in #96.

statik: Thanks for reviewing! I've redone the change to use a mode passed in the BUILD files with a sensible default modeled after what is done in pkg_tar.

Gloix:
Does the mode flag affect all source files? Say I want to pack an executable and a README file. Will they both have to be executable?

mitchelldavis:
Can we get this pushed through?

statik:
Thank you @aiuto! Life is busy, thank you for your work maintaining rules_pkg.

aiuto:
To answer @Gloix 's question. This looks like it will do all files in the same way.
I think of this as a stop-gap.  The more principled solution will use pkg_filegroup (when that gets finished) to finely specify modes and owners (in platform agnostic way) on a per-file basis.

PR #102, opened 2019-10-08
Kaoru/issue 54/pass target arch

KaoruDev:
## Problems
When building RPMs we cannot set the RPM's metadata on build. Thus when we try to install it in the target host, RPM complains that the RPM package was not built for the host's OS / CPU architecture.

## Solution
Allow `pkg_rpm` to accept an optional `target_platform` attr, which is passed to the `rpmbuild` setting the rpm's target platform.

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -155,11 +155,11 @@ class RpmBuilder(object):
   TEMP_DIR = 'TMP'
   DIRS = [SOURCE_DIR, BUILD_DIR, TEMP_DIR]
 
-  def __init__(self, name, version, release, arch, rpmbuild_path, debug=False):
+  def __init__(self, name, version, release, target_platform, rpmbuild_path, debug=False):
     self.name = name
     self.version = GetFlagValue(version)
     self.release = GetFlagValue(release)
-    self.arch = arch
+    self.target_platform = target_platform
     self.files = []
     self.rpmbuild_path = FindRpmbuild(rpmbuild_path)
     self.rpm_path = None
@@ -220,6 +220,10 @@ class RpmBuilder(object):
         '--buildroot=%s' % buildroot,
         self.spec_file,
     ]
+
+    if self.target_platform:
+        args.append("--target=%s" % self.target_platform)
+
     os.environ['RPM_BUILD_ROOT'] = buildroot
     p = subprocess.Popen(
         args,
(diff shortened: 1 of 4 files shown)


pkg/make_rpm.py
```diff
@@ -155,11 +155,11 @@ class RpmBuilder(object):
   TEMP_DIR = 'TMP'
   DIRS = [SOURCE_DIR, BUILD_DIR, TEMP_DIR]
 
-  def __init__(self, name, version, release, arch, rpmbuild_path, debug=False):
+  def __init__(self, name, version, release, target_platform, rpmbuild_path, debug=False):
```
aiuto: Please add this as a new option in addition to arch. Both have their place.  
It is actually a bug that arch is not being used, it should be. When we are missing platform, we can default arch from the target CPU for the toolchain.

pkg/rpm.bzl
```diff
@@ -142,7 +142,8 @@ pkg_rpm = rule(
             mandatory = True,
             allow_single_file = spec_filetype,
         ),
-        "architecture": attr.string(default = "all"),
+        "architecture": attr.string(default = "noarch"),
```
aiuto: In the long run, we should be aiming to help people built the spec file, so we want to be able to create a variable to define *BuildArch*.  Perhaps the default for this could be '@auto@'.  And within the RpmBuilder __init__() we compute arch. If arch is @auto@ we check target and parse out the first word. If no target, then we use 'noarch'

pkg/tests/make_rpm_test.py
```diff
@@ -156,13 +157,14 @@ def testBuild(self):
...
           'touch RPMS/test.rpm',
+          'echo "$@" > RPMS/test.rpm',
           'echo "Wrote: $PWD/RPMS/test.rpm"',
       )
       os.chmod(dummy, 0o777)
 
       with PrependPath([outer]):
         # Create the builder and exercise it.
-        builder = make_rpm.RpmBuilder('test', '1.0', '0', 'x86', None)
+        builder = make_rpm.RpmBuilder('test', '1.0', '0', 'x86_64-linux', None)
```
aiuto: The target should conform to the format we say we require. Maybe  'x86_64-generic-linux'

pkg/tests/make_rpm_test.py
```diff
@@ -174,6 +176,13 @@ def testBuild(self):
         # Make sure files exist.
         self.assertTrue(FileExists('test.rpm'))
 
+        found_arg = False
```
aiuto: Can you split the test so that there is one for --target and one without --target.

aiuto:
@KaoruDev friendly ping

KaoruDev:
I'm gonna be honest, I moved away from this library because it wasn't fitting my needs, I probably won't have time to modify the PR in any reasonable time. If anyone can pick this up that would be great. Sorry 😭

Issue #112, opened 2019-11-01
Reproducible RPM build host and build time


mmikitka:
If you want to control the RPM `Build host` metadata, use `_buildhost` in your RPM spec

If you want to control the `Build time` metadata, the `SOURCE_DATE_EPOCH` environment variable must be declared within the context of running rpmbuild.

PR is forthcoming with support for `SOURCE_DATE_EPOCH`

For context, see:
* https://reproducible-builds.org/docs/source-date-epoch/
* https://wiki.yoctoproject.org/wiki/Reproducible_Builds

Related to https://github.com/bazelbuild/rules_pkg/issues/105

aiuto:
Closing.  The build time part is done.  The rest should be addressed in #53

wxlg1117:
> 
> 
> If you want to control the RPM `Build host` metadata, use `_buildhost` in your RPM spec
> 

how to do ? please show a sample .spec ,thank you very much!

PR #113, opened 2019-11-01
Rpm source date epoch

mmikitka:
For https://github.com/bazelbuild/rules_pkg/issues/112

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -156,7 +156,7 @@ class RpmBuilder(object):
   DIRS = [SOURCE_DIR, BUILD_DIR, TEMP_DIR]
 
   def __init__(self, name, version, release, arch, rpmbuild_path,
-               source_date_epoch, debug=False):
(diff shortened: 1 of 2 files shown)


pkg/make_rpm.py
```diff
@@ -221,11 +223,17 @@ def CallRpmBuild(self, dirname):
         self.spec_file,
     ]
     os.environ['RPM_BUILD_ROOT'] = buildroot
+    env = {
```
aiuto: This mix of os.environ setting and env is strange. It never should have been that way in the first place.  It is not even clear how things worked the way they are, because using env at all would have suppressed RPM_BUILD_ROOT from being passed (at least in Python 3.7)  If we put everything in env it may be easier to understand.

mmikitka: I made the suggested change in the most recent commit, and also addressed the test failures.

mmikitka:
@aiuto Anything else required?

PR #117, opened 2019-11-11
pkg_deb: Support Breaks and Replaces.

benjaminp:
https://www.debian.org/doc/debian-policy/ch-relationships.html#overwriting-files-and-replacing-packages-replaces


pkg/pkg.bzl
```diff
@@ -344,9 +346,11 @@ pkg_deb_impl = rule(
         "depends_file": attr.label(allow_single_file = True),
         "suggests": attr.string_list(default = []),
         "enhances": attr.string_list(default = []),
+        "breaks": attr.string_list(),
```
aiuto: Can you make it default=[] for consistency.  Not that any of them need the default, but it will be easier for someone to understand in the future if they are all the same

benjaminp:
I have continued the redundant defaults.

benjaminp:
Thank you for the review.

Issue #114, opened 2019-11-06
Reproducibility: pkg_deb_impl with postinst and postrm attribute results in unpredictable control.tar.gz file


andreas-0815-qwertz:
### Description of the problem / feature request:

Using

```
http_archive(
    name = "rules_pkg",
    url = "https://github.com/bazelbuild/rules_pkg/releases/download/0.2.4/rules_pkg-0.2.4.tar.gz",
    sha256 = "4ba8f4ab0ff85f2484287ab06c0d871dcb31cc54d439457d28fd4ae14b18450a",
)

```

in my WORKSPACE together with a homegrown wrapper macro for `pkg_deb_impl`

```
def atos_pkg_deb(name, basename, **kwargs):
    out_deb = basename + ".deb"
    out_changes = basename + ".changes"
    pkg_deb_impl(
        name = name,
        out = name + ".deb",
        deb = out_deb,
        changes = out_changes,
        **kwargs
    )

```

results in a non-predictable file order of the postinst and postrm scripts in the generated `control.tar.gz`, when used in a target definition like

```
atos_pkg_deb(
    name = "ATOS-trc-target-deb-cpu_a",
    basename = "atos-A-TraceCfg-" + atos_version + "-powerpc",
    architecture = "powerpc",
    data = ":ATOS-trc-target-tarball",
    depends = ["cohpa (>= 2)"],
    description = """
 ATO Service
 Copyright: Bombardier Transportation
 Automatic Train Operation Service software
    """,
    maintainer = "http://www.bombardier.com",
    package = "atos-A-TrcConfig",
    version = atos_version,
    postinst = ":postinstall-trace-cohpa",
    postrm = ":postremove-trace-cohpa",
)

```

When I am building the target twice in a row with an intermediate `bazel clean`, I get the output attached with diffoscope 78.

[atos-A-TraceCfg-0.3.0.0-powerpc.html.gz](https://github.com/bazelbuild/rules_pkg/files/3815848/atos-A-TraceCfg-0.3.0.0-powerpc.html.gz)

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

Please see description above.

### What operating system are you running Bazel on?

Debian GNU/Linux 9.11

### What's the output of `bazel info release`?

`release 1.1.0`

(using bazelisk 1.1.0)

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

n.a.

[...]

PR #120, opened 2019-11-29
Make sequence of filenames in control.tar predictable

andreas-0815-qwertz:
Replacing plain Python dict by an OrderedDict for Python versions
before 3.7, so that iteration order of "extrafiles" is determined by
insertion order.  Since 3.7 iteration order of plain dict itself is
stable.

https://github.com/bazelbuild/rules_pkg/issues/114

### DIFF
--- a/pkg/make_deb.py
+++ b/pkg/make_deb.py
@@ -24,6 +24,11 @@ import tarfile
 import textwrap
 import time
 
+if sys.version_info < (3, 7):
+  from collections import OrderedDict
+else:
+  OrderedDict = dict
+
 
 # list of debian fields : (name, mandatory, wrap[, default])
 # see http://www.debian.org/doc/debian-policy/ch-controlfields.html
@@ -166,7 +171,7 @@ def CreateDeb(output,
               conffiles=None,
               **kwargs):
   """Create a full debian package."""
-  extrafiles = {}
+  extrafiles = OrderedDict()
   if preinst:
     extrafiles['preinst'] = (preinst, 0o755)
   if postinst:
@@ -296,7 +301,7 @@ def main():
                       help='The output file, mandatory')
   parser.add_argument('--changes', required=True,
                       help='The changes output file, mandatory.')
-  parser.add_argument('--data', required=True, 
+  parser.add_argument('--data', required=True,
                       help='Path to the data tarball, mandatory')
   parser.add_argument(
       '--preinst',


pkg/make_deb.py
```diff
@@ -24,6 +24,11 @@
 import textwrap
 import time
 
+if sys.version_info < (3, 7):
```
aiuto: We want to drop support for Python2, not make it work better.

aiuto: version_info is a 5 tuple. The compare should be on parts 0 and 1 if we need it.

And why the 3.7 test? collections.OrderedDict is available since 3.1

andreas-0815-qwertz:
Sorry for the long delay, I have been unavailable for some time ...

1. My rationale for the conditional import of OrderedDict is not related to
Python2 compatibility.  Rather, plain Python `dict` is assured to have a
stable iteration order (following the insertion order) from Python 3.7 on,
please see
https://mail.python.org/pipermail/python-dev/2017-December/151283.html and
https://docs.python.org/3/whatsnew/3.7.html.  Thus, the OrderedDict
implementation in the collections module is no longer needed in make_deb.py
from Python 3.7 on.  Using plain dict from 3.7 on might make make the
module robust against future changes regarding OrderedDict, see for example
https://twitter.com/raymondh/status/773978885092323328 from Raymond
Hettinger.

2. Tuples are guaranteed to compare lexicographically in Python.  The first
position of difference in two compared tuples decides about the outcome,
like follows:

Python 3.5.3 (default, Sep 27 2018, 17:25:39)
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> (3, 7, 0) < (3, 7)
False
>>> (3, 6, 9) < (3, 7)
True
>>> sys.version_info < (3, 5)
False
>>> sys.version_info
sys.version_info(major=3, minor=5, micro=3, releaselevel='final', serial=0)

I believe, this style of version comparison is not too uncommon, for
instance see
https://stackoverflow.com/questions/1093322/how-do-i-check-what-version-of-python-is-running-my-script
.


Please let me know, if this clarifies the issues or if you want me to
change the patch.


Thanks in advance,

Andreas


Am Mi., 4. Dez. 2019 um 22:25 Uhr schrieb aiuto <notifications@github.com>:

aiuto:
apologies myself for the long reply chain. I had to miss work for most of December and added some new responsibilities in January.  Together that created a large backlog that I am finally getting out from under.

Issue #116, opened 2019-11-11
Prepend prefix "./" to all files added to control.tar by make_deb


andreas-0815-qwertz:
dpkg packs files in control.tar using a "./".  I have no authoritative source for this, but the [Debian Administrator's Handbook](https://debian-handbook.info/browse/stable/packaging-system.html) for instance confirms this.

There are tools out there, relying on this convention; in my case it is the ancient version 0.1.8 of [opkg](https://github.com/shr-project/opkg).

aiuto:
Fixed in #121

PR #121, opened 2019-11-29
Add "./" prefix to files in control.tar

andreas-0815-qwertz:
This improves compatibility to Debian packages created using dpkg.

https://github.com/bazelbuild/rules_pkg/issues/116


aiuto:
I don't mean "yuuch" at your change. I mean yuuch at the deb package behavior that does the ./ prefix.

Issue #36, opened 2019-06-10
for discussion: pkgfilgroup as a way of structuring package inputs


aiuto:
The current packaging rules are not the most user friendly. This issue describes the key parts of what Google uses internally. I intend this to prompt discussion.

The basic idea is

- we have a pkgfilegroup rule that gathers sources and information about where they belong in the distributon
- the debian, rpm or tar builder depends on a *list* of pkgfilegroup rules to gather the content and build the package
- for the rpm case, you do not use classic .spec files to specify file modes. The pkg_rpm rule builds one for you and you include it from your .spec file.

So, it is a fairly heavy buy-in. You get more fine grained control over packaging, at the expense of doing things differently than you might have in the past. This is one of the motivators for not releasing Google's code originally. It is crufty in ways that other's might not like.

# pkgfilegroup

**pkgfilegroup(name, srcs, attr, prefix, section, strip_prefix)**

## srcs	
List of labels; optional

A list of rules that are dependencies of this rule. All output of each of the rules mentioned in the srcs attribute will be included in the packages that depend on this pkgfilegroup.

## attr	
List of strings; optional

Permissions and ownership of the installed files. This list has 3 elements. The first element is what permissions the installed files should have, specified in octal. The second element is the username or UID that the files should be user-owned by. The third element is the username or GID that the files should be group-owned by. Each element may instead be "-" which specifies that package shouldn't care about that particular element and leave it at the default.

## prefix	
String; required

Where the package should install these files. All output of each of the rules mentioned in the srcs attribute to this rule will be installed on the end-users machine relative to this prefix.

## section	
String; optional

Type of file being installed. The string can be blank, doc, or config. Additionally, config can take two optional arguments: missingok and noreplace. These are all syntactically valid values for section:
doc
config
config(missingok)
config(noreplace)
config(missingok,noreplace)
genrpm passes the value of section through to the %files section of the RPM's .spec file. RPM handles documentation and configuration files slightly differently from normal files. For example, RPM will not just overwrite configuration files. It will make sure the older configuration file, if installed, is preserved. missingok means that an error is not reported if rpm --verify is run and the file is missing. noreplace means that, when the RPM is being upgraded, it should not move the existing file out of the way. Instead, it should create the new file with an .rpmnew extension file next to the existing file. 

gendeb populates the conffiles control with the srcs of dependency pkgfilegroups with section set to config. gendeb ignores doc and the config optional arguments.

## strip_prefix	
String; optional; default is "."

[...]

PR #122, opened 2019-12-05
Add pkgfilegroup for archive-independent file package structuring mechanisms

nacl:
Currently, package contents need to be assembled in each of their `pkg_*` rules, and each one has a different mechanism for doing so.  This provides a unified interface, the `pkgfilegroup` rule, for accomplishing this task. Packaging rules can be written to take advantage of this in future changes.

Analysis tests are also included.

This is inspired by #36.

This is currently incomplete.  Features in development include:

- File renaming
- Non-file creation (e.g. directories, device nodes)

### DIFF
--- a/pkg/.bazelignore
+++ b/pkg/.bazelignore
@@ -1,0 +1,1 @@
+tests/external_project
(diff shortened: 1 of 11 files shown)


nacl:
Supplanted by #128.

PR #127, opened 2020-01-14
[Refactor] adding strip_prefix, remap_paths, include_runfiles to pkg_zip

grahamjenson:
I need to add `include_runfiles` attribute to `pkg_zip and wanted to bring the  `pkg_tar` and `pkg_zip` rules closer to parity, so this PR adds the prepare_files method which both rules use to generate the list of added files.

Also updated the readme.

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -103,7 +103,7 @@ Here, the Debian package is built from three `pkg_tar` targets:
 
 ```python
 pkg_tar(name, extension, strip_prefix, package_dir, srcs,
-mode, modes, deps, symlinks)
+mode, modes, deps, symlinks, include_runfiles)
 ```
 
 Creates a tar file from a list of inputs.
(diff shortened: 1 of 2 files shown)


grahamjenson:
@googlebot I signed it!

kigster:
Any word when this may be reviewed and merged?

cfeckardt:
+1 would like to see this merged soon too.

aiuto:
I think we should be trying for a more general solution that lets us define packages of files, how they remap, and if runfiles are included in rules distinct from tar, deb and rpm.  Then the packagers should get all the requisite remapping info in a provider from that rule.  https://github.com/bazelbuild/rules_pkg/pull/128 is trying to go down that path.  Let's see how that plays out first.

sreeramaraju:
`pkgfilegroup` from #127 is not public visible. Is there a workaround to pass strip_prefix for pk_zip rule till work related to that is merged to upstream?

davschne:
`strip_prefix` is now available for `pkg_zip` (see #221  above). I created the PR before I realized this one existed - whoops! I support this PR, though - it’s more complete and better leverages existing code. Could use some tests for `remap_paths` and `include_runfiles`.

PR #128, opened 2020-01-14
Add pkgfilegroup for package-independent destination mappings

nacl:
This adds an experimental rule, `pkgfilegroup`, along with associated Providers,
that gives rule developers and users a consistent mechanism for using the output
of bazel targets in packaging rules.

Inspired by #36.

Other capabilities that are provided by this that were not mentioned in #36 are:

- Creation of empty directories (`pkg_mkdirs`)
- Exclusion of files from a `pkgfilegroup` (`excludes`)
- Renames of files in a `pkgfilegroup` (`renames`)

### DIFF
--- a/pkg/.bazelignore
+++ b/pkg/.bazelignore
@@ -1,0 +1,1 @@
+tests/external_project
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -89,3 +89,12 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
 )
+
+# Used within experimental/tests/ for external repo genpkg tests
+filegroup(
+    name = "loremipsum_txt",
+    srcs = [
+        "testdata/loremipsum.txt",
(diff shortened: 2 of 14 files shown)


pkg/WORKSPACE
```diff
@@ -1,4 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 workspace(name = "rules_pkg")
 
 load("//:deps.bzl", "rules_pkg_dependencies")
+
 rules_pkg_dependencies()
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
```
aiuto: .bzl style is to put all the loads at the top of the file.

But the secondary question is why skylib is added here rather  than in rules_pkg_dependencies?

aiuto: Never mind. I had my federation hat on.  And that comment itself probably won't make sense.

PR #129, opened 2020-01-15
Add prototype pkgfilegroup-based RPM builder

nacl:
NOTE: only the third commit (currently bd8d531) should be reviewed as a part of this change.  The first two should be reviewed as a part of #128.

This change provides a prototype `pkgfilegroup`-based RPM builder in the form of
the `gen_rpm` rule.  See #128 for more details on `pkgfilegroup`. 

The RPM generator was derived from `make_rpm.py` in `pkg/` and supports a number
of features over and above what's available in `pkg_rpm`.  As written, it, given
a template like the one provided, can construct many full-fledged RPM
packages entirely within Bazel.  In most cases, the templates will only need to
be customized with advanced logic and other macros that are not settable via
bazel itself; `gen_rpm` will write much of the preamble, `%description` text,
`%install` scriptlets, and `%files` based on rule-provided inputs.

Documentation outside of the source files is not yet available.  This was
empirically tested on RPM packages internal to VMware with positive results;
actual tests of the rules are not yet ready.

This, naturally, is incomplete, and is missing capabilities such as:
- Configurable compression
- Configurable Provides/Requires
- SRPM emission
- Reproducibility
- Configurable stripping
- Configurable construction of "debug" packages

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -27,6 +27,8 @@ py_library(
     # I do not want to say '@rules_pkg'. That forces my name on everyone who uses
     # me. '@' seems to be the outer workspace.
     visibility = [
+        "@//experimental:__pkg__",
+        "@rules_pkg//experimental:__pkg__",
         "@//tests:__pkg__",
         "@rules_pkg//tests:__pkg__"
     ],
--- a/pkg/experimental/BUILD
+++ b/pkg/experimental/BUILD
@@ -13,6 +13,22 @@
 # limitations under the License.
 
 exports_files(
-    glob(["*.bzl"]),
+    glob([
+        "*.bzl",
+        "template.spec.in",
+    ]),
     visibility = ["//visibility:public"],
+)
+
+# Used by gen_rpm in genrpm.bzl.
+py_binary(
+    name = "make_rpm",
+    srcs = ["make_rpm.py"],
+    python_version = "PY2",
+    srcs_version = "PY2AND3",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@rules_pkg//:archive",
+        "@rules_pkg//:make_rpm_lib",
+    ],
 )
--- a/pkg/experimental/genpkg.bzl
+++ b/pkg/experimental/genpkg.bzl
@@ -343,6 +343,7 @@ pkgfilegroup = rule(
     special requirements.  Consult `pkg_mkdirs` for more details.
     """,
     implementation = _pkgfilegroup_impl,
+    # @unsorted-dict-items
     attrs = {
         "srcs": attr.label_list(
             doc = """Files/Labels to include in this target filegroup""",
@@ -537,6 +538,7 @@ pkg_mkdirs = rule(
     distribution's documentation for more details.
     """,
     implementation = _pkg_mkdirs_impl,
+    # @unsorted-dict-items
     attrs = {
         "dirs": attr.string_list(
             doc = """Directory names to make within the package""",
--- a/pkg/experimental/make_rpm.py
+++ b/pkg/experimental/make_rpm.py
@@ -1,0 +1,397 @@
+# vim: ts=2:sw=2:
+
+# Copyright 2017-2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""A simple cross-platform helper to create an RPM package."""
+
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+import argparse
+import contextlib
+import fileinput
+import os
+import re
+import shutil
+import subprocess
+import sys
+import tempfile
+from string import Template
+
+from helpers import GetFlagValue
+
+# Setup to safely create a temporary directory and clean it up when done.
+@contextlib.contextmanager
+def Cd(newdir, cleanup=lambda: True):
+  """Change the current working directory.
+
+  This will run the provided cleanup function when the context exits and the
+  previous working directory is restored.
+
+  Args:
+    newdir: The directory to change to. This must already exist.
+    cleanup: An optional cleanup function to be executed when the context exits.
+
+  Yields:
+    Nothing.
+  """
+
+  prevdir = os.getcwd()
+  os.chdir(os.path.expanduser(newdir))
+  try:
+    yield
+  finally:
+    os.chdir(prevdir)
+    cleanup()
+
+
+@contextlib.contextmanager
+def Tempdir():
+  """Create a new temporary directory and change to it.
+
+  The temporary directory will be removed when the context exits.
+
+  Yields:
+    The full path of the temporary directory.
+  """
+
+  dirpath = tempfile.mkdtemp()
+
+  def Cleanup():
+    shutil.rmtree(dirpath)
+
+  with Cd(dirpath, Cleanup):
+    yield dirpath
+
+
+WROTE_FILE_RE = re.compile(r'Wrote: (?P<rpm_path>.+)', re.MULTILINE)
+
+
+def FindOutputFile(log):
+  """Find the written file from the log information."""
+
+  m = WROTE_FILE_RE.search(log)
+  if m:
+    return m.group('rpm_path')
+  return None
+
+def SlurpFile(input_path):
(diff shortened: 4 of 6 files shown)


pkg/BUILD
```diff
@@ -89,3 +91,12 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
 )
+
+# Used within experimental/tests/ for external repo genpkg tests
```
aiuto: This broke things last time I merged it.
It seems unused, and there is no file testdata/loremipsum.txt. It is in tests/testdata.

nacl: Yeah, I have no idea how this got in here.  It will be removed when this PR is updated.

nacl:
> Can you merge with the current upstream so this is a smaller change.

Done.

aiuto:
Hi. I'm back from vacation, so I can take a look again.
There is a lot going on here.  It seems two orthogonal things going on.
1. working with pkgfilegroup
2. general improvements to make_rpm.py itself

What are your plans to get the improvements migrated from experimental up to the top level package?
In the long run, we are going to need that.

pkg/experimental/make_gen_rpm.py
```diff
@@ -0,0 +1,401 @@
...
+      if not os.path.exists(dst_dir):
+        os.makedirs(dst_dir, 0o777)
+      shutil.copy(os.path.join(original_dir, f), dst_dir)
+
+    # Copy the spec file, updating with the correct version.
+    spec_origin = os.path.join(original_dir, spec_file)
+    self.spec_file = os.path.basename(spec_file)
+    replacements = {}
+    if self.pre_scriptlet is not None:
+      replacements['BZL_PRE_SCRIPTLET_FROM_FILE'] = "%pre\n" + self.pre_scriptlet
```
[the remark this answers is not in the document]
nacl: Gotcha, thanks for the info.  Coincidentally, I'm on vacation myself right now, so it'll probably take me at least a few days to take a peek.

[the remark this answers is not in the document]
nacl: This should be addressed in 9e9e7a6.

Tested by running `rpm -qp --scripts $GENERATED_PACKAGE` on a package with both scriptlets as files and scriptlets as inline strings.

nacl:
> Hi. I'm back from vacation, so I can take a look again.
> There is a lot going on here. It seems two orthogonal things going on.
> 
>     1. working with pkgfilegroup
> 
>     2. general improvements to make_rpm.py itself
> 
> 
> What are your plans to get the improvements migrated from experimental up to the top level package?
> In the long run, we are going to need that.

From what I can tell, it shouldn't be too difficult to combine the two together; the logic could be made more robust to multiple template types.  Could also change the pkg_rpm starlark logic, but I think changes to `make_rpm.py` would be easier.

pkg/experimental/genrpm.bzl
```diff
@@ -0,0 +1,611 @@
...
+
+"""Provides rules for creating RPM packages via pkgfilegroup and friends."""
+
+load("@rules_pkg//experimental:genpkg.bzl", "PackageDirInfo", "PackageFileInfo")
+
+rpm_filetype = [".rpm"]
+
+spec_filetype = [".spec", ".spec.in"]
+
+def _gen_rpm_impl(ctx):
```
aiuto, line 23: Why don't we leave it as pkg_rpm.   The genrpm was a google internal name. The other rules in the package are pkg_

pkg/experimental/genrpm.bzl
```diff
@@ -0,0 +1,611 @@
...
+    )
+    files.append(rpm_files_file)
+    args.append("--file_list=" + rpm_files_file.path)
+
+    for f in ctx.files.data:
+        args.append(f.path)
+
+    #### Call the generator script.
+
+    # TODO(katre): Generate a source RPM.
```
aiuto, line 328: I don't think katre is signing up for this. And getting source RPMs is an entirely aspirational goal at this time.

pkg/experimental/genrpm.bzl
```diff
@@ -0,0 +1,611 @@
...
+
+            """,
+        ),
+        "license": attr.string(
+            doc = """RPM "License" tag.
+
+            The software license for the code distributed in this package.
+
+            """,
+            mandatory = True,
```
aiuto, line 500: I don't think this should be mandatory at this point. Internal things (which is the bulk of Bazel usage) usually don't think about their license terms.

pkg/experimental/make_gen_rpm.py
```diff
@@ -0,0 +1,397 @@
...
+      dst_dir = os.path.join(RpmBuilder.BUILD_DIR, os.path.dirname(f))
+      if not os.path.exists(dst_dir):
+        os.makedirs(dst_dir, 0o777)
+      shutil.copy(os.path.join(original_dir, f), dst_dir)
+
+    # Copy the spec file, updating with the correct version.
+    spec_origin = os.path.join(original_dir, spec_file)
+    self.spec_file = os.path.basename(spec_file)
+    replacements = {}
+    replacements['BZL_PRE_SCRIPTLET'] = "%pre\n" + self.pre_scriptlet
```
aiuto, line 225: The BZL_ part seems unnecessary.

aiuto:
What I was trying to get at was to not to do everything here and replace, but to take the rpm enhancements (like built in spec-file, pre/post scripts) and move them to the mainline code. Then have experimental derive from that.

But I'll be OK with getting this merged so that others can provide feedback no the pkg_filegroup stuff. While that is improving, we can backport the new features to the mainline code first, and bring in pkgfilegroup as we reach concensus.

pkg/experimental/make_gen_rpm.py
nacl: Hm.  I put `BZL_` part in from to make it easier to distinguish between hand-added parts of the template and ones that will be filled out by rules_pkg itself.

I don't feel attached to it, though.  I'll look into this.

pkg/experimental/genrpm.bzl
nacl: Unfortunately, `rpmbuild` bombs if it's not provided or if it's blank.

I can perhaps add a default value in there like "commercial" or "internal use only", but I feel that could be confusing.  How about I add some documentation as a compromise?

pkg/experimental/genrpm.bzl
nacl: Sure thing; I'll remove this.  This is actually also in pkg/rpm.bzl too: https://github.com/bazelbuild/rules_pkg/blob/master/pkg/rpm.bzl#L92

pkg/experimental/genrpm.bzl
nacl: Hm.  I think I'm fine with that.  Would it also make sense to rename the file to `rpm.bzl`, or perhaps `pkg_rpm.bzl`? It's easy enough for people to do things like:

```
load("@rules_pkg//experimental:rpm.bzl", pkg_rpm_ex = "pkg_rpm")
```

nacl:
> While we are still experimenting, can you name it pkg_filegroup, rather than pkgfilergroup so it matches the convention of all the other names.

Certainly.  How about I take care of this in a change that will go in after this one?

pkg/experimental/genrpm.bzl
nacl: Addressed in 229a244, filename changed to `rpm.bzl`.

pkg/experimental/genrpm.bzl
nacl: Removed in 229a244.

pkg/experimental/genrpm.bzl
nacl: Documentation added in 229a244

pkg/experimental/make_gen_rpm.py
nacl: Addressed in 229a244.

nacl:
> What I was trying to get at was to not to do everything here and replace, but to take the rpm enhancements (like built in spec-file, pre/post scripts) and move them to the mainline code. Then have experimental derive from that.
> 
> But I'll be OK with getting this merged so that others can provide feedback no the pkg_filegroup stuff. While that is improving, we can backport the new features to the mainline code first, and bring in pkgfilegroup as we reach concensus.

Ah, OK.  Fair enough.

I'll work on this next after I clear out my stack of pending changes.  I got symlink support and an RPM build tester in the pipe.

nacl:
It just occurred me that I'm probably making reviews difficult by force-pushing updates all the time.  Sorry about that.  Here's the text diff for the latest change:

[9e9e7a6-to-229a244.diff.txt](https://github.com/bazelbuild/rules_pkg/files/4477787/9e9e7a6-to-229a244.diff.txt)

aiuto:
One thing I truly hate about github reviews is that it is telling me I requested changes, and it won't point me to the place I asked for them.  There are more things, but that is one.


> > While we are still experimenting, can you name it pkg_filegroup, rather than pkgfilergroup so it matches the convention of all the other names.
> 
> Certainly. How about I take care of this in a change that will go in after this one?

Sure.

PR #132, opened 2020-01-24
Improve handling of sources from external repositories

### LINKED ISSUE #131: undefined behavior when pkg_tar.srcs comes from external repository
`pkg_tar` maps files based on each file's [`short_path`][1] attribute. If a file `package/file.txt` belongs to an external repository, `extrepo`, `short_path` evaluates to `../extrepo/package/file.txt`. `pkg_tar` with `strip_prefix="."` then creates archives with invalid `../` pathname elements:

```console
$ tar -tf test-tar-strip_prefix-dot.tar
./
./etc/
./etc/nsswitch.conf
tar: Removing leading `./..' from member names
./../
tar: Removing leading `./../' from member names
./../bazel_tools/
./../bazel_tools/tools/
./../bazel_tools/tools/python/
./../bazel_tools/tools/python/runfiles/
./../bazel_tools/tools/python/runfiles/runfiles.py
```

[1]: https://docs.bazel.build/versions/master/skylark/lib/File.html#short_path

beasleyr-vmw:
Avoid use of [`File.short_path`][1] when mapping filenames, because when
creating archives from files imported from external repositories we'll create
archive members with leading `./../` prefixes.  Instead, we'll stick to stripping
to leading `File.root.path` (if present) from `File.path`, resulting in archive
members like `external/repo/package/file.txt`.

[1]: https://docs.bazel.build/versions/master/skylark/lib/File.html#short_path

Resolves #131.

### DIFF
--- a/pkg/path.bzl
+++ b/pkg/path.bzl
@@ -13,9 +13,26 @@
 # limitations under the License.
 """Helper functions that don't depend on Skylark, so can be unit tested."""
 
+def safe_short_path(file_):
+    """Like `File.short_path` but safe for use with files from external repositories.
+    """
+    # Note: "F" is "File", "FO": is "File.owner".  (Lifted from genpkg.bzl.)
+    # | File type | Repo     | `F.path`                                                 | `F.root.path`                | `F.short_path`          | `FO.workspace_name` | `FO.workspace_root` |
+    # |-----------|----------|----------------------------------------------------------|------------------------------|-------------------------|---------------------|---------------------|
+    # | Source    | Local    | `dirA/fooA`                                              |                              | `dirA/fooA`             |                     |                     |
+    # | Generated | Local    | `bazel-out/k8-fastbuild/bin/dirA/gen.out`                | `bazel-out/k8-fastbuild/bin` | `dirA/gen.out`          |                     |                     |
+    # | Source    | External | `external/repo2/dirA/fooA`                               |                              | `../repo2/dirA/fooA`    | `repo2`             | `external/repo2`    |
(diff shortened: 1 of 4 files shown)


aiuto:
Hmmm.  This may have broken build_test.  The conffiles are coming out as ./config instead of config.
This may, or may not, be a problem. Can you look back at that.

beasleyr-vmw:
> Hmmm. This may have broken build_test. The conffiles are coming out as ./config instead of config.
> This may, or may not, be a problem. Can you look back at that.

Ack. Will look today.

beasleyr-vmw:
I don't think my patch was it. `git bisect` implicates another change:

```text
2f5c9815a7bde4f18acfde268bd63fedd107d87c is the first bad commit
commit 2f5c9815a7bde4f18acfde268bd63fedd107d87c
Author: andreas-0815-qwertz <57450822+andreas-0815-qwertz@users.noreply.github.com>
Date:   Wed Dec 4 22:32:01 2019 +0100

    Add "./" prefix to files in control.tar (#121)

    This improves compatibility to Debian packages created using dpkg.

    https://github.com/bazelbuild/rules_pkg/issues/116

:040000 040000 127832105299103ae09275519e9ed27af5b31c63 d269fdf8ae903dc58d1873615ad7da7f5752163e M      pkg
bisect run success
```

`//tests:build_test` succeeds after reverting that change. Seems like that commit should've been coupled with an update to the test data. PR coming shortly.

ali5h:
this removes external pkgs dirs from root of runfiles and breaks tar of python targets, since in python rules python paths points dirs at the root of runfiles instead of `{workspace/external/...}`

PR #142, opened 2020-03-25
Ensure the experimental/ tree lands in the output release archive

### LINKED ISSUE #104: Release assets don't match repo structure
The release assets helpfully produce tarballs that make recent releases look like old releases. This has avoided breaking users that update from release to release.

<details>
<summary> rules_pkg-0.2.4.tar.gz </summary>

```
$ curl -sLo -  https://github.com/bazelbuild/rules_pkg/releases/download/0.2.4/rules_pkg-0.2.4.tar.gz | tar tzf -
./
./BUILD
./README.md
./__init__.py
./archive.py
./build_tar.py
./build_zip.py
./deps.bzl
./helpers.py
./make_deb.py
./make_rpm.py
./path.bzl
./pkg.bzl
./rpm.bzl
./version.bzl
./releasing/
./releasing/BUILD
./releasing/__init__.py
./releasing/defs.bzl
./releasing/print_rel_notes.py
./releasing/release_tools.py
```

</details>

Unfortunately, that creates a divergence with the structure of the repo. Users that either follow commits on master or switch from releases to commits (to, say, get a particular fix before the next release) will observe strange-seeming issues like:

```
ERROR: error loading package '': Every .bzl file must have a corresponding package, but '@rules_pkg//:deps.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
ERROR: error loading package '': Every .bzl file must have a corresponding package, but '@rules_pkg//:deps.bzl' does not have one. Please create a BUILD file in the same or any parent directory. Note that this BUILD file does not need to do anything except exist.
```

That's because `:deps.bzl` is actually `pkg:deps.bzl` in the repo.

This divergence seems unusual given my experience with other `rules_*` projects (go, nodejs, etc). With those projects, I am mostly able to flip between releases, commits, and my own forks. I can imagine that the history of rules_pkg makes that different or harder.

Is this expected? Do you intend to close the gap between release assets and the repo structure in the future?

Thanks!

nacl:
The `experimental/` tree containing the then `pkgfilegroup` (and now `pkg_filegroup`) rule was missing from the 0.2.5 release archive.
    
This change ensures that the files in the experimental/ tree are included in release packages so users can use the rules within.
    
Testing was done by trying to run the experimental `pkg_rpm` rule on a dummy project importing rules_pkg.

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -10,7 +10,14 @@ exports_files(
 
 filegroup(
     name = "standard_package",
-    srcs = glob(["BUILD", "*.bzl", "*.py", "*.md"]),
+    srcs = glob([
+        "*.bzl",
+        "*.py",
+        "*.md",
+    ]) + [
+        "BUILD",
+        "//experimental:standard_package",
+    ],
     visibility = ["@//distro:__pkg__"],
 )
 
--- a/pkg/experimental/BUILD
+++ b/pkg/experimental/BUILD
@@ -32,3 +32,15 @@ py_binary(
         "@rules_pkg//:make_rpm_lib",
     ],
 )
+
+filegroup(
+    name = "standard_package",
+    srcs = glob([
+        "*.bzl",
+        "*.py",
+    ]) + [
+        "BUILD",
+        "template.spec.in",
+    ],
+    visibility = ["@//:__pkg__"],
+)


pkg/distro/BUILD
```diff
@@ -19,7 +19,7 @@ pkg_tar(
     mode = "0444",
     # Make it owned by root so it does not have the uid of the CI robot.
     owner = "0.0",
-    package_dir = ".",
+    package_dir = "pkg",
```
aiuto, line 22: This changes the location of the rules. Everyone using it would have to change from @rules_pkg//:pkg.bzl to @rules_pkg//pkg:pkg.bzl

nacl: Yes, that was the intent of the original issue.

Even if this code is not needed, the rest of it is to ensure that the experimental RPM generator lands in the release tarball.  Let me clean this up.

aiuto: Hmm... I've been thinking yesterday that this is something we might want to do anyway.
How about we discuss an optimal arrangement for the code first, and then move to that.

aiuto: That is,  move most of the files from pkg/*.bzl to pkg/pkg/*.bzl, but leave distro and tests at the same level.  Now the distribution tarball is only  pkg/... experimental/... and the top level BUILD and license.  We can put in a forwarder for a while so code using pkg:pkg.bzl will still work during migration.

WDYT?

aiuto: @dannysullivan because that move would impact docker's near term plan to point at archive.py

nacl: > That is, move most of the files from pkg/.bzl to pkg/pkg/.bzl, but leave distro and tests at the same level. Now the distribution tarball is only pkg/... experimental/... and the top level BUILD and license. We can put in a forwarder for a while so code using pkg:pkg.bzl will still work during migration.
>
> WDYT?

Makes sense to me.  We may even additionally want to move to the model of:

- `pkg/defs.bzl` provides forwarders for all non-experimental rules
- `pkg/internal/*.bzl` provides rule implementations

Not sure where the python and other scripts would go.  Same directory?  experimental/ could follow the same pattern.

In the meantime, I removed the line in question and included all the files needed to run the experimental rpm packager.

aiuto: I'm not a fan of internal. It makes the code harder to trace and, without enforced visibility, it does not ensure that people won't depend on it anyway.

PR #144, opened 2020-03-26
make_deb: fix length computation for extrafiles

matzf:
When adding the extrafiles (e.g. preinst, postinst etc.) with non-ascii characters in the content to the tar, a potentially faulty length was used. This would result in truncated files.

~~This seems to only occur when using files that are the result of build rules, i.e. the problem does not seem to occur when using a plain file as input for e.g. postinst.~~ Correction: apparently this difference does no longer exist, my completely un-educated guess is that this difference existed only in python2.

This seems to have been be an oversight in https://github.com/bazelbuild/bazel/commit/70e3c339bea02f6e9273b13de11d078ca6c2752f where the same problem was fixed for the main control file.
This change makes the handling of the length of the extrafiles identical to the main control file.

### DIFF
--- a/pkg/make_deb.py
+++ b/pkg/make_deb.py
@@ -149,9 +149,10 @@ def CreateDebControl(extrafiles=None, **kwargs):
       if extrafiles:
         for name, (data, mode) in extrafiles.items():
           tarinfo = tarfile.TarInfo('./' + name)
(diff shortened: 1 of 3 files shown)


aiuto:
This broke the python 2 test. I'm doing a quick fix.

Issue #138, opened 2020-03-06
Test rules_pkg on buildkite


laurentlb:
It seems like rules_pkg is not tested on buildkite

See https://github.com/bazelbuild/continuous-integration/issues/new/choose

aiuto:
duplicate of #10

PR #155, opened 2020-04-16
Preserve mtimes from input tar files in TarFileWriter output tars

### LINKED ISSUE #149: Feature: TarFileWriter tool optionally preserves input file mtimes
See https://github.com/bazelbuild/bazel/pull/11028 for initial conversation.

I work on a team that maintains deployment tools that use https://github.com/bazelbuild/rules_docker/ to build docker images for applications. We noticed that static files in deployed applications always had a Last-Modified value of the unix epoch. We traced the issue back to the TarFileWriter class, which copies all input files into an output tar and preserves most file metadata but overrides the mtime to a provided value (defaulting to 0).

I understand why, in many cases, users would want predictable mtimes for build reproducibility. I'm not sure I understand why this tool should always override mtimes, instead of making that an opt-in or opt-out feature. In our use case, we're passing an entire tar file through TarFileWriter as an input - I think it would be reasonable to allow the output metadata to match the input metadata.

dannysullivan:
This tool should not need to override mtimes for files contained in tar files that are added to the output of TarFileWriter, since these files are either constructed by other parts of the build process (in which case we can rely on the caching and reproducibility of those steps) or they are constructed outside of the bazel system, in which case we shouldn't need to alter them.

See discussion on [this issue](https://github.com/bazelbuild/rules_pkg/issues/149) for more information.

Fixes #149

### DIFF
--- a/pkg/tests/archive_test.py
+++ b/pkg/tests/archive_test.py
@@ -241,7 +241,8 @@ class TarFileWriterTest(unittest.TestCase):
 
   def testPreserveTarMtimesTrueByDefault(self):
     with archive.TarFileWriter(self.tempfile) as f:
-      input_tar_path = os.path.join(testenv.TESTDATA_PATH, "tar_test.tar")
+      input_tar_path = self.data_files.Rlocation(
+          os.path.join("rules_pkg", "tests", "testdata", "tar_test.tar"))
       f.add_tar(input_tar_path)
       input_tar = tarfile.open(input_tar_path, "r")
       for file_name in f.members:
@@ -251,7 +252,9 @@ class TarFileWriterTest(unittest.TestCase):
 
   def testPreserveTarMtimesFalse(self):
     with archive.TarFileWriter(self.tempfile, preserve_tar_mtimes=False) as f:
-      f.add_tar(os.path.join(testenv.TESTDATA_PATH, "tar_test.tar"))
+      input_tar_path = self.data_files.Rlocation(
+          os.path.join("rules_pkg", "tests", "testdata", "tar_test.tar"))
+      f.add_tar(input_tar_path)
       for output_file in f.tar:
         self.assertEqual(output_file.mtime, 0)
 
--- a/pkg/tests/archive_test.py
+++ b/pkg/tests/archive_test.py
@@ -227,6 +227,20 @@ class TarFileWriterTest(unittest.TestCase):
       f.add_tar(datafile, name_filter=lambda n: n != "./b", root="/foo")
     self.assertTarFileContent(self.tempfile, content)
 
+  def testPreserveInputTarMtimes(self):
+    input_tar_path = os.path.join(testenv.TESTDATA_PATH, "tar_test.tar")
+    with archive.TarFileWriter(self.tempfile) as f:
+      f.add_tar(input_tar_path)
+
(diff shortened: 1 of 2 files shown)


dannysullivan:
@aiuto please let me know if I missed something about your recommendation - my understanding is that removing this line gives us the behavior you were describing, where output tars always respect the mtimes of input tars.

aiuto:
So this is roughly https://github.com/bazelbuild/bazel/pull/11028/files
Let's err on the safe side and provide the option to update the mtime, but turn if off by default. Just like the suggestion there. Or, to put it differently. Change the behavior so it preserves mtime, but leave an escape valve.

dannysullivan:
No problem! This PR should be up to date with that requested change now.

pkg/tests/archive_test.py
```diff
@@ -227,19 +227,33 @@ def testMergeTarRelocated(self):
...
+  def testPortableMtime(self):
+    with archive.TarFileWriter(self.tempfile, default_mtime="portable") as f:
+      self.assertEqual(f.default_mtime, 946684800)
 
-    with tarfile.open(self.tempfile, "r:") as output_tar:
-      for output_file in output_tar:
-        self.assertEqual(input_mtimes[output_file.name], output_file.mtime)
+  def testPreserveTarMtimesTrueByDefault(self):
+    with archive.TarFileWriter(self.tempfile) as f:
+      input_tar_path = os.path.join(testenv.TESTDATA_PATH, "tar_test.tar")
```
aiuto, line 244: Where did testenv come from?  The tests are not passing because of this not being defined.  Suggest
      input_tar_path = self.data_files.Rlocation(
          os.path.join("rules_pkg", "tests", "testdata", "tar_test.tar"))

pkg/tests/archive_test.py
```diff
@@ -227,19 +227,33 @@ def testMergeTarRelocated(self):
...
+      f.add_tar(input_tar_path)
+      input_tar = tarfile.open(input_tar_path, "r")
+      for file_name in f.members:
+        input_file = input_tar.getmember(file_name)
+        output_file = f.tar.getmember(file_name)
+        self.assertEqual(input_file.mtime, output_file.mtime)
+
+  def testPreserveTarMtimesFalse(self):
+    with archive.TarFileWriter(self.tempfile, preserve_tar_mtimes=False) as f:
+      f.add_tar(os.path.join(testenv.TESTDATA_PATH, "tar_test.tar"))
```
aiuto, line 254: f.add_tar(self.data_files.Rlocation(
          os.path.join("rules_pkg", "tests", "testdata", "tar_test.tar")))

pkg/tests/archive_test.py
dannysullivan: Done; I was running the tests in another environment that had that environment setup. Is there a good way for me to see the test run results for these repo?

pkg/tests/archive_test.py
dannysullivan: Done.

pkg/tests/archive_test.py
aiuto: I sync to head, patch in the changes and do `bazel test ...`
We should get that running on the Bazel CI systems. There is a bug for that. #138

Issue #157, opened 2020-04-16
Don't force pkg_tar to use Python 2


davschne:
I'm having trouble understanding why `build_tar.py` [is specified to use Python 2](https://github.com/bazelbuild/rules_pkg/blob/1c799ba44baf2a79bd58580bed41ef414480c76f/pkg/BUILD#L40). In my project, which uses `pkg_tar`, this is causing problems, because we use a toolchain that only includes Python 3. I know there are some workarounds, but I wonder if it's possible to simply remove the `python_version = "PY2"` attribute when creating the `build_tar` target.

aiuto:
It may be historical accident.  I converted almost everything else, but still left behind the ability to use python2 for make_deb.py.  I don't see a need to support python2 forever.

aiuto:
fixed in #158

PR #158, opened 2020-04-16
Remove attribute python_version from build_tar target

davschne:
So build_tar.py isn't forced to run under Python 2.

See related issue: https://github.com/bazelbuild/rules_pkg/issues/157

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -81,7 +81,7 @@ py_binary(
 py_binary(
     name = "make_rpm",
     srcs = ["make_rpm.py"],
-    python_version = "PY2",
+    python_version = "PY3",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
--- a/pkg/experimental/BUILD
+++ b/pkg/experimental/BUILD
@@ -24,7 +24,7 @@ exports_files(
 py_binary(
     name = "make_rpm",
     srcs = ["make_rpm.py"],
-    python_version = "PY2",
+    python_version = "PY3",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
--- a/pkg/tests/BUILD
+++ b/pkg/tests/BUILD
@@ -121,7 +121,7 @@ py_test(
         "archive_test.py",
     ],
     data = [":archive_testdata"],
-    python_version = "PY2",
+    python_version = "PY3",
     srcs_version = "PY2AND3",
     tags = [
         # archive.py requires xzcat, which is not available by default on Mac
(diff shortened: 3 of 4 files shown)


pkg/BUILD
```diff
@@ -37,7 +37,6 @@ py_library(
 py_binary(
     name = "build_tar",
     srcs = ["build_tar.py"],
-    python_version = "PY2",
```
aiuto, line 40: Please use python_version = PY3, for consistency with the rest.

davschne: Okay!

davschne: Should I change `make_rpm` to use `"PY3"` as well? Otherwise, it will be the only one in this file that uses `"PY2"`, except for `make_deb_py2`, which I figure should stay as is.

I notice several of the tests use `"PY2"` as well. Perhaps they should change also?
- [`archive_test`](https://github.com/bazelbuild/rules_pkg/blob/ce53425bc5449268ade670346bc39d8c52b1f822/pkg/tests/BUILD#L124)
- [`make_rpm_test`](https://github.com/bazelbuild/rules_pkg/blob/ce53425bc5449268ade670346bc39d8c52b1f822/pkg/tests/BUILD#L148)
- [`helpers_test`](https://github.com/bazelbuild/rules_pkg/blob/ce53425bc5449268ade670346bc39d8c52b1f822/pkg/tests/BUILD#L163)

aiuto: If you can update the tests, that would be great.

davschne: Okay, done. I've updated everything to use `"PY3"` except the previously noted `make_deb_py2`.

aiuto:
Opps. The CLA bot is angry.

davschne:
Thanks for the PR approval. I'm just waiting for my employer's approval of the CLA. Hopefully soon!

aiuto:
I may have made this obsolete by doing something similar myself while waiting for the CLA. PTAL

davschne:
No worries. I didn’t expect the CLA approval to take so long; hopefully it goes through soon.

davschne:
CLA approval is working now, but the build is broken. Any idea why `//tests:helpers_test` is failing?

nacl:
> CLA approval is working now, but the build is broken. Any idea why `//tests:helpers_test` is failing?

Looking at:

https://storage.googleapis.com/bazel-untrusted-buildkite-artifacts/ea2b7e8d-0597-40c7-aebb-5042b1b3596e/tests/helpers_test/attempt_3.log

Seems that that there's a python2-ism in `helpers_test.py` that should be fixed.  The issue in the logs should be fixable by making the values to be written bytestrings instead of normal strings (`b'foo'` instead of just `'foo'`).  This should be reproducible locally too.

davschne:
Thanks, @nacl ; I'll see if I can resolve that.

PR #162, opened 2020-04-24
remove useless BUILD file.  update readme

aiuto:
The top level BUILD was a leftover from a restructuring. Since there is no corresponding WORKSPACE, it is useless.
The README was out of date.


aiuto:
Thanks.

Issue #10, opened 2018-04-24
Set up CI for packaging rules


katre:
We should be testing packaging rules against changes both in this repo, and in Bazel itself.

PR #165, opened 2020-04-27
Add starter CI setup

aiuto:
Notes:
WORKSPACE is not at the top level of the source, so all tests set `working_directory` to pkg. I
 wonder if there is a way to factor that up.
I want to start with Linux only now, because I have doubts about windows and macos portability. I will add those later.


aiuto:
Fixes #10

aiuto:
Thanks.  I forgot about that.

PR #169, opened 2020-05-05
convert py_binaries which are python3 ready to python3

aiuto:
I left RPM because it is not quite there yet and I don't know who requires it.
There is also the python2 version of make_deb, because people wanted that
:archive can probably convert, but I don't know if docker is ready for that.

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -38,8 +38,8 @@ py_library(
 py_binary(
     name = "build_tar",
     srcs = ["build_tar.py"],
-    python_version = "PY2",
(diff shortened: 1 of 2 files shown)


dannysullivan:
> :archive can probably convert, but I don't know if docker is ready for that.
Do you mean the rules_docker repo? I can check with them.

aiuto:
> > :archive can probably convert, but I don't know if docker is ready for that.
> 
> Do you mean the rules_docker repo? I can check with them.

Yes. But it is not urgent. I am happy to do these bit by bit.

Issue #174, opened 2020-05-11
Document experimental packaging rules


nacl:
The rules in pkg/experimental should be documented somewhat.  We (I) need to provide things like:

- An introduction
- General usage
- Thoughts on where things are going
- Stardoc emission

nacl:
Important things to discuss:

- How implicit package directory structures are created
- How package builders may "de-relevatize" paths

aiuto:
Also
- where to assign modes, owners, and groups to files
- symlinks

aiuto:
Obsolete. There is no more experimental.

PR #175, opened 2020-05-11
Provide pkg_mklinks for creation of in-package symbolic links

nacl:
Symbolic links are often included within packages, referring to files and
directories within and without.

This commit provides a rule, `pkg_mklinks`, which allows for the creation of
arbitrary symbolic links within a package, and support within the experimental
`pkg_rpm` rule to emit them.

`buildifier` was also opportunistically applied to files in this change.

Tests for integrating with the experimental RPM packager are available at https://github.com/nacl/rules_pkg/commit/352899c56f808d8e8dde1496f351e8134ee87dd1, and can be added to this PR after #160 is merged.


pkg/experimental/pkg_filegroup.bzl
```diff
@@ -578,6 +591,83 @@ pkg_mkdirs = rule(
...
+
+            The permissions value defaults to "0777".  The user/group values
+            default to "-".
+            """,
+            default = {"unix": ["0777", "-", "-"]},
+        ),
+        "section": attr.string(
+            doc = """Symlink section mapping.
+
+            Legal values for this attribute are:
```
aiuto, line 652: We should consider a way to not repeat this each time. But that can be another pass.

You don't really enforce this, do you? It's only the rpm builder which will.  Perhaps a more interesting way is to remove any restrictions on section, and change it to be a universal tag which is used by rpm, deb, whatever as they find appropriate. We could have each package builder filter the rules it does not understand. OTOH, that could lead to unintended missing specs, so probably not.

pkg/experimental/rpm.bzl
```diff
@@ -304,6 +327,19 @@ install -d %{{buildroot}}/{0}
                 install_script_pieces.append(install_dir_stanza_fmt.format(
                     d,
                 ))
+        if PackageSymlinkInfo in elem:
```
aiuto, line 330: Why don't we do the duplicate checks here rather than above?
It would only be one instance of the loop to get the destination elements.
It also seems like the error would be more useful if the fully formed destination was in the error.

pkg/experimental/pkg_filegroup.bzl
nacl: Finally got some time to look at all this.  Sorry for the delays.

> We should consider a way to not repeat this each time. But that can be another pass.

Yeah, some sort of common attributes list wouldn't be a bad idea.

> You don't really enforce this, do you? It's only the rpm builder which will. Perhaps a more interesting way is to remove any restrictions on section, and change it to be a universal tag which is used by rpm, deb, whatever as they find appropriate. We could have each package builder filter the rules it does not understand. OTOH, that could lead to unintended missing specs, so probably not.

Yeah, I really don't do much other than check for whether or not values are mentioned in a list.

Your proposal is indeed more flexible, along with your concerns.  I get the feeling they could perhaps be mitigated if there was a way to debug them easily -- perhaps a custom [command-line](https://docs.bazel.build/versions/master/skylark/config.html) flag and/or a rule attribute that increases logging or causes failures?

If you think the above is a good idea, I'll create some tasks to work on this.

pkg/experimental/rpm.bzl
nacl: The initial reason was to avoid having too much clutter in the loops.

You do make good points, though.  I'd rather not look into this now, but I've tracked it as a part of #191.

pkg/experimental/pkg_filegroup.bzl
```diff
@@ -578,6 +591,83 @@ pkg_mkdirs = rule(
...
+    This rule results in the creation of one or more symbolic links a package.
+    The link may point to a location outside of it.
+
+    """,
+    implementation = _pkg_mklinks_impl,
+    attrs = {
+        "links": attr.string_dict(
+            doc = """Link mappings to create.
+
+            The keys of this dict are the link names, the values are the source
```
aiuto, line 621: Do they really have to be source files? You might want the results of a target.
Can they be dangling?

pkg/experimental/pkg_filegroup.bzl
```diff
@@ -578,6 +591,83 @@ pkg_mkdirs = rule(
...
+                "Four-digit octal permissions string (e.g. "0755") or "-" (don't change from what's provided),
+                "User Id, or "-" (use current user)",
+                "Group Id, or "-" (use current group)",
+            ]
+            ```
+
+            The permissions value defaults to "0777".  The user/group values
+            default to "-".
+            """,
+            default = {"unix": ["0777", "-", "-"]},
```
aiuto, line 647: So what will happen on other OSes?  Random values?

pkg/experimental/pkg_filegroup.bzl
```diff
@@ -578,6 +591,83 @@ pkg_mkdirs = rule(
...
+    """,
+    implementation = _pkg_mklinks_impl,
+    attrs = {
+        "links": attr.string_dict(
+            doc = """Link mappings to create.
+
+            The keys of this dict are the link names, the values are the source
+            files.
+
+            Note that relative paths are important here: the package builder MAY
```
aiuto, line 624: I don't understand how you are applying "relative" here. 
Let's say you have a package  with a subfolder "data" (no BUILD file, so it is in the same package)

pkg_mklinks(links = {"bar":  "data/bar.dat"})

That should obviously be allowed.  What if data/bar.dat is actually another package, but not a dep of some other rule in this one. Is that allowed?

Could it be links = {"baz/bar":  "whatever"}?
Where baz does not exist anywhere in the source tree?

nacl:
Very interesting.  Unfortunately, I've been doing some high-priority internal porting work that has prevented me from looking at these reviews.  Sorry for not being responsive.

I hope to be able to start taking a look at these again later in July.

On 6/29/20, 23:27, "aiuto" <notifications@github.com<mailto:notifications@github.com>> wrote:


@aiuto commented on this pull request.

I think what we need before continuing with things like this is a good example of how they would be used to achieve desired effects. This current implementation looks like it would only work for RPM.

What I have seen people need though, is more complex things like

bar/BUILD:
cc_binary(name = "foo", data=["//some/other/place:data"])

pkg_filegroup(
name="dist",
content = [
pkg_file(src=":foo", dest=foo_bin/foo, attr=...)
pkg_file(src="//some/other/place:data", dest=foo_bin/data)
pkg_link(name="foo", dest="foo_bin/foo")
]
)

The above would, hypothetically, describe a unit of 3 files. Now, that should be consumable by another filegroup in another location (lets say the place where we gather foo and a few other programs), And finally the pkg_rpm should just be able to depend on that intermediate.

________________________________

In pkg/experimental/pkg_filegroup.bzl<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fbazelbuild%2Frules_pkg%2Fpull%2F175%23discussion_r447379194&data=02%7C01%7Capsaltis%40vmware.com%7C0bbe6b1e967d432a39e208d81ca57c3a%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637290844370827813&sdata=Nkdb9jn7qorpW%2ByWsQQpKC0%2BOtLZDaUaNo%2BNpOrqBb0%3D&reserved=0>:

> +        ),

+    ]

+

+pkg_mklinks = rule(

+    doc = """pkg_filegroup-like-rule for symlink management.

+

+    This rule results in the creation of one or more symbolic links a package.

+    The link may point to a location outside of it.

+

+    """,

+    implementation = _pkg_mklinks_impl,

+    attrs = {

+        "links": attr.string_dict(

+            doc = """Link mappings to create.

+

+            The keys of this dict are the link names, the values are the source

Do they really have to be source files? You might want the results of a target.
Can they be dangling?

________________________________

In pkg/experimental/pkg_filegroup.bzl<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fbazelbuild%2Frules_pkg%2Fpull%2F175%23discussion_r447379413&data=02%7C01%7Capsaltis%40vmware.com%7C0bbe6b1e967d432a39e208d81ca57c3a%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637290844370837811&sdata=BNEtZL0tmCqHGDPPlxkyAudK7nEihiXS0lnqR7HT9dc%3D&reserved=0>:

> +            doc = """Attributes to set for the output targets.

+

+            Must be a dict of:

+

+            ```

+            "unix" : [

+                "Four-digit octal permissions string (e.g. "0755") or "-" (don't change from what's provided),

+                "User Id, or "-" (use current user)",

+                "Group Id, or "-" (use current group)",

+            ]

+            ```

+

+            The permissions value defaults to "0777".  The user/group values

+            default to "-".

+            """,

+            default = {"unix": ["0777", "-", "-"]},

So what will happen on other OSes? Random values?

________________________________

In pkg/experimental/pkg_filegroup.bzl<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fbazelbuild%2Frules_pkg%2Fpull%2F175%23discussion_r447381226&data=02%7C01%7Capsaltis%40vmware.com%7C0bbe6b1e967d432a39e208d81ca57c3a%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C637290844370837811&sdata=q3Zex2LqBOpvL6RnOgzMdVE9SlUAsfWSmZk5ZVW8zWI%3D&reserved=0>:

> +pkg_mklinks = rule(

+    doc = """pkg_filegroup-like-rule for symlink management.

+

+    This rule results in the creation of one or more symbolic links a package.

+    The link may point to a location outside of it.

+

+    """,

+    implementation = _pkg_mklinks_impl,

+    attrs = {

+        "links": attr.string_dict(

+            doc = """Link mappings to create.

+

+            The keys of this dict are the link names, the values are the source

+            files.

+

+            Note that relative paths are important here: the package builder MAY

I don't understand how you are applying "relative" here.
Let's say you have a package with a subfolder "data" (no BUILD file, so it is in the same package)

pkg_mklinks(links = {"bar": "data/bar.dat"})

That should obviously be allowed. What if data/bar.dat is actually another package, but not a dep of some other rule in this one. Is that allowed?

Could it be links = {"baz/bar": "whatever"}?
Where baz does not exist anywhere in the source tree?

aiuto:
No problem. I have been super busy myself.

pkg/experimental/pkg_filegroup.bzl
nacl: > Do they really have to be source files? You might want the results of a target.

I think we may be crossing wires on terminology here.  In this case I'm referring to source-destination mappings within the archives themselves.

I'll take a peek 

>  Can they be dangling?

Yes.

I'll work on some documentation updates.

pkg/experimental/pkg_filegroup.bzl
nacl: The theory is that packager implementation can chose to interpret permission values as they wish.  Tarfiles on Windows, for example, will probably still want to use UNIX-style filesystem permissions.

The `unix` permissions mapping was intended entirely to support the common UNIX filesystem packaging use case (tars, debs, rpms).  Other keys to the dictionary could theoretically be added to support additional permissions types (UNIX extended attributes, Windows permissions, etc.).

There probably is a way to make this look prettier, but I haven't given it much thought.

pkg/experimental/pkg_filegroup.bzl
nacl: > I don't understand how you are applying "relative" here.

The issue at hand is that RPM packages only support defining paths relative to a root.  For example, if you say, have a mapping like this:

```
pkg_mklinks(links = {"bar": "data/bar.dat"})
```

This creates a symbolic link within the package named `bar` pointing to a file named `data/bar.dat`.  No checks are made to confirm whether `data/bar.dat` exists; the link is simply created.

Tar archives give you the option of specifying a path relative to the directory in which it's being extracted, or absolute relative to the UNIX filesystem root.  RPM packages are limited to files specified relative to the filesystem root, although [relocatable packages](http://ftp.rpm.org/max-rpm/ch-rpm-reloc.html) are possible to create.  However, AFAICT, paths in `%files` must still be specified relative to the filesystem root (or in this case, the relocation prefix root).

> Let's say you have a package with a subfolder "data" (no BUILD file, so it is in the same package)
>
>pkg_mklinks(links = {"bar": "data/bar.dat"})
>
> That should obviously be allowed. What if data/bar.dat is actually another package, but not a dep of some other rule in this one. Is that allowed?

This behavior is supported, yes.  It's tested as a part of the unit test updates added along with this pull request.

> Could it be links = {"baz/bar": "whatever"}?
> Where baz does not exist anywhere in the source tree?

Absolutely.  One of the nitty gritty details of this design is the implicit directory structures created by the various `pkg_filegroup`-like rules.  This will have to be thought about.

Regardless, my takeaway from this is that the documentation needs to be made more clear.  I'll adjust the documentation to make it clearer that we're explicitly talking about paths within packages.

nacl: Documentation updates done.

I decided to throw out mentions of how relative paths are controlled in favor of documenting it more comprehensively as a part of #174.

nacl:
> I think what we need before continuing with things like this is a good example of how they would be used to achieve desired effects.

Yeah, I've been a bit behind in this and haven't been able to explore using the existing implementation for, say, a tarfile or zip builder.

> This current implementation looks like it would only work for RPM. 

The implementation of `pkg_filegroup` or `pkg_mklinks`?  Both?

Regardless, how'd you come to this conclusion?  It could very well end up being that way; I'm honestly not sure.

> What I have seen people need though, is more complex things like
>
> bar/BUILD:
```python
cc_binary(name = "foo", data=["//some/other/place:data"])

pkg_filegroup(
    name="dist",
    content = [
        pkg_file(src=":foo", dest=foo_bin/foo, attr=...)
        pkg_file(src="//some/other/place:data", dest=foo_bin/data)
        pkg_link(name="foo", dest="foo_bin/foo")
    ]
)
```
> The above would, hypothetically, describe a unit of 3 files. Now, that should be consumable by another filegroup in another location (lets say the place where we gather foo and a few other programs), And finally the pkg_rpm should just be able to depend on that intermediate.

I agree composability of `pkg_filegroup`s is a good thing to have, but it hasn't so far been necessary for the use cases I've seen, however limited they may be.

However, I wonder how this could be implemented.  Is it possible to have a macro like this:

```python
def pkg_file(**kwargs):
    unique_name = _get_name_somehow() # from the args perhaps
    _some_rule_that_defines_the_package_mapping(name = unique_name, **kwargs)
    return ":" + unique_name
```

and use it in a rule?  I honestly never tried.  Such an implementation may indeed be easier to use, but may make debugging harder.

How would one get access to the `pkg_file` mapping's name?  Rules like that could perhaps benefit from being given anonymous names; line numbers may be enough for users to debug issues with their implementations.

pkg/experimental/pkg_filegroup.bzl
nacl: Documentation updated in https://github.com/bazelbuild/rules_pkg/pull/175/commits/4f31b82249c842f01540e477750b84745579a057

aiuto:
Sorry for the slow reply. I've been tied up with things for the last few days and am about to head out for a short vacation. I can take a look again on the 22nd.

aiuto:
cc/ @dannysullivan 

Fortuitously, I just saw a solution to your quesiton of how to get the pkg_file mapping.  We could use a provider for that. Something like

```
pkg_file = provider(fields={src, dest, attr, ...})
pkg_link = provider(fields={name, dest})

def pkg_filegroup(name, content):
  ...
  n_inputs = 0
  input_targets = []
  for target in content:
     if pkg_file in target:
        f_target = target[pkg_file]
           src = getattr(f_target, "src")
           attr = getattr(f_target, "attr", "0444")
           ... use src & attr as attributes to the pkg_filegroup invocation we are building
           ... OR have a _pkg_file rule which takes f_target and just echos the provider data
                     create a target of that rule here and add that target to srcs for pkg_filegroup
           target_name = 'pkg_src_%d' % n_inputs
           n_inputs += 1
           _pkg_file(name = target_name, file=f_target)
           input_targets.append(':' + target_name)
    if pkg_link in target:
        ... 
```

pkg/experimental/pkg_filegroup.bzl
dannysullivan: Hi @nacl - did you ever get a chance to create a follow-up tasks/issue to decide on how to reduce this repetition? I can do so if not.

nacl: Yup, check out #191.

nacl:
@aiuto I feel like this change can probably go in as-is and we can begin some broader design work.  WDYT?

As for that, I've been working on a quick mockup for a more hierarchical `pkg_filegroup` implementation that can be used as a discussion-starter.  Where do you think the discussion should be started?  In the meantime, I created a draft review in #212.

dannysullivan:
Merging as-is and moving the broader design conversation elsewhere sounds good to me - @aiuto, any thoughts?

aiuto:
I guess I am OK with that, because it is still in experimental.
Can you take care of the merge to see if that works correctly?

nacl:
> I guess I am OK with that, because it is still in experimental. Can you take care of the merge to see if that works correctly?

Done, thanks!  Looks good on my end.

PR #182, opened 2020-05-19
Add a minimal WORKSPACE file to the distributon.

### LINKED ISSUE #171: Add a top level WORKSPACE file to the distribution
This seems to be needed with the bazel integration tests.

I could argue that the bazel integration tests are hopelessly contorted and impossible to maintain, so I should fix those. But this is going to be easier.

blocking https://github.com/bazelbuild/bazel/pull/11270

aiuto:
We extract this from the top level workspace so it is easier to keep in sync.

I don't really like using sed for this, as it is another portability problem. OTOH, this is not a runtime problem, only distribution time, and we are very far away from being able to create distributions on windows.

Fixes #171

### DIFF
--- a/pkg/WORKSPACE
+++ b/pkg/WORKSPACE
@@ -23,7 +23,7 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
-### INTERNAL ONLY
+### INTERNAL ONLY - lines after this are not included in the release packaging.
 #
 # Include dependencies which are only needed for development here.
 
--- a/pkg/WORKSPACE
+++ b/pkg/WORKSPACE
@@ -23,6 +23,10 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
+### INTERNAL ONLY
+#
+# Include dependencies which are only needed for development here.
+
 http_archive(
     name = "bazel_stardoc",
     url = "https://github.com/bazelbuild/stardoc/archive/0.4.0.zip",
--- a/pkg/distro/BUILD
+++ b/pkg/distro/BUILD
@@ -13,6 +13,7 @@ load("@rules_python//python:defs.bzl", "py_test")
 pkg_tar(
     name = "rules_pkg-%s" % version,
     srcs = [
+        ":small_workspace",
         "//:standard_package",
         "//releasing:standard_package",
     ],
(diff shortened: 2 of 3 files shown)


pkg/WORKSPACE
```diff
@@ -23,6 +23,10 @@ load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
 
 bazel_skylib_workspace()
 
+### INTERNAL ONLY
+#
+# Include dependencies which are only needed for development here.
```
achew22, line 28: It's not 100% clear that there isn't a `END INTERNAL ONLY` or something like that. Maybe add a `lines after this are not included in the release version of the WORKSPACE file`?

aiuto: Having an END would mean more chance for pilot error (and certainly more chance for sed error).
But +1 to improving the comment.  PTAL

achew22: Sorry to not be clear. I think deleting to the end of file is definitely the right way to go.

aiuto: OK. Then we are on the same page.

pkg/distro/BUILD
```diff
@@ -25,6 +26,13 @@ pkg_tar(
     strip_prefix = ".",
 )
 
+genrule(
+    name = "small_workspace",
+    srcs = ["//:WORKSPACE"],
+    outs = ["WORKSPACE"],
+    cmd = "sed -e '/### INTERNAL ONLY/,$$d' $(location //:WORKSPACE) >$@",
```
nacl, line 33: My understanding from the [documentation](https://docs.bazel.build/versions/master/build-ref.html#workspace) is that it is not strictly necessary to strip these out unless commands are going to be run from within the `rules_pkg` workspace itself, which I imagine to be rare at best.  Does the CI testing do this or need a "complete" WORKSPACE somehow?

The release WORKSPACE could also be empty if it's otherwise not used, if `pkg/deps.bzl` was kept up-to-date.

OTOH, keeping this `sed` call in could also help with documentation.  What do you think?

aiuto: The intent is that if you expanded the distribution tarball and did `bazel test ...` it would work.
We don't need skydoc because the distro does not include the distro folder where we invoke skydoc.

That aside, I think you are getting to the seeming strangeness of this PR. This is not for the benefit of normal users. It is because I am trying to get the Bazel integration tests to work with this repo rather than the (obsolete) pkg_tar that they use. I can't explain why the external repos all require their own WORKSPACE - the testing infrastructure is too spaghetti to follow - but this seems to help.

nacl: > The intent is that if you expanded the distribution tarball and did `bazel test ...` it would work.
> We don't need skydoc because the distro does not include the distro folder where we invoke skydoc.

Fair enough.  I was wondering about something like this, but at least as 0.2.6, tests are excluded from the release tarball.  Are there plans to add them in?  They're relatively small at this point, so I don't see why not.

> That aside, I think you are getting to the seeming strangeness of this PR. This is not for the benefit of normal users. It is because I am trying to get the Bazel integration tests to work with this repo rather than the (obsolete) pkg_tar that they use. I can't explain why the external repos all require their own WORKSPACE - the testing infrastructure is too spaghetti to follow - but this seems to help.

Ah, OK.  Yay code pasta, I guess.

With all of these combined, the intent is clear.  Thanks!

PR #186, opened 2020-05-27
Add licenses clauses to BUILD files

aiuto:
This is not strictly needed, but it makes import to companies that do strict license checks (like Google) easier.
Also run buildifer on the changed files.

### DIFF
--- a/pkg/distro/BUILD
+++ b/pkg/distro/BUILD
@@ -1,13 +1,15 @@
-package(
-    default_visibility = ["//visibility:private"],
-)
-
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
 load("@rules_pkg//:pkg.bzl", "pkg_tar")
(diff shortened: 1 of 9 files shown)


pkg/distro/BUILD
```diff
@@ -1,14 +1,16 @@
...
 load("//releasing:defs.bzl", "print_rel_notes")
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
 load("@rules_python//python:defs.bzl", "py_test")
 
+licenses(["notice"])
+
+package(
+    default_visibility = ["//visibility:private"],
+)
```
nacl, line 12: nit: could probably throw this out if you wanted to

aiuto: I think I want it private. The code should not be able to depend on the rules which build the distribution from the code which depends on the rules which build the distribution from the code which ....

nacl: Hah. I totally agree that it should be private here; I was more commenting on how it's redundant, since `default_visibility` defaults to `//visibility:private`.

Issue #101, opened 2019-10-07
[FR] - Allow custom tar name


KaoruDev:
## Problem
When creating an RPM, I'd like to use the `%setup` macro, which I believe assumes you have a tar file with the same `NVR`. As it currently stands the `tar` rule will default to using the label name. I'd like to make this configurable (also using --define) with the default going to the label. I'd be happy to open a PR!

## Goal
Allow users to modify tar name.

aiuto:
I have no way to evaluate if this is a reasonable feature or not.  The %setup docs seem to talk about unpacking source tarballs, but then rpmbuild would try to build them, which is exactly what we wouldn't do with bazel.

Can you provide an example of what the goal is, and how you would expect to use it?

davschne:
I think this would be a generally useful feature, beyond just the problem identified above (which I confess I don't understand). As my team works with `pkg_tar`, we keep having to come up with awkward workarounds for the limitation that the generated archive must always be named the same as the corresponding target. To me it seems allowing this property to be configurable introduces flexibility without unduly complicating the design of the macros/rules.

aiuto:
I have to say, I don't understand the %setup thing either. I don't use RPM enough to understand the nuances.
The last comment seems related to #193

davschne:
Heh, yes, the proposals in #193 are a more complicated approach to creating the archive name (as well as `package_dir`). I'm interested to see where that discussion leads. Compared with that, I feel the present proposal is low-hanging fruit. For the project Konstantin and I are working on, both would be useful.

aiuto:
#193 is now ready for review.  It will allow you to rename archives any way you want. 
In the simple form, you just add another attribute with the name you want.
In the more complex usage, you also point to a rule providing a map of name/value pairs you can substitute into the filename. (e.g. {cpu}).   
Once that gets submitted, I can do the same thing for deb, rpm and zip.

aiuto:
Resolved by #198.

Issue #193, opened 2020-06-24
pkg_tar: Need archive name and package_dir to be configurable


konste:
We use pkg_tar rule to package build artifacts in a specific way requested by consumers. That requires the archive name and package_dir attribute to be calculated by a custom Bazel rule, in other words we need those parameters configurable. We would gladly contribute necessary changes to pkg_tar, but before jumping on the coding we would like to make sure the approach we take is acceptable. 
So far I only found two ways to make parameters configurable. Please let me know if I am missing something.
One way is to supplement the attribute of type "string" with another attribute of type "label" which would point to the file containing the value to use. It is exactly how package_dir_file attribute is implemented, but we would need to add another one for the archive name.
Another possible way is to introduce a new attribute **pkg_tar_config** of type "label" which would expect new PkgTarInfo provider. Our custom rule would calculate required strings, stick them into PkgTarInfo and return it. **pkg_tar_config** attribute then points to that custom rule and pkg_tar extracts the strings from the provider. 

Please let us know if either of the above two ways sounds like an acceptable solution and we can work on PR. If neither is acceptable then please explain how else parameters can be made configurable.

Workaround with renaming the archive after it is created does not work because .tgz contains zipped TAR and while we can rename .tgz we cannot rename inner .TAR archive and it is a blocker for us.

aiuto:
What is wrong with making a macro that wraps the pkg_tar call so that it can calculate the name and package_dir attributes?
Also for renaming the extension of the dependency, why can't you set the extension file on that target which is the dependency?

On a different tack, I don't want to add a Provider specific to tar. nacl@ is working on an enhanced pkg_filegroup provider that will handle things like renaming files, re-treeing other packages, and assigning the attributes desired in the package.  This is essentially all the things needed to define the layout of the package, without being specific to any package format. Then rpm, tar, deb, zip, and whatever will reuse that shared provider.

konste:
-- What is wrong with making a macro that wraps the pkg_tar call so that it can calculate the name and package_dir attributes?

We need name and package_dir to depend on build configuration, for example, target platform, bitness, debug/release, workspace status, custom command line parameters. All that data are not available at the Loading phase when macro is expanded. We can only access it in a custom rule at the Analysis phase and custom rule calculates the final name and package_dir based on many parameters.

-- Also for renaming the extension of the dependency, why can't you set the extension file on that target which is the dependency?

[...]

PR #194, opened 2020-06-24
Allow custom archive name for pkg_tar, pkg_deb, and pkg_zip targets

davschne:
Resolves Issue #101

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -299,7 +299,7 @@ pkg_tar_impl = rule(
     },
 )
 
-def pkg_tar(**kwargs):
+def pkg_tar(name, **kwargs):
     """Creates a .tar file. See pkg_tar_impl."""
 
     # Compatibility with older versions of pkg_tar that define files as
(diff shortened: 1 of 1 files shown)


pkg/pkg.bzl
```diff
@@ -312,9 +312,12 @@ def pkg_tar(**kwargs):
                       "This attribute was renamed to `srcs`. " +
                       "Consider renaming it in your BUILD file.")
                 kwargs["srcs"] = kwargs.pop("files")
+    archive_name = kwargs.get("archive_name") or name
```
aiuto, line 315: I'm torn over the best name for this attribute since the common word for tar it is archive; for zip, file; for .deb/.rpm, package.  But having universal word for the concept is better than being specific for each type. 

How does `package_name` sound?  Maybe pkg_name?

davschne: Regarding the name of the attribute, I struggled a bit with that, and I was interested to hear what you'd prefer to call it. I considered both `pkg_name` and `package_name`. I had a hard time deciding between them, because this library uses the abbreviation in some places and the full word in other places. Since `pkg` seems to be used for the names of rules and macros whereas `package` seems to be used for attribute names, I'd think the latter may be preferable for consistency. Ultimately I settled on `archive_name`, though, because the word "package" is heavily overloaded in this space. More specifically, `pkg_deb` already has a parameter called `package` that means something slightly different, and I thought it would be confusing for that macro to accept both `package` and `package_name`. What do you think?

Yes, I'm aware of #193. I work with Konstantin (who opened that issue), and we've discussed his proposal. Konstantin suggested the macros could be written so that it would be possible either to specify the package/archive name (and `package_dir`) directly or to have them generated via a provider.

terinjokes:
This doesn't work because `pkg_deb_impl` doesn't know the `archive_name` keyword.

```
no such attribute 'archive_name' in 'pkg_deb_impl' rule
```

aiuto:
You're right. I have a PR that fixes it, but I want to write some tests for the various cases. I won't get to that this week, most likely.  In the meantime, you could try the patch.

PR #201, opened 2020-07-13
Use GNU_FORMAT for tarfiles for dpkg compatibility.

itdaniher:
Add the format=tarfile.GNU_FORMAT to tarfile.open arguments to address Py3.8 change "The default format for new archives was changed to PAX_FORMAT from GNU_FORMAT."

and resulting

> unsupported PAX tar header type 'x'

on `dpkg --install`

Google CLA signed today.


itdaniher:
I've verified this fixes the "unsupported PAX tar header" issue when building debs containing tarballs via the following patch:

```patch
-http_archive(
+git_repository(
     name = "rules_pkg",
-    sha256 = "4ba8f4ab0ff85f2484287ab06c0d871dcb31cc54d439457d28fd4ae14b18450a",
-    url = "https://github.com/bazelbuild/rules_pkg/releases/download/0.2.4/rules_pkg-0.2.4.tar.gz",
+    remote = "https://github.com/itdaniher/rules_pkg.git",
+    commit = "b16882f7d5f15201dd3d49b5aadd67cabca57b9f",
+    patch_cmds = ["mv pkg/* ."]
 )
```

itdaniher:
@aiuto Thanks for the timely response. 

This is not a new feature, but a bugfix for a behavior change introduced in Python 3.8 causing `rules_pkg` creates invalid deb files if they contained tarballs. dpkg is not capable of parsing the created PAX-type headers, resulting in 
> corrupted filesystem tarfile in package archive: unsupported PAX tar header type 'x'"

upon install attempt.

I'll open an issue for this bugfix, which merely ensures rules_pkg produces the same format of archive on Ubuntu 20.04 as it does in your 14.04 CI.

Can you elaborate the use case for a flag (enabling PAX_FORMAT) which creates invalid deb files?

andrewalker:
@itdaniher oh, you hit basically the same issue as I did! Maybe the tests here are helpful: https://github.com/bazelbuild/rules_pkg/pull/204

andrewalker:
@aiuto it's worth noting that this patch is a no-op in Python < 3.8, since the default was GNU_FORMAT up until that point. This avoids a breaking change by keeping the default the same, so I also don't see the point in a flag.

aiuto:
I see your point, but I want to look into this deeper.  That probably won't be until next week.

aiuto:
Sorry I took so long. I wanted to fix the testing of debian packages first.  One #211 goes in, I will look at this and #204 and see which to apply.

aiuto:
Let's go with #204, since that has test cases for long file names.

Issue #207, opened 2020-07-31
strip_prefix=.  does not obey folder boundaries


aiuto:
Given packages   //tests and //tests_extra, with targets
tests:a == ["foo/bar"] and tests_extra == ["foo/bar"] 
we would expect the content to be
```
./foo/bar       # //tests is correctly stripped
./tests_extra/foo/bar"  

But for the second file we get   "./_extra/foo/bar".

dannysullivan:
@aiuto I just reproduced this and am planning on fixing it - please let me know if that's okay with you, or if there's anything else I should know here.

aiuto:
I'm always happy if other people fix bugs.

PR #208, opened 2020-07-31
Stop using the bazel_tools runfiles library in tests.

aiuto:
Using runfiles make it difficult to run the tests with Google's blaze. That slows down the progress for eventually converging the tools.

This also pointed out a bug in strip_prefix=.
https://github.com/bazelbuild/rules_pkg/issues/207


aiuto:
This won't work.  We need runfiles() to get tests with data to run on windows.

PR #212, opened 2020-08-04
Hierarchical pkg_filegroup implementation mockup

nacl:
This change is intended to start a discussion on how we might make a "nestable" `pkg_filegroup` rule, as @aiuto mentioned in the [review for `pkg_mklinks`](https://github.com/bazelbuild/rules_pkg/pull/175#issuecomment-663059149).

The mechanism for calling rules out in other rules was tested and is known to work, but I wonder about what to do with the `name` attribute.

A thought that comes to mind is that `pkg_file` could emit a specification that a (currently nonexistent) `pkg_filegroup` rule could use to provide a name to a "real" `pkg_file` rule (or however it is implemented).


aiuto:
Sorry for the slow response. I got busy with other responsibilities. I'll try to look in the next couple of days.

aiuto:
The 5 minute impression is that this needs a glob() in one of the sample cases to show how that behaves.

aiuto:
I'm still waiting to come out from under a deluge of work. It's annual
performance review season and that has put me off code for a couple of
weeks.

nacl:
Just posted an update that makes the code runnable.  It doesn't support much, but demonstrates how the file mapping mechanism might work.

I'm imagining things like globs will work similar to the existing `pkg_filegroup`.  If there are any collisions between directory permissions, I'm also imagining a concept of "ownership" that will allow us to select a "winner" in that regard.  Only one `pkg_filegroup` can "own" an individual directory in a package..

With this sort of scheme, I think we may be able to change the name of `pkg_filegroup` to something like `pkg_dir` to represent a node in the hierarchy.  WDYT?

aiuto:
Collisions to the final destination path of a package should fail hard. It is the most straightforward thing we can do.
A single input, however, can be mapped to multiple destination paths in the enclosing package.

You're right that pkg_filegroup is a strange name. bug pkg_dir is a misnomer too, because it can include things globed in from subdirs. It is occuring to me that we should brainstorm on this in a shard document rather than trying to make code work.     A meeting room and a whiteboard would help too.

The big questions.
- what expresses existing use cases in an easy to understand way
- how do we re-root things in a natural manner
- is the pkg_file inside the pkg_filegroup too extreme.   I would like to show that to a broader community than just us.
  pro:  It is a readable way of expressing a structured attribute.
  con: no one else does this.  It means pkg_filegroup MUST be a macro rather than a rule.

I think the pro outweighs the con, but I am not sure.  This might not even be legal syntax for Starlark.

aiuto:
Whoops.... I did not mean to close this.

aiuto:
Ah.... I see it.  The PR is against the master branch instead of main. It may need a rebase.

nacl:
Hm.  Apparently I can't change the destination branch either, like I can with an already open PR.  Sigh.

I'll open up a new PR for this.

Issue #213, opened 2020-08-07
Creating and examining tar files on Windows seems to loose the directory entries.


aiuto:
While working on #211 I noticed that the directory elements to not appear.
This might be because Python tarfile read on windows might not be reporting them,
or tarfile write might not be creating them. 

It's not like there is a lot of demand for producing debian packages from windows
at this time, so I'm just filing this for now to come back and resolve later.

nkoroste:
I think the same is true on Linux and Mac

PR #215, opened 2020-08-07
use runfiles for pkg_tar_test to enable on Windows

aiuto:
Using os.environ('RUNFILES') does not work on WIndows.


aiuto:
I did not set up a windows machine yet.  I discovered the problem while trying to replace runfiles with reading the environment. in a different PR. When I did that, the windows test broke, but linux and macos worked. Then I understood the deeper problem.   For this PR, all I needed to do was switch to the library, test on linux and macos. The next step will be to deal with the deeper problem of #213.  For that I will need a windows machine. I can get cloud instance for that.

PR #221, opened 2020-08-10
Add attribute strip_prefix to pkg_zip

### LINKED ISSUE #170: pkg_zip removes all directories when packing
pkg_zip uses an invocation of `dest_path` that triggers a buggy edge case (when strip_path is None)

https://github.com/bazelbuild/rules_pkg/blob/06e57ddaa6220ade969e710832d84ca1a377b522/pkg/pkg.bzl#L388-L389

The faulty logic for the None case in dest_path causes it to strip the entire (package-relative) directory of a file:

https://github.com/bazelbuild/rules_pkg/blob/7a991dea418ab17c7e86f0a7b5e7d4a87ef4304b/pkg/path.bzl#L42-L49

Running `bazel` with the `-s` switch demonstrates that this is indeed what gets passed to the python tool:

```
bazel-out/host/bin/external/rules_pkg/build_zip 
  -o bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_zip.zip 
  -d / 
  -t 315532800 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/README.md=/README.md' 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/_BUILD=/_BUILD' 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/index.d.ts=/index.d.ts' 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/index.js=/index.js' 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/package.json=/package.json' 
  'bazel-out/darwin-fastbuild/bin/packages/aaa/preTokenLambda/preTokenLambda_npm/src/preTokenLambda.d.ts=/preTokenLambda.d.ts'

### PR DESCRIPTION
**Pull request by davschne**
Resolves #170

I've tried to replicate the behavior of the eponymous attribute in `pkg_tar`, though I find this behavior counterintuitive in some cases. Specifically, see the test packages:
- `//tests:test-tar-strip_prefix-empty`
- `//tests:test-tar-strip_prefix-none`

My intuition is that the packages built from these targets should contain:
```
/
/etc/
/etc/nsswitch.conf
```
In other words, the default behavior (when `strip_prefix` isn't provided or is empty) should be to create paths in the archive relative to the package of the target.

Instead they contain only:
```
/nsswitch.conf
```
In other words, subdirectories are ignored.

So to clarify, this PR resolves #170 only when the value of `strip_prefix` isn't falsy in Starlark.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -390,9 +390,6 @@ def pkg_deb(name, package, **kwargs):
         **kwargs
     )
 
-def _format_zip_file_arg(f):
-    return "%s=%s" % (_quote(f.path), dest_path(f, strip_prefix = None))
-
 def _pkg_zip_impl(ctx):
     args = ctx.actions.args()
 
@@ -401,10 +398,12 @@ def _pkg_zip_impl(ctx):
     args.add("-t", ctx.attr.timestamp)
     args.add("-m", ctx.attr.mode)
 
-    args.add_all(
-        ctx.files.srcs,
-        map_each = _format_zip_file_arg,
-    )
+    for f in ctx.files.srcs:
+        arg = "%s=%s" % (
+            _quote(f.path),
+            dest_path(f, compute_data_path(ctx.outputs.out, ctx.attr.strip_prefix))
+        )
+        args.add(arg)
 
     args.set_param_file_format("multiline")
     args.use_param_file("@%s")
@@ -434,6 +433,7 @@ pkg_zip_impl = rule(
         "timestamp": attr.int(default = 315532800),
         "mode": attr.string(default = "0555"),
         "out": attr.output(),
+        "strip_prefix": attr.string(),
         # Implicit dependencies.
         "build_zip": attr.label(
             default = Label("//:build_zip"),
--- a/pkg/tests/BUILD
+++ b/pkg/tests/BUILD
@@ -5,6 +5,7 @@ licenses(["notice"])
 load("//:pkg.bzl", "pkg_deb", "pkg_tar", "pkg_zip")
 load("//:rpm.bzl", "pkg_rpm")
 load("@rules_python//python:defs.bzl", "py_test")
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
 
 genrule(
     name = "generate_files",
@@ -13,6 +14,12 @@ genrule(
         "usr/titi",
     ],
     cmd = "for i in $(OUTS); do echo 1 >$$i; done",
+)
+
+copy_file(
+    name = "zipcontent_loremipsum",
+    src = "testdata/loremipsum.txt",
+    out = "zipcontent/loremipsum.txt",
 )
 
 pkg_tar(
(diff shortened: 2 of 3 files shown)


aiuto:
It's going to be a few days before I can look at this - I'm behind on lots
of things because of the power outages in the NYC area last week.
My first thought is - can we avoid this and build on the pkg_filegroup work
that @nacl is exploring?

davschne:
No worries, @aiuto. I'm not yet familiar with the `pkg_filegroup` proposal, but I'll look into it and provide feedback. You can hold off reviewing this PR until then.

Subsequent to posting this PR, I noticed [this one](https://github.com/bazelbuild/rules_pkg/pull/127) that adds several attributes to `pkg_zip`, including `strip_prefix`. If we decide to go ahead with the change, I'll see if I can consolidate my work with that work.

davschne:
@aiuto, while I'm not sure I fully understand https://github.com/bazelbuild/rules_pkg/pull/212 or the motivations for it, my understanding is that it would introduce breaking changes to the APIs of `pkg_zip`, `pkg_tar`, and `pkg_deb` - that it would represent a new pattern for how to form archives. Is that correct? It may indeed be a good direction to go in the long term, say for the next major version of this library. My team needs increased parity between `pkg_zip` and `pkg_tar` in the near-term, though, and it would be nice if we didn't have to adapt all our existing packaging code to a new API in order to achieve it. I'd argue for merging this PR, or better yet https://github.com/bazelbuild/rules_pkg/pull/127, in the near future, since they're aimed at filling in some gaps in the current API, and not block their progress while a new API is being considered.

aiuto:
Yes. The new idea is a breaking change.  But your point about next major version is well taken.
We can keep this for now, but I am not going to invest my own cycles in remapping and prefix striping on a per rule basis.
We can easily branch this to support 1.x and 2.x development. In the 2.x tree we'll have a more principled (and understandable) way of composing packages with parity across all rules.

davschne:
Sounds good, @aiuto. Thanks for the merge. I believe #127 will fill in some additional gaps in `pkg_zip`’s current API with respect to the other packaging macros. I doubt you’d need to invest your own time in ironing out the discrepancies - others are eager to work on it! :)

cplee:
looks like this landed on `master` instead of `main`.  Also, will you be planning a release of this anytime soon or if not, how would I use an unreleased version of these rules?

nacl:
> looks like this landed on master instead of main.

Bah.  Thanks for the catch.

I opened a PR (#229) to merge `master` into `main` to rectify this, but I'm not sure of the best way to get that past the CLA bot.  @aiuto?

> Also, will you be planning a release of this anytime soon or if not, how would I use an unreleased version of these rules?

I can't really speak to releases, but if you want to use the rules at HEAD, you'd need to do something like what's described here: https://github.com/bazelbuild/bazel/issues/10062#issuecomment-642144553

aiuto:
I think the CLA bot failure is because you are forwarding (for lack of the
right term) a commit from davschne (
https://github.com/bazelbuild/rules_pkg/pull/229/commits/15fb7a0f803c616668daa8fd48dcbe8e1df225d4
)
and that is confusing the bot w.r.t. authorship.

@davschne  Would you like to try the equivalent?  Or do you mind me just
copying out the change and re-applying
to main?

davschne:
Sure, I'll do it.

PR #225, opened 2020-08-21
Support "provides" in pkg_deb

### LINKED ISSUE #133: Feature: support provides field in pkg_deb
It would be great to have support for adding the `provides` field to a debian package.
See also https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides

Judging from the code it should be fairly easy to add.

guoshimin:
Fixes #133


pkg/tests/pkg_deb_test.py
```diff
@@ -160,6 +160,7 @@ def test_description(self):
 	'Built-Using: some_test_data',
 	'Replaces: oldpkg',
 	'Breaks: oldbrokenpkg',
+    'Provides: hello',
```
aiuto, line 163: The indent does not look right here. Can you check?

guoshimin: Hmm, there seems to be a mixture of tabs and spaces in the original file. I changed the indentation for this line to be a tab to match the existing style.

PR #227, opened 2020-08-24
Avoid stripping prefixes with incomplete directory names.

dannysullivan:
Fixes https://github.com/bazelbuild/rules_pkg/issues/207.

The only complication here was the fact that the parameters `strip_prefix` and `f` have had the output data directory prepended to them by the time they are passed into `dest_dir`, so we can't just throw out `strip_prefix` entirely if the last directory is an incomplete match; instead, we need to act as if `strip_prefix` was set to `.` (or not specified at all) for those files. This issue was caught by `:pkg_tar_test`, which prepends the directory `tests` to all outputs.

Please let me know if you can think of a better approach than calling `compute_data_path` twice in `pkg.bzl`.

### DIFF
--- a/pkg/path.bzl
+++ b/pkg/path.bzl
@@ -39,10 +39,9 @@ def _short_path_dirname(path):
         return ""
     return sp[:last_pkg]
 
-def dest_path(f, strip_prefix):
+def dest_path(f, strip_prefix, data_path_without_prefix = ""):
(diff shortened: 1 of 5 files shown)


aiuto:
@dannysullivan  Were you going to merge this?

dannysullivan:
Sorry about that! I made this PR right before going on vacation and forgot to check on it when I got back - merging now.

PR #230, opened 2020-09-09
Add attribute strip_prefix to pkg_zip

davschne:
Duplicates PR #221 targeting default branch `main` instead of `master`.

* Add attribute strip_prefix to pkg_zip
* Add tests for pkg_zip strip_prefix behavior

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -392,9 +392,6 @@ def pkg_deb(name, package, **kwargs):
         **kwargs
     )
 
-def _format_zip_file_arg(f):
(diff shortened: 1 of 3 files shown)


aiuto (approved):

alexeagle:
@aiuto hey there, how about a rules_pkg release? latest is from March.

aiuto:
Sure. I had hoped to have more features before a new release, but we are
did a lot of work in the last few months and are now more in a design phase
rather than feature sprint, so why not. Stay tuned.

PR #233, opened 2020-09-24
Add LICENSE to distribution tarball

### LINKED ISSUE #231: rules_pkg-0.2.6.tar.gz not redistributable — missing license file
The release https://github.com/bazelbuild/rules_pkg/releases/download/0.2.6-1/rules_pkg-0.2.6.tar.gz is missing a license file. According to https://github.com/bazelbuild/rules_pkg/blob/0.2.6-1/LICENSE it is licensed under the Apache 2.0 license, and according to § 4a of the license
 
>      (a) You must give any other recipients of the Work or
>          Derivative Works a copy of this License; and

In order for rules_pkg-0.2.6.tar.gz to be redistributable a copy of the LICENSE file should be included in the tarball.

aiuto:
Fixes #231

To be landed in 0.2.7.

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -16,6 +16,7 @@ filegroup(
         "*.md",
     ]) + [
         "BUILD",
+        "LICENSE",
         "//experimental:standard_package",
     ],
     visibility = ["//distro:__pkg__"],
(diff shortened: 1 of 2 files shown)


nacl (approved):
It'd be nice if we could just take the LICENSE from the package root, but this works too.

Issue #228, opened 2020-08-29
Add a CHANGELOG/release notes generator


aiuto:
Looking through github logs is tedious. We could have a tool which would read the logs back to a particular point (usually last major or minor release) and use that to:

Extract RELNOTES sections from them and print them.
If they fix issues, find those issues and add them to an "Issues Fixed" section
Extract authors and use that to build a thank-you section.

The bazel release tool does something similar, but it is in bash and not designed for flexibility.

nacl:
Was this resolved by #234?

aiuto:
No.  I have tools that update CHANGELOG from commit comments, but the tools are not ready to check in. They need tests and portability beyond linux.Since they call git to get the comments, I need to build a git-toolchain to autodetect git on each platform. Then, how do I unit test the comment extraction tools?  I hacked the features together in a couple of hours. Making it production ready will take about 10 more.   It feels like classic yak-shaving.

statik:
I don't know if it does everything you are looking for, but I've been getting good mileage out of https://github.com/marketplace/actions/release-drafter on various projects to gather up release notes as merges happen.

achew22:
To toss another one out, [`chyle`](https://github.com/antham/chyle) is written in go and doesn't have any dependencies on system `git` which might make multi-os compatibility easier.

Alternatively, you could try using [`go-git`](https://github.com/go-git/go-git) (a pure go implementation of `git`'s storage/transport protocol) directly and probably get something without too much fuss.

aiuto:
I would rather not use Go, because then one has to bring in the Go compiler for any platform just to test this tangential feature of the library.  I've been writing in Python. What might work is a pure python git module that can be imported with an http_archive. 

A side concern is that I would someday like to switch Bazel to use these tools rather than the shell scripts we have to gather logs. There will be objections to bringing in new repositories to third_party. Without arguing that the objections are valid or not, the fewer dependencies on other repos I take the better.

PR #234, opened 2020-09-24
start a changelog. synthesize history back a few versions

aiuto:
I have code with does this in another pending change, but it does not have enough unit tests yet.

But it is incredibly hard to unit test, because it has to run git commands to get the stuff.
I am debating trying to merge that early to build out features, then figure out how to test it with
a pre-created git repo for the tests.

Part of #228

### DIFF
--- a/pkg/CHANGELOG.md
+++ b/pkg/CHANGELOG.md
@@ -1,0 +1,199 @@
+# Release 0.2.6
+
+**New Features**
+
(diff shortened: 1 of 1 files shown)


dannysullivan (approved):
LGTM - merging the generator before the tests are finished SGTM if you feel like the tests can realistically get added shortly afterwards.

PR #236, opened 2020-09-24
Use symlink actions for RPM packages instead of custom actions

nacl:
Like #232, but in the RPM builders.  No other callouts to `ln(1)` are made
directly by `rules_pkg`.


aiuto (approved):

aiuto:
Thanks.  I'll merge right now

PR #237, opened 2020-09-25
bump version and add a note about required bazel version

### LINKED ISSUE #235: Bazel version compatibility should be documented
The change in #232 broke compatibility with Bazel 3.1.0 and older due to the use of `ctx.actions.symlink`: the [`target_file` argument was not present in 3.1.0](https://docs.bazel.build/versions/3.1.0/skylark/lib/actions.html#symlink).

This is fine and all, but we should advertise a minimum compatible version to limit user potential confusion.

aiuto:
Fixes #235 

Well, at least a little.  There should be more formal ways to do this.

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -10,6 +10,12 @@
   </ul>
 </div>
 
+
+## Release Notes
+
+Version 0.3.0 requires Bazel 3.2.0 or above.
+
+
 ## Overview
 
 These build rules are used for building various packaging such as tarball
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -13,7 +13,7 @@
 
 ## Release Notes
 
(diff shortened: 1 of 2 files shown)


pkg/README.md
```diff
@@ -10,6 +10,12 @@
   </ul>
 </div>
 
+
+## Release Notes
+
+Version 0.3.0 requires Bazel 3.2.0 or above.
```
nacl, line 16: This could perhaps say "Version 0.3.0 or later" to be maximally clear, but this is good as-is, though.

nacl (approved):
Yay 0.3.0!

It would indeed be nice if this could be a little more formalized, but I don't think there's a way to do so.

pkg/README.md
aiuto: Done

PR #240, opened 2020-09-29
experimental/rpm.bzl: Make compression settings configurable

### LINKED ISSUE #239: RPM compression schemes should be configurable
`rpmbuild(8)` allows payload compression to be customized with the `_binary_payload` macro.

When packages reach sufficient size, compression schemes can very much be tuned with dramatic effects on build times: release builds may be want to use slower compression with higher ratios, and developer builds may want to use faster compression with lower ratios.

A rule interface would be very nice and `select`able.  Alternatives would be to start injecting custom macros into the RPM spec file and having `rpmbuild` evaluate them instead of the Bazel build.

nacl:
This change adds a new attribute `binary_payload_compression` to the
experimental `pkg_rpm` rule.  It is implemented in terms of command-line options
to `rpmbuild(8)`, which are passed in via `make_rpm.py`.

We do this by instructing `rpmbuild` to define the `_binary_payload` macro on its command line.  Tests are also provided.

This can also be added to the non-experimental `pkg_rpm` rule in the future; the
code would be similar.

Fixes #239.


aiuto (approved):
I think this looks good.
I just mailed a PR to split out RPM tests so that, for PRs like this, you won't have to muck around with .bazelci. We should coordinate on the order of these two.

nacl:
> I think this looks good.
> I just mailed a PR to split out RPM tests so that, for PRs like this, you won't have to muck around with .bazelci. We should coordinate on the order of these two.

For reference, that PR is #242.

How about this one goes in first? Honestly, it's not too big of a deal either way.

aiuto:
Whoops. I submitting mine first.

nacl:
> Whoops. I submitting mine first.

NP.  Lemme fix this up.

nacl:
After discussing this change with my colleagues, we've agreed that we should simplify the feature somewhat -- we believe it is adequate to directly override the compression control (`%_binary_payload`) rather than introduce additional substitution logic in the template files.

I should have this ready tomorrow.

If such advanced features are required, then we can introduce a RPM macro override interface.

aiuto:
SGTM.  You're the RPM expert.

nacl:
> After discussing this change with my colleagues, we've agreed that we should simplify the feature somewhat -- we believe it is adequate to directly override the compression control (`%_binary_payload`) rather than introduce additional substitution logic in the template files.

This is now done.  @aiuto, PTAL, but please don't merge it in; I need to tweak the original commit message in the squashed commit, but don't want to lose the review's history.

Should be a quick one.  Thanks!

pkg/experimental/rpm.bzl
```diff
@@ -722,12 +722,14 @@ pkg_rpm = rule(
             Must be a form that `rpmbuild(8)` knows how to process.  Current known
             options depend on the `rpmbuild` binary, and may include:
 
+            - "w0.ufdio" (no compression)
```
aiuto, line 725: Does it really help to have this list here?
It seems like you have to look at the rpmbuild manual page to understand which you would want,
and you are playing catch-up with future additions.

aiuto (approved):

pkg/experimental/rpm.bzl
nacl: Strictly speaking, no.  Unfortunately, this feature seems to somewhat undocumented what exists of it seems to be hidden beneath some other code.

I'll just link to some example in the RPM source code and leave some breadcrumbs saying where users can find it on their systems.

PR #243, opened 2020-09-30
Refactor docs as a first step to extending them.

aiuto:
Refactor docs so
- it is easier to regenerate the reference and drop in place
- we have a place to hang examples (left a placeholder)

Note: examples is not under docs because I want it to be more visible when looking at the source code for the repository.


README.md
```diff
@@ -15,14 +15,10 @@ CI:
...
 
-We are currently in the process of migrating the
-[Bazel packaging rules](https://docs.bazel.build/versions/master/be/pkg.html)
-from Bazel to this repo. Tracking issue:
-https://github.com/bazelbuild/bazel/issues/8489.
-
 Bazel now uses this rule set for packaging it's distribution. It still contains
-a vestigial early version of these rules, but those will be removed in a future
-release.
+a vestigial early version of pkg_tar, but that will be removed in a future
```
nacl, line 19: Nit: consider quoting pkg_tar like `` `pkg_tar` ``.

README.md
```diff
@@ -15,14 +15,10 @@ CI:
 *   [pkg](https://github.com/bazelbuild/rules_pkg/tree/master/pkg) - Rules for
     building packages of various types.
 
-We are currently in the process of migrating the
-[Bazel packaging rules](https://docs.bazel.build/versions/master/be/pkg.html)
-from Bazel to this repo. Tracking issue:
-https://github.com/bazelbuild/bazel/issues/8489.
-
 Bazel now uses this rule set for packaging it's distribution. It still contains
```
nacl, line 18: Nit: "its"

pkg/docs/reference.md
```diff
@@ -0,0 +1,582 @@
...
+      <td><code>extension</code></td>
+      <td>
+        <code>String, default to 'tar'</code>
+        <p>
+            The extension for the resulting tarball. The output
+            file will be '<i>name</i>.<i>extension</i>'. This extension
+            also decide on the compression: if set to <code>tar.gz</code>
+            or <code>tgz</code> then gzip compression will be used and
+            if set to <code>tar.bz2</code> or <code>tar.bzip2</code> then
+            bzip2 compression will be used.
```
nacl, line 52: It would be nice if the formatting for this could use more markdown rather than HTML, but that can be left for a different day.

pkg/examples/readme.md
```diff
@@ -0,0 +1,3 @@
+# rules_pkg - Examples
```
nacl, line 1: Consider naming the file `README.md` to match conventions.

Another alternative for this would be to use the github wiki feature for examples -- might make it easier to contribute.  WDYT?

pkg/README.md
```diff
@@ -1,28 +1,24 @@
...
+-   [Roadmap](#roadmap)
+-   [Reference](docs/reference.md)
+-   [Examples](examples/readme.md)
 
+<a name="notes"></a>
 ## Release Notes
 
 Version 0.3.0 or later requires Bazel 3.2.0 or later.
 
-
```
nacl, line 18: Perhaps instead of (or in addition to) this, we could have a "requirements" section that mentions what tools need to be present?  `rpmbuild` for `pkg_rpm` comes to mind.

If you don't want to do this now, I can add one after this is committed.

pkg/docs/reference.md
```diff
@@ -0,0 +1,582 @@
+# rules_pkg - reference
```
nacl, line 1: I wonder if this reference should be split up into multiple files as to perhaps be easier to navigate.  No need to do this now, though.

nacl (approved):
Cool.  All comments are nits or things we could do later on.

> Note: examples is not under docs because I want it to be more visible when looking at the source code for the repository.

Sounds good to me.

pkg/docs/reference.md
aiuto: I *think* stardoc can emit markdown, but I am not sure. In any case, I wanted to make this PR easier to review but just splitting the file and not modifying so much.

pkg/README.md
aiuto: Great idea. I just don't know how to say it now. Let's do a PR just for that, because there are so many ways to slice and dice requirements which only impact some rules and the matrix of rules version + bazel version. This, in turn, becomes interesting if we version the rules with branches, so the matrix vanishes into branched sources.

pkg/docs/reference.md
aiuto: I am not sure. I think it depends on how we want to present the common attributes. Do we pull them out to a first section, in which case I would keep them together, or copy them to each rule. For that distinct pages might be better.

PR #246, opened 2020-10-07
Provide providers for new packaging rule interfaces

nacl:
This change provides a set of three providers that can be used to represent
actions, and one that can represent a group of related actions.

These are intended to be used in rules like the `pkg_filegroup` interfaces used
in pkg/experimental, but with more flexibility and standardization.

`buildifier` was also applied opportunistically.

These are based on the comments discussed in our call earlier this week.

NOTE: there are still some points I would like to iron out, marked as TODO's in the code.  Nothing too terribly complicated, but they do impact the "meaning" behind some of the fields.

### DIFF
--- a/pkg/providers.bzl
+++ b/pkg/providers.bzl
@@ -77,9 +77,5 @@ PackageFilegroupInfo = provider(
         "pkg_files": "list of child PackageFilesInfo providers",
         "pkg_dirs": "list of child PackageDirInfo providers",
         "pkg_symlinks": "list of child PackageSymlinkInfo providers",
-
-        # TODO: does this really belong here?  Seems like the rerooting should
-        # just be done by the rule that creates this provider.
-        "prefix": """string: Base path of all the destinations in the 'children' attribute""",
     },
 )
--- a/pkg/providers.bzl
+++ b/pkg/providers.bzl
@@ -74,7 +74,9 @@ PackageSymlinkInfo = provider(
 PackageFilegroupInfo = provider(
     doc = """Provider representing a collection of related packaging providers""",
     fields = {
-        "children": """list of child providers, one of PackageFilesInfo, PackageDirInfo, or PackageSymlinkInfo""",
+        "pkg_files": "list of child PackageFilesInfo providers",
+        "pkg_dirs": "list of child PackageDirInfo providers",
+        "pkg_symlinks": "list of child PackageSymlinkInfo providers",
 
         # TODO: does this really belong here?  Seems like the rerooting should
         # just be done by the rule that creates this provider.
--- a/pkg/providers.bzl
+++ b/pkg/providers.bzl
@@ -14,12 +14,11 @@
 
 """Packaging related providers."""
 
-
 PackageArtifactInfo = provider(
     doc = """Metadata about a package artifact.""",
     fields = {
-        "label": "Label which produced it",
         "file_name": "The file name of the artifact.",
+        "label": "Label which produced it",
     },
 )
 
@@ -27,5 +26,71 @@ PackageVariablesInfo = provider(
     doc = """Variables which may be substituted into package names and content.""",
     fields = {
         "values": "Dict of name/value pairs",
+    },
+)
+
+# For the below, attributes always look something like this:
+#
+# ```
+#   {"unix": ["0755", "root", "root"]}
+# ```
+#
+
+# TODO: should all of these be singluar or plural?
+#
+# Seems like there's no harm in specifying a plural, even in rules: it would
+# allow some degree of simplification of rules without needing to specify a
+# custom manifest.
+
+# TODO: should we note exceptions from groups of files/labels?  I feel that
+# information is best encoded at the rule level, which seems to necessitate a
+# "plural" structure as defined below.
+PackageFilesInfo = provider(
+    doc = """Provider representing the installation of one or more files to destination with attributes""",
+    fields = {
+        "attributes": """dict of string -> string list: Attributes to apply to installed file(s)""",
+        "source_dest_map": """dict of File -> string: source-destination map""",
+    },
+)
+
+PackageDirInfo = provider(
+    doc = """Provider representing the creation of a single directory in a package""",
+    fields = {
+        "attributes": """dict of string -> string list: Attributes to apply to created directory""",
+        "destination": """string: Installed directory name""",
+    },
+)
+
+PackageSymlinkInfo = provider(
+    doc = """Provider representing the creation of a single symbolic link in a package""",
+    fields = {
+        "attributes": """dict of string -> string list: Attributes to apply to created symlink""",
+        "destination": """string: Filesystem link 'name'""",
+        "source": """string: Filesystem link 'target'""",
+    },
+)
+
+# Grouping provider: the only one that needs to be consumed by packaging (or
+# other) rules that materialize paths.
+
+PackageFilegroupInfo = provider(
+    doc = """Provider representing a collection of related packaging providers""",
+    fields = {
+        "children": """list of child providers, one of PackageFilesInfo, PackageDirInfo, or PackageSymlinkInfo""",
+
+        # TODO: does this really belong here?  Seems like the rerooting should
+        # just be done by the rule that creates this provider.
+        "prefix": """string: Base path of all the destinations in the 'children' attribute""",
+
+        # TODO: I'm increasingly unsure of where "section" should go.  From a
+        # high-level view, it makes sense that a "section" applies to a grouping
+        # of files, but it's hard to tell what a "section" actually is.
+        #
+        # For example, one can imagine a section of "documentation", or
+        # "config," files but then each "config" file may have different
+        # properties, but are still "config" files.
+        #
+        # Feels clumsy to me.
+        "section": """string: related file properties, such as "config", "config(noreplace)", "config(missingok)", "doc" """,
     },
 )
--- a/pkg/providers.bzl
+++ b/pkg/providers.bzl
@@ -36,26 +36,25 @@ PackageVariablesInfo = provider(
 # ```
 #
 
-# TODO: should all of these be singluar or plural?
-#
-# Seems like there's no harm in specifying a plural, even in rules: it would
-# allow some degree of simplification of rules without needing to specify a
-# custom manifest.
 PackageFilesInfo = provider(
     doc = """Provider representing the installation of one or more files to destination with attributes""",
     fields = {
         "attributes": """dict of string -> string list: Attributes to apply to installed file(s)""",
-        # TODO(nacl): determine what types actually should represent the sources here.  Files, or Labels?
-        # TODO(nacl): Should this be a map of destination -> source instead?
-        "source_dest_map": """Map of file sources to destinations""",
+        # TODO(nacl): determine what types actually should represent the sources
+        # here.  Files, or Labels?
+
+        # This is a mapping of destinations to sources to allow for the same
+        # target to be installed to multiple locations within a package within a
+        # single provider.
+        "dest_src_map": """Map of file destinations to sources""",
     },
 )
 
-PackageDirInfo = provider(
-    doc = """Provider representing the creation of a single directory in a package""",
+PackageDirsInfo = provider(
+    doc = """Provider representing the creation of one or more directories in a package""",
     fields = {
-        "attributes": """dict of string -> string list: Attributes to apply to created directory""",
-        "destination": """string: Installed directory name""",
+        "attributes": """dict of string -> string list: Attributes to apply to created directories""",
+        "dirs": """string list: installed directory names""",
     },
 )
 
@@ -64,13 +63,12 @@ PackageSymlinkInfo = provider(
     fields = {
         "attributes": """dict of string -> string list: Attributes to apply to created symlink""",
         "destination": """string: Filesystem link 'name'""",
-        "source": """string: Filesystem link 'target'""",
+        "source": """string or Label: Filesystem link 'target'""",
     },
 )
 
 # Grouping provider: the only one that needs to be consumed by packaging (or
 # other) rules that materialize paths.
-
 PackageFilegroupInfo = provider(
     doc = """Provider representing a collection of related packaging providers""",
     fields = {
--- a/pkg/providers.bzl
+++ b/pkg/providers.bzl
@@ -41,15 +41,13 @@ PackageVariablesInfo = provider(
 # Seems like there's no harm in specifying a plural, even in rules: it would
 # allow some degree of simplification of rules without needing to specify a
 # custom manifest.
-
-# TODO: should we note exceptions from groups of files/labels?  I feel that
-# information is best encoded at the rule level, which seems to necessitate a
-# "plural" structure as defined below.
 PackageFilesInfo = provider(
     doc = """Provider representing the installation of one or more files to destination with attributes""",
     fields = {
         "attributes": """dict of string -> string list: Attributes to apply to installed file(s)""",
-        "source_dest_map": """dict of File -> string: source-destination map""",
+        # TODO(nacl): determine what types actually should represent the sources here.  Files, or Labels?
+        # TODO(nacl): Should this be a map of destination -> source instead?
+        "source_dest_map": """Map of file sources to destinations""",
     },
 )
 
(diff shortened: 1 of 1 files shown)


pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
         "values": "Dict of name/value pairs",
     },
 )
+
+# For the below, attributes always look something like this:
+#
+# ```
+#   {"unix": ["0755", "root", "root"]}
+# ```
+#
```
aiuto, line 37: I think it is important to be clear hear that destination is always an absolute path in the final archive, and that rules creating pkg_* do not prepend in the package path. That makes the arithmetic clear.  if you manifest is

== foo/bar/BUILD
manifest = [
  ":my_bin",    "usr/bin/my_bin"
  "glob(["data/***",  "/usr/share/my_bin/lib"
]

that the dest paths will be   /usr/bin/my_bin, /usr/share/my_bin/lib/data/...
And that if you put it in pkg_rpm, the files will be installed exactly there.
Eventually, pkg_filegroup wil have ways to reroot the entities.

pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
...
+# other) rules that materialize paths.
+
+PackageFilegroupInfo = provider(
+    doc = """Provider representing a collection of related packaging providers""",
+    fields = {
+        "children": """list of child providers, one of PackageFilesInfo, PackageDirInfo, or PackageSymlinkInfo""",
+
+        # TODO: does this really belong here?  Seems like the rerooting should
+        # just be done by the rule that creates this provider.
+        "prefix": """string: Base path of all the destinations in the 'children' attribute""",
```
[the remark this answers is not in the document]
nacl: Sounds good to me.

pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
...
+# custom manifest.
+
+# TODO: should we note exceptions from groups of files/labels?  I feel that
+# information is best encoded at the rule level, which seems to necessitate a
+# "plural" structure as defined below.
+PackageFilesInfo = provider(
+    doc = """Provider representing the installation of one or more files to destination with attributes""",
+    fields = {
+        "attributes": """dict of string -> string list: Attributes to apply to installed file(s)""",
+        "source_dest_map": """dict of File -> string: source-destination map""",
```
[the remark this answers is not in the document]
nacl: That's what I thought initially, but then we need the ability to create exclusion lists somewhere outside of `pkg_filegroup` as it is now.  If this is `File`s, then we can do everything in pkg_filegroup [like we're doing now](https://github.com/bazelbuild/rules_pkg/blob/36e8363031c761525a6a917cf7ecb84f6ccb32eb/pkg/experimental/pkg_filegroup.bzl#L457-L465).

If the existing interface is not desirable, what would be the best way to accomplish file exclusion?

I explained some of my reasoning for this [inline](https://github.com/bazelbuild/rules_pkg/pull/246/files/be5badadb629c9c24a5dd772f1f518a7b6c7e8fd#diff-b8ef3be79dde93ffccffae703ab11150f6ab8e887345e165dfc2db91c042981cR39-R47).

WDYT?

pkg/providers.bzl
nacl: To confirm, the rules that make the non `PackageFileGroupInfo` providers should not allow the construction of new paths?  That sounds good to me.

Does this affect subtraction at all, a la `strip_prefix`?

pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
...
+        # just be done by the rule that creates this provider.
+        "prefix": """string: Base path of all the destinations in the 'children' attribute""",
+
+        # TODO: I'm increasingly unsure of where "section" should go.  From a
+        # high-level view, it makes sense that a "section" applies to a grouping
+        # of files, but it's hard to tell what a "section" actually is.
+        #
+        # For example, one can imagine a section of "documentation", or
+        # "config," files but then each "config" file may have different
+        # properties, but are still "config" files.
```
nacl, line 91: I'd appreciate a comment on this particular comment.  The concern is, for e.g. rpm, there would need to be a distinct `pkg_filegroup` for `config`, and `config(noreplace)`, which seems really clumsy to me.

pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
...
+# ```
+#   {"unix": ["0755", "root", "root"]}
+# ```
+#
+
+# TODO: should all of these be singluar or plural?
+#
+# Seems like there's no harm in specifying a plural, even in rules: it would
+# allow some degree of simplification of rules without needing to specify a
+# custom manifest.
```
nacl, line 43: I like this idea myself.  WDYT?

pkg/providers.bzl
aiuto: I'm not sure what you mean by should not allow the construction of new paths.

We have to think about what we want for strip_prefix.  The only thing I know
for certain is that it should apply to incoming deps and change the final
destination paths in the output. But, there is so much room for interpretation
in that. For example
- what happens if all the files in the deps do not have the same prefix? Is that an error, or do only some files strip the prefix?
- do we even need strip_prefix at all? Isn't that just a degenerate case of regex path rewriting?

pkg/providers.bzl
aiuto: I actually don't understand. I don't see how the plural-ness of the name impacts the rule. Or maybe I misunderstand the point entirely.

pkg/providers.bzl
```diff
@@ -29,3 +28,69 @@ PackageVariablesInfo = provider(
...
+    },
+)
+
+# Grouping provider: the only one that needs to be consumed by packaging (or
+# other) rules that materialize paths.
+
+PackageFilegroupInfo = provider(
+    doc = """Provider representing a collection of related packaging providers""",
+    fields = {
+        "children": """list of child providers, one of PackageFilesInfo, PackageDirInfo, or PackageSymlinkInfo""",
```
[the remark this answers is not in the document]
nacl: I thought I put a comment in response to this; apparently I missed it.

Anyway, I think this is a reasonable idea, given that we aren't going to be adding new providers particularly often.  The only missing one I see is device nodes and maybe registry entries.

Do we need to think of any additional customization points here?

pkg/providers.bzl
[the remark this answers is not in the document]
nacl: > Wouldn't the rule producing this provider have the exclusion?

Yes.  The immediate idea is that the producing rule would process the dependencies and produce a source-destination mapping in `PackageFilesInfo` that excludes the excluded labels.  This is more or less clear to me.

Thinking on the `Label` vs `File` question: my understanding is that the proper way to access the provider would be something like this:

```python
for dep in ctx.attr.deps:
    pfi = dep[PackageFilesInfo]
    for src, dest in pfi.source_dest_map:
        # Schedule action for copying src to dest
```

The question is thus: how we figure out where `src` is?

If `src` is a `Label`, it's not immediately clear: the way seems to be to use a [`Target`](https://docs.bazel.build/versions/master/skylark/lib/Target.html) object, however, I couldn't find a way to make a `Target` object from a `Label`.  Also, since `Label`s can ultimately point to more than one location, this further complicates the packaging rules.

If `src` is a `File`, we do not need to do any additional computations in the packaging rules: we just schedule a copy from `src` to `dest` by using the existing path information.

Since we need to get the `File`s anyway, it seems to be easier to just pass them along in the provider and let the packaging rules sort them out than it would be to try have them process `Label`s.  Thus, I chose a map of `File` -> destination here.

Does this make sense?  Am I missing something here?

pkg/providers.bzl
nacl: > I'm not sure what you mean by "should not allow the construction of new paths".

Now that I think about it, that was perhaps poorly worded.

I wonder, though, is the overall question rule implementation detaisl?  AFAICT all these providers need to do is inform package builders of how packages should be structured.  Any sort of path reconstruction metadata seems to be out-of-scope for a `Provider` interface.

---

> We have to think about what we want for strip_prefix. The only thing I know
for certain is that it should apply to incoming deps and change the final
destination paths in the output.

Agreed

> - what happens if all the files in the deps do not have the same prefix? Is that an error, or do only some files strip the prefix?
 
The existing `strip_prefix` behavior in the `pkg_filegroup` ignores prefixes that don't match.  I wouldn't mind changing that behavior to failure.

> - do we even need strip_prefix at all? Isn't that just a degenerate case of regex path rewriting?

It is, but we cannot use regexps in starlark directly.  We would have to constrain the syntax significantly and then write a parser, and then that just devolves into a mess.

The other option would be to defer the path reconstruction to the packaging step.  I wouldn't be particularly opposed to this, but since the regexp processing would be handled by external tooling, this could lead to additional confusion.

I feel that regexp substitution would would only complicate matters and should not be added in.

pkg/providers.bzl
nacl: The singular vs plural issue does indeed impact the rule interface.  If, for example, the directory provider is singular, then you need to have to rules like this:

```python
pkg_mkdir(name = 'a', dir_name = 'a')
pkg_mkdir(name = 'b', dir_name = 'b')
```

Since `PackageDirectoryInfo` can only contain information on one directory.   A rule that creates multiple directories simultaneously would have to return a `PackageFilegroupInfo` provider in this scheme.

If we have a plural `PackageDirectoriesInfo` provider, then a rule like

```python
pkg_mkdirs(name = 'dirs', dirs = ['a', 'b'])
```

would return one `PackageDirectoriesInfo` provider, which can then be included in a `pkg_filegroup` (or equivalent) rule.

I'm in favor of this because it allows for rules to be specified more cleanly.  Does this make sense?

pkg/providers.bzl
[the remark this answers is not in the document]
aiuto: I don't think so at this time. This is greenfield design, so there are no users. If we find we need more, we can add them in the future.

pkg/providers.bzl
aiuto: Thanks. I understand now. 
Do you think it save enough space in practice to justify this one being different?

The most frequent time I have wanted a mkdir is if I want to explicitly set the attributes of a folder that would otherwise be created with a default.
```
pkg_filegroup(
   name=docs,
   srcs=glob([docs/**]) + ["docs_dir"],
   strip_prefix='docs', path='/usr/share/MY_APP/docs',
   attrs = '0444',
)

pkg_mkdir(name="docs_dir", dir="/usr/share/MY_APP/docs", attrs="0555", owner='man', group='man')
```

So, I would not be creating many directories at once, but rather one per logical group of files.
Do you have any real-world use cases that I am missing?

pkg/providers.bzl
aiuto: That sounds reasonable.  Let's drop 'secton'

pkg/providers.bzl
aiuto: Rereading your words "all of these"   Did you imply for anything else beyond,  mkdir and symlinks?
FWIW, my thoughts on mkdir would also apply to symlink.  I would want a pkg_filegroup that represents installing the binary in /usr/share/MY_APP/bin  and creating a symlink from /usr/bin/MY_APP to /usr/share/MY_APP/bin. Then I can talk about the pair as a nice unit.

pkg/providers.bzl
aiuto: Sounds good
- strip_prefix with a non-match is a fail.
- no regex

With those simplifications, we could defer remappings until actual packaging. By that I mean if we had
```
pkg_filegroup(  strip=x, prefix=y, srcs = pkg_filegroup(srcs=[...], prefix=x))
```
we do not have to actually modify the list of paths from the inner group. We could just pass the right things up so that the final packager rule can apply both transforms in order. 

I don't think we have to settle that today.  Let's get out a version that can do the basic mkdir and symlink (which has pent up feature requests) and when we try to implement nice path remappings, we can refine.

pkg/providers.bzl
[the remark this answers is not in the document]
aiuto: I think we need labels in a depset so that they can be inputs to the pkg_rpm or pkg_tar.
I'm happy to defer the words until we try to implement some of these rules.

aiuto:
I think we should talk about overall plan.

I'm happy enough with most of this to use it to start implementing a few simple things.
If we had mkdir and symlink, I would slip those into pkg_tar and zip just to prove them out.
I find that actually building the code shakes out the design issues.

pkg/providers.bzl
nacl: First comment first:

> So, I would not be creating many directories at once, but rather one per logical group of files.
> Do you have any real-world use cases that I am missing?

The only cases would perhaps be when you have to apply custom permissions to entire custom directory structures.  Doing grouping like this would save users having to write many `pkg_mkdir` rules scattered about the BUILD file, and instead have a cluster of centralized rules that specifies custom permissions/ownership.

```
pkg_mkdirs(
    name = "core_dirs",
    dirs = [
        "/opt/MYAPP",
        "/opt/MYAPP/bin",
        "/opt/MYAPP/etc",
        "/opt/MYAPP/etc/MYAPP.d",
        ...
    ],
    attrs = {"unix": ("0755", "MYAPP", "MYAPP")},
)
```

FWIW, I do like the idea of pairing, though, but it looks like it might introduce a lot of boilerplate.  If the providers are plural, it would give flexibility in specifying targets and would not preclude the pairing pattern.

I vote for making providers plural if only to support additional flexibility for rule authors.

> Rereading your words "all of these" Did you imply for anything else beyond, mkdir and symlinks?

I was thinking it would apply to everything, now and in the future.

> FWIW, my thoughts on mkdir would also apply to symlink. I would want a pkg_filegroup that represents installing the binary in /usr/share/MY_APP/bin and creating a symlink from /usr/bin/MY_APP to /usr/share/MY_APP/bin. Then I can talk about the pair as a nice unit.

I like the sound of this too.

---

Ultimately, the question is about the providers, and I don't think we should be creating such restrictions.  We could, perhaps, create rules that only create one directory or one symlink, and have the providers only contain a single item in their corresponding list/dict.

pkg/providers.bzl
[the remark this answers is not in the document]
nacl: That's fine with me.  I'll update the documentation string accordingly.

pkg/providers.bzl
nacl: Agreed.

pkg/providers.bzl
nacl: Will do.

pkg/providers.bzl
nacl: Agreed.

> I think we should talk about overall plan.
> 
> I'm happy enough with most of this to use it to start implementing a few simple things.
> If we had mkdir and symlink, I would slip those into pkg_tar and zip just to prove them out.
> I find that actually building the code shakes out the design issues.

Sounds good to me.  I'll undraftify it now.

The only real outstanding question for me is the singular/plural thing, but we can do that in an additional review as we try this out, if you'd like.

nacl:
> I think we should talk about overall plan.
> 
> I'm happy enough with most of this to use it to start implementing a few simple things.
> If we had mkdir and symlink, I would slip those into pkg_tar and zip just to prove them out.
> I find that actually building the code shakes out the design issues.

Sounds good to me. I'll undraftify it now.

The only real outstanding question for me is the singular/plural thing, but we can do that in an additional review as we try this out, if you'd like.

pkg/providers.bzl
aiuto: We want filegroup to be able to glob() files, so that has to have a provider with a list of files. We can't glob() symlinks and directories, because they do not exist in the source tree.  So I think of them differently, but your example is a good case for mkdirs.

OTOH, your last example raised a new design question.
Am I allowed to mix pkg_mkdir destinations with implicit directories from a pkg_filegroup? 
For example, I add this:
```
pkg_filegroup(
   name = "core_bin",
   srcs = [":my_bin"],
   prefix = "/opt/MYAPP/bin",
)

pkg_tar(
   name = "MYAPP.tgz",
   srcs = [
      ":core_bin",
      ":core_dirs",
   ]
)
```

Do we intend that tar should sort the destinations and do the right thing? I believe the answer is yes.
And also, that if `core_bin` also defines the output path /opt/MYAPP/bin explicitly, that is an error.

nacl: > We want filegroup to be able to glob() files, so that has to have a provider with a list of files. We can't glob() symlinks and directories, because they do not exist in the source tree. So I think of them differently, but your example is a good case for mkdirs.

Another way to think about this would be to describe directories and symlinks can as "data", and adding a rule to adding multiple data entries at the same time would make sense.

WDYT?  Again, we can finish this up in a separate change.

> Do we intend that tar should sort the destinations and do the right thing? I believe the answer is yes.

Agreed.  This was one of my plans initially, I don't think I ever got around to it.

> And also, that if core_bin also defines the output path /opt/MYAPP/bin explicitly, that is an error.

Not sure what you mean here.  Do you mean a duplicate file/directory?  If so, I also agree.

aiuto: Yes. duplicate files.
symlinks may want to be singular so it can point to a label. as discussed off this PR.

nacl: Done.  I don't think the symlink provider is quite right (I'll make another comment for that discussion), but in the name of having something mostly working and in the tree, this should be good to go.

nacl:
I thought about the symlink provider some more; it could still be "plural", but the rules interface still have to be singular if it could accept a label.

We could create a `pkg_symlink_filegroup` rule that produces a `PackageFilegroupProvider` of symlinks.

We can definitely take care of this later on.

aiuto (approved):

PR #253, opened 2020-10-29
Fix error when using 'archive_name' attribute

thii:
pkg_tar_impl doesn't have an 'archive_name' attribute, it has to be
popped out of keyword arguments here.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -349,12 +349,17 @@ def pkg_tar(name, **kwargs):
                       "This attribute was renamed to `srcs`. " +
                       "Consider renaming it in your BUILD file.")
                 kwargs["srcs"] = kwargs.pop("files")
-    archive_name = kwargs.get("archive_name") or name
+    archive_name = kwargs.pop("archive_name", default = None)
     extension = kwargs.get("extension") or "tar"
+    if archive_name:
+        if kwargs.get("package_file_name"):
+            fail("You may not set both archive_name and package_file_name")
+        print("archive_name is deprecated. Use package_file_name instead.")
+        kwargs["package_file_name"] = archive_name + "." + extension
     pkg_tar_impl(
         name = name,
-        out = kwargs.get("out") or (archive_name + "." + extension),
-        **kwargs,
+        out = kwargs.get("out") or (name + "." + extension),
+        **kwargs
     )
 
 # A rule for creating a deb file, see README.md


thii:
cc @davschne who originally landed the change.

davschne:
Seems reasonable. Does submitting a dictionary with the unrecognized key to `pkg_zip_impl` via `**kwargs` raise an error, then?

thii:
```starlark
pkg_zip(
    name = "test_zip_empty",
    archive_name = "foo",
    srcs = [],
)
```

```console
$ bazel build :test_zip_empty
ERROR: /BUILD:141:8: //:test_zip_empty: no such attribute 'archive_name' in 'pkg_zip_impl' rule
```

davschne:
Understood. I endorse this PR.

thii:
Updated to fix pkg_deb and pkg_tar too.

As a side note, it seems the current implementation also disallows the use the `out` attribute, because of a similar error.

aiuto (commented):
What I would rather see is the removal of archive_name and porting package_file_name to zip, deb and rpm.

This would be a short term fix until that happens, but for pkg_tar, we should drop the archive_name use entirely, and just pass name+extension in as the out.

thii:
Updated to remove it.

davschne:
No, wait; my team is depending on that attribute! If removed, it would need to be replaced with equivalent functionality, e.g. the package_file_name attribute @aiuto mentioned.

thii:
> No, wait; my team is depending on that attribute!

Does it work for you? I'm getting the `no such attribute 'archive_name' ` error when I tried to use it.

> If removed, it would need to be replaced with equivalent functionality

The `out` attribute provides an equivalent functionality.

davschne:
Okay, sure, `out` provides equivalent functionality, but it's a breaking change to the API just to rename an attribute for no apparent reason.

Yes, it is working for us, and it appears to be working in the automated testing CI for this repo. I wonder if you're using a different version of Bazel or if there's some other configuration difference. Again, I'm in favor of the original change you made, especially if that fixes the error you're encountering. I do however object to the removal or renaming of the attribute unless there's a strong reason for it, for instance the introduction of a more flexible mechanism like `package_file_name`.

thii:
I’m not strongly tied to this change. We would love to have a way to customize the archive name too.

This is _not_ a rename—`out` is an existing outputs attribute. I was confused at first when there were two ways to specify custom archive name, and none of them works. (I’m using the latest release of Bazel (3.7.0), without any custom configuration)

I couldn’t find a test for `archive_name` in the automated CI tests here. Are they downstream tests else?

aiuto:
How about this. 
- only change pkg_tar for now.  Do the others when we add package_file_name to those
- fail if archive_name and package_file_name are both used
- if archive_name is specified
  package_file_name = archive_name + "." + extension
  optional: print deprecation warning
- out should remain out.

That will be slightly different than the current behavior, in that you might get a symlink instead of a file, but it 
will have the name you expect.

```
archive_name = kwargs.pop("archive_name")
if archive_name:
   if kwargs.get('package_file_name'):
      fail("can't do both")
   print('archive_name is deprecated. use package_file_name')
   kwargs['package_file_name'] = archive_name + '.'+ extension
pkg_tar_imp(
   name=name
   **kwargs
)
```

davschne:
Sounds good to me, @aiuto.

Regarding breaking API changes, do you plan to implement semantic versioning sometime in the near future? Or, put another way, when do you plan to post the first major-version release? I think it'd be very helpful for people and organizations (and there seem to be at least a few) depending on this library to understand how new releases affect the API.

aiuto:
I'm trying to minimize backward breakages even while at 0.x.
Before 1.x I think there are a few important things that need to happen.

1. rules, if given a TAR target in `srcs` will just include the tarball as is
   - but they also have the ability to expand the tarball
      - in its entirety, with all paths preserved,
      - with any number of filters or transforms - but you may have to declare those outside the final target
2. arbitrary symlinks and directory creation (thanks to @nacl for working through design of this with me) in all rule types

I think I can do #1 without breaking anyone existing, but I want to have created a path toward deprecating automatic tar expansion and making users declare (in some way) that tar inputs get expanded or not.  
I am thinking late Q4 or early Q1 for a 1.0.0.

aiuto:
@davschne Your question reminded me that we have a discussion list. https://groups.google.com/g/rules-pkg-discuss
I will strive to remember to post on it more often.

davschne:
Cool. Thanks for the details, @aiuto. I just joined the discussion list.

thii:
Thanks for the suggestion. I’ve updated the fix. My original interest was to fix pkg_zip but I will submit it in another PR.

aiuto (approved):

thii:
Friendly ping

PR #256, opened 2020-11-05
Add examples of how to name packages with build time values.

aiuto:
They could be better, but they serve to get the point across quickly.

They are also executable, as all sample code should be.


aiuto:
I forgot about this. Once Thanksgiving season hits, all non P1 work gets pushed aside until January.
Context: #193

dannysullivan (approved):

Issue #50, opened 2019-01-29
make pkg_tar's "./" mandatory prefix optional


beasleyr-vmw:
### Description of the problem / feature request:

Files embedded in a tar created by `pkg_tar` include a mandatory "./" prefix. I would like to make this prefix optional.

### Feature requests: what underlying problem are you trying to solve with this feature?

This is an interoperability/migration problem. My request incurs tech debt to finance Bazel adoption.

I'm trying to migrate a constellation of builds from Maven to Bazel. When I port project X to Bazel, I still need to support some number of preexisting downstream builds. These builds extract subsets of files from tars using whitelists which are **not** anchored with a leading "./". When fed a tar generated by Bazel's `pkg_tar`, a whitelist of "foo" doesn't match "./foo", and so the untar action is a no-op.

Rather than have to discover/fix every trivial difference downstream, I would rather start by producing Bazel bits which are drop-in replacements for Maven bits. Our internal teams can then coordinate among themselves to pay down the debt and remove use of any interop flags.

For the record, a change to remove the "./" prefix was [committed back in 2017][1] but soon [rolled back][2]. I think an _option_ to remove the prefix would help folks like myself w/o risk of breaking existing Bazel users.

[1]: https://github.com/bazelbuild/bazel/commit/72ab9322ac4ea6b743bc304b7b5d05f8b5b56239#diff-002c180b932b0dbae88c2559af105850
[2]: https://github.com/bazelbuild/bazel/commit/36178dddc45ee39e302516232c83b0fbc4c5dc37#diff-002c180b932b0dbae88c2559af105850

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

n/a

### What operating system are you running Bazel on?

Linux - CentOS 6.6 and 7.2
macOS - High Sierra (I think)

### What's the output of `bazel info release`?

"development version"

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

We track Bazel's Git repo and apply a small number of patches atop official releases. We're currently building against 0.21.0.

### What's the output of `git remote get-url origin ; git rev-parse master ; git rev-parse HEAD` ?

n/a

###  Have you found anything relevant by searching the web?

Yes -- https://github.com/bazelbuild/bazel/commits/master/tools/build_defs/pkg/archive.py:
[...]

PR #261, opened 2020-12-07
Fix issue #50 by adding package_base, but without changing current behaviour

dmayle:
As mentioned in the title, this is a fix for #50 which I've tested at this repository and correctly allows controlling the package base without changing the default.

Setting the default on line 303 to "" would allow the default to be changed, introducing breaking behaviour.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -76,6 +76,7 @@ def _pkg_tar_impl(ctx):
 
     # Start building the arguments.
     args = [
+        "--root_directory=" + ctx.attr.package_base,
         "--output=" + output_file.path,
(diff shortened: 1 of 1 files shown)


dmayle:
@googlebot I signed it!

douglas-plezentek:
@googlebot I signed it!

douglas-plezentek:
@googlebot I consent.

aiuto (approved):
LOL... That was too easy. 
I never would have suspected that the build_tar tool was so silly that it was purposefully prefixing './' to all paths.

aiuto:
Thanks for doing this.

PR #263, opened 2020-12-10
pkg_deb: allow passing deb and changes directly

amscanne:
(More detail in the commit.)

I've knocked this out quickly in anticipation of some feedback on a better way to solve this problem. Please let me know if I'm holding it wrong!

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -417,8 +417,12 @@ def pkg_deb(name, package, archive_name=None, **kwargs):
     archive_name = archive_name or name
     version = kwargs.get("version") or ""
     architecture = kwargs.get("architecture") or "all"
-    out_deb = "%s_%s_%s.deb" % (package, version, architecture)
-    out_changes = "%s_%s_%s.changes" % (package, version, architecture)
+    out_deb = kwargs.pop("deb", None)
+    if not out_deb:
+        out_deb = "%s_%s_%s.deb" % (package, version, architecture)
+    out_changes = kwargs.pop("changes", None)
+    if not out_changes:
+        out_changes = "%s_%s_%s.changes" % (package, version, architecture)
 
     pkg_deb_impl(
         name = name,
--- a/pkg/tests/BUILD
+++ b/pkg/tests/BUILD
@@ -272,14 +272,19 @@ py_test(
 
 pkg_deb(
     name = "test-deb",
+    architecture = select({
+        "//conditions:default": "all",
+    }),
     breaks = ["oldbrokenpkg"],
     built_using = "some_test_data (0.1.2)",
+    changes = "titi_test_all.changes",
     conffiles = [
         "/etc/nsswitch.conf",
         "/etc/other",
     ],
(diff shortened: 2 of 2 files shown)


amscanne:
This is actually broken in a non-trivial way, because the test depends on being able to refer to out_deb and out_changes explicitly in its runfiles. (Is this something that anyone does outside the test? I'm not sure.) I could change the test behavior, but I'd like to get guidance on this first.

Note that other people have hit this problem as well:
https://stackoverflow.com/questions/56190105/selecting-different-values-for-pkg-debs-package-and-architecture-attributes#new-answer

amscanne:
Okay, I think I misunderstood the constraints. The latest patch just lets you pass a select for the architecture, so that you can at least work around the problem. The API for pkg_deb is otherwise unchanged.

aiuto:
Can you start with an issue for this where we can come up with an overall design.  The package_file_name attribute from pkg_tar will be available to pkg_dep too. That takes care of half this PR in a uniform way across rules.
Do we need that same capabilty for changes or is there a default format for changes which we should always use?

matzipan:
AFAIU the problem is that you can't pass a select to `architecture`. And I think the proposed workaround was to pass debs and changes from the outside so this wouldn't be a blocker.

aiuto:
I think #282 makes this obsolete.   The net effect of that PR is that you can set changes explictly if you want. But if you don't, we use basename(package_file_name).changes. In turn, `package_file_name` might be built from the defaults or computed from things only known at build time (like cpu).

aiuto:
Sorry, @matzipan, I didn't actually answer your question.

I don't understand why you are saying that architecture is not select()able.  I just tried using a select for architecture, along with the defaulting for package_file_name and it just worked.  

```
pkg_deb(
    name = "deb_select_arch",
    data = ":example1",
    description = "what it does",
    maintainer = "someone@somewhere.com",
    package = "foo_b",
    version = "1",
    architecture = select({
        ":special_build": "NewCPU",
        "//conditions:default": "x36",
    }),
)


$ bazel build :deb_select_arch
Target //:deb_select_arch up-to-date:
  bazel-bin/foo_b_1_x36.deb
$ bazel build :deb_select_arch --define=SPECIAL=1
Target //:deb_select_arch up-to-date:
  bazel-bin/foo_b_1_NewCPU.deb
```

Of course, looking at the stackoverflow question, I see that it might not have been selectable before #282, but all the output file computations are made more rationally now.

matzipan:
Hi aiuto,

Thanks for the response.  I was not aware of #282, which seems to address that. I think I've seen it in other Bazel extensions as well, so it's interesting that now I know how this can be fixed (I think it was rules_docker).

Andrei

aiuto:
#282 is submitted now. This PR might be obsolete.

amscanne:
Based on your comment above, I’m happy and will close this issue. I’ll reopen if there’s something that isn’t working as expected.

Flores-21 (commented):
M

PR #267, opened 2021-01-04
Use released version of rules_python

limdor:


aiuto (approved):

graywolf-at-work:
Would be nice to get .tiny release with this change to silence the 

```
DEBUG: Rule 'rules_python' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1564776078 -0400"
```

There are workarounds but this would make it work out of the box.

aiuto:
I want a few more things in before a 4.0 release and I am not set up for cherrypick releases.

The best practice is to add a current rules_python stanza to your own WORKSPACE file. That gives you explicit control over what you get.  Otherwise, you are relying on the order you include other repositories to impact what versions of their common transitive dependencies you get.

PR #272, opened 2021-01-19
Make helpers_test work on Windows

aiuto:
We can't run the test from within a NamedTemporaryFile context because the file is locked while being written so we can not read it. Instead:
- use NamedDirectory
- create the args file within that directory
- close the file
- then check the content
- let the NamedDirectory context clean up the result


aiuto:
Yes. I do believe this fixes #146  too. Thanks for finding that.

aiuto:
Weekly, friendly ping

pkg/tests/helpers_test.py
```diff
@@ -27,20 +28,20 @@ def testStripped(self):
     self.assertEqual(helpers.GetFlagValue('value ', strip=True), 'value')
 
   def testNonStripped_fromFile(self):
-    with tempfile.NamedTemporaryFile() as f:
-      f.write(b'value ')
-      f.flush()
-      self.assertEqual(helpers.GetFlagValue(
-          '@{}'.format(f.name),
-          strip=False), 'value ')
+    with tempfile.TemporaryDirectory() as temp_d:
```
vmax, line 31: @aiuto I don't know if tests are supposed to be 2-3 cross-compatible, but FYI `TemporaryDirectory` is only present since 3.2, see [ref](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory)

vmax (approved):
LGTM :+1:

dannysullivan (approved):

PR #274, opened 2021-01-19
Create alias for distro creation rule; add copyright in distro/BUILD

nacl:
Building the distribution tarball for rules_pkg requires you to know the version
number, which isn't always known offhand.

This change creates an `alias` rule in `distro/BUILD` which makes building the
"distro" tarball as simple as naming `//distro` in a rule or `bazel`
command-line.

Additionally, a copyright statement for `distro/BUILD` was added, dated back to
the original creation of the file in this repository.


aiuto (approved):

aiuto:
Thanks

Issue #275, opened 2021-01-20
Make tests resiliant to lack of XZ compression support in the python toolchain


aiuto:
Python before 3.something does not suport lzma by default. We should detect that and limit the number of compression schemes allowed (and tested) depending on platform.

aiuto:
If anyone tries to do this: the problem is that you need to create a toolchain that auto_configures based on your python to see if it has lzma.  The payback is small.

PR #276, opened 2021-01-20
Refactor usage of xz compression so tests are less fragile across platforms.

aiuto:
Refactor usage of xz compression so that support for it is limited to one .bzl and one .py file.

Context: #275

Some Python implementations and/or build environments do not have support for xz compression.  In order to make it easier to developer and use the package on those platforms pkg.bzl now provides an indicator of support. That is used in the test cases to conditionally add tests for xz compressed tar files. The python code follows a similar pattern.

This allows someone importing the rules to change only one or two locations to change the test patterns to conform to the constriants of their environment.

Next step: Make the connection between the Starlark and Python code automatic.


dannysullivan (approved):

aiuto:
Please file a bug or send a change.

Issue #161, opened 2020-04-20
rules_pkg outputs should be usable as runfiles


nacl:
In #160, it was discovered that it is not possible to use the outputs of the experimental`pkg_rpm` rule as runfiles inputs for other rules, quite possibly due to not returning a`DefaultInfo` provider.  Given they are simple binary blobs, packaging outputs should be available for consumption in other rules this way.

This may apply to the other rules in rules_pkg too, and should be investigated and rectified if runfiles are found not to be supported.

PR #282, opened 2021-01-26
add package_file_name to pkg_deb and pkg_zip

### LINKED ISSUE #283: pkg_rpm rule definition uses deprecated 'outputs' parameter
We should stop using this deprecated attribute.

aiuto:
This unifies pkg_deb, pkg_tar, and pkg_zip so that they all have these 3 parameters
- out: an an output file name. This will end up being an actual file, or a symlink to a file with a more complex name
- package_file_name: The parameterizable name for the output file. The name can include template variables which might be calculated at analysis time, such as the C++ library version specified by the selected C++ toolchain.
- package_variable: The label of a rule that provides variables to use for substitution in package_file_name.

If package_file_name is not specified, out is the single output file. If it is specified, the default output of the target (DefaultInfo.files) will be the file named by package_file_name and out will be a symlink to it.

The examples are expanded to show how this is used in practice.

See #193 for discussion.

Also, a partial fix for #161.   runfiles now includes all the outputs of these pkg_* rules

pkg_rpm is excluded from this PR. We need to resolve #283 first.

RELNOTES:
- `pkg_deb`, `pkg_tar`, and `pkg_zip` all now use `package_file_name` and `package_variables` to provide build time configurable names for their outputs.
- `archive_name` is deprecated for all rules.  See #284


dannysullivan (approved):

examples/naming_package_files/package_upload.bzl
```diff
@@ -0,0 +1,51 @@
...
+    return
+
+debian_upload = rule(
+    implementation = _debian_upload_impl,
+    doc = """A demonstraion of consuming PackageArtifactInfo to get a file name.""",
+    attrs = {
+        "package": attr.label(
+            doc = "Package to upload",
+            mandatory = True,
+            providers = [[PackageArtifactInfo]],
```
nacl, line 40: Nit: the nested-list syntax is not necessary when there's only one required provider

examples/naming_package_files/package_upload.bzl
```diff
@@ -0,0 +1,51 @@
...
+    content = ["# Uploading %s" % package_basename]
+    for f in ctx.attr.package[DefaultInfo].default_runfiles.files.to_list():
+        if f.basename.startswith(package_basename):
+            content.append("gsutil cp %s gs://%s/%s" % (
+                f.path,
+                ctx.attr.host,
+                f.basename,
+            ))
+    ctx.actions.write(ctx.outputs.out, "\n".join(content))
+    return
```
nacl, line 31: Nit: explicit `return` is not required

pkg/distro/BUILD
```diff
@@ -97,6 +97,7 @@ genrule(
 bzl_library(
     name = "rules_pkg_lib",
     srcs = [
+        "//:internal/util.bzl",
```
nacl, line 100: So, instead of creating a `private` package, you're going for `internal`?

Nothing necessarily wrong with that; most rulesets are taking the `private` route.  I recall you saying you didn't like that as (IIRC) it can be a bit confusing to users.

pkg/distro/BUILD
```diff
@@ -36,8 +36,8 @@ pkg_tar(
     srcs = [
         ":small_workspace",
         "//:standard_package",
-        "//toolchains:standard_package",
         "//releasing:standard_package",
+        "//toolchains:standard_package",
```
nacl, line 40: Consider mentioning buildifier updates in your final message.  I see it in one of your intermediate commits.

pkg/docs/reference.md
```diff
@@ -11,12 +11,30 @@
...
+
+These attributes are used in several rules within this module.
+
+**ATTRIBUTES**
+
+| Name  | Description | Type | Mandatory | Default |
+| :------------- | :------------- | :-------------: | :-------------: | :------------- |
+| out | Name of the output file. This file will always be created and used to access the package content. If `package_file_name` is also specified, `out` will be a symlink. | String | required |  |
+| package_file_name |  The name of the file which will contain the package. The name may contain variables in the form `{var}`. The values for substitution are specified through `package_variables`.| String | optional | package type specific |
+| package_variables |  A target that provides `PackageVariablesInfo` to substitute into `package_file_name`.| <a href="https://bazel.build/docs/build-ref.html#labels">Label</a> | optional | None |
```
nacl, line 25: Could this be laid out so that the whitespace is aligned when this file is rendered as text?

Not a big deal, though.  I may do it myself for `pkg_files` and friends.

pkg/internal/util.bzl
```diff
@@ -0,0 +1,59 @@
...
+load(":providers.bzl", "PackageArtifactInfo", "PackageVariablesInfo")
+
+def setup_output_files(ctx, package_file_name=None):
+    """Handle processing of out and package_file_name.
+
+    By default, output is to the "out" attribute. If package_file_name is
+    given, we will write to that name instead, and out will be a symlink to it.
+
+    Callers should:
+       - write to ouput_file
```
nacl, line 25: Nit: `output_file`

pkg/internal/util.bzl
```diff
@@ -0,0 +1,59 @@
...
+       - return a PackageArtifactInfo provider of the form:
+            label: ctx.label.name
+            file_name: output_name
+
+    Args:
+      ctx: context
+      package_file_name: computed value for package_file_name
+
+    Returns:
+      ouputs: list(output handles)
```
nacl, line 36: Nit: `outputs`

pkg/internal/util.bzl
```diff
@@ -0,0 +1,59 @@
...
+       - add outputs to their returned DefaultInfo(files) provider
+       - return a PackageArtifactInfo provider of the form:
+            label: ctx.label.name
+            file_name: output_name
+
+    Args:
+      ctx: context
+      package_file_name: computed value for package_file_name
+
+    Returns:
```
nacl, line 35: Perhaps this should say "tuple of ...", or perhaps the return value should be a `struct`?

`struct`s don't support destructuring assignment like with tuples, but I think I'd be fine with either.

pkg/pkg.bzl
```diff
@@ -15,6 +15,7 @@
 
 load(":path.bzl", "compute_data_path", "dest_path")
 load(":providers.bzl", "PackageArtifactInfo", "PackageVariablesInfo")
+load("internal/util.bzl", "setup_output_files")
```
nacl, line 18: Prefix `"internal/util.bzl" like with the other `load`s?

pkg/pkg.bzl
```diff
@@ -187,10 +172,36 @@ def _pkg_tar_impl(ctx):
...
+
+    outputs, output_file, output_name = setup_output_files(
+        ctx,
+        package_file_name = package_file_name,
+    )
+
+    # If the user does not provide the changes file, compute it.
+    # TBD: Should the user even be allowed to provide it? That is, is there any
+    # value in ability to have the basename of the .deb and the .changes file
+    # be different?
```
nacl, line 192: Should there be a task for this?

pkg/pkg.bzl
```diff
@@ -382,7 +401,10 @@ pkg_deb_impl = rule(
     implementation = _pkg_deb_impl,
     attrs = {
         "data": attr.label(mandatory = True, allow_single_file = tar_filetype),
-        "package": attr.string(mandatory = True),
+        "package": attr.string(
+            doc = "Package name",
+            mandatory = True,
+        ),
```
nacl, line 407: I'm not a fan of this particular attribute name.  It seems like this might be confusing when compared with `package_file_name` below?

For the experimental `pkg_rpm` rule, we went with `rpm_name`, which was used in place of `name` when provided.  How about doing something similar here, like `package_name`?  It could be added to the common attributes documentation list.

This can be deferred, but should be assigned a task so that deb/rpm/$SYSTEM_PACKAGE_MANAGER have consistent naming schemes for metadata like this.

pkg/pkg.bzl
```diff
@@ -415,10 +443,16 @@ pkg_deb_impl = rule(
...
         "provides": attr.string_list(default = []),
 
-        # Outputs.
+        # Common attributes
         "out": attr.output(mandatory = True),
-        "deb": attr.output(mandatory = True),
-        "changes": attr.output(mandatory = True),
+        "package_file_name": attr.string(doc = "See Common Attributes"),
+        "package_variables": attr.label(
+            doc = "See Common Attributes",
```
nacl, line 450: Could this be a markdown hyperlink? This occurs a few places below (and maybe above, I might have missed some).

pkg/pkg.bzl
```diff
@@ -428,37 +462,35 @@ pkg_deb_impl = rule(
...
     """Creates a deb file. See pkg_deb_impl."""
-    archive_name = archive_name or name
-    version = kwargs.get("version") or ""
-    architecture = kwargs.get("architecture") or "all"
-    out_deb = "%s_%s_%s.deb" % (package, version, architecture)
-    out_changes = "%s_%s_%s.changes" % (package, version, architecture)
-
+    if archive_name:
+        # buildifier: disable=print
+        print("archive_name is deprecated. Use package_file_name or out to name the output.")
```
nacl, line 472: Consider quoting attribute names here (e.g. `'archive_name'`) to ease readability.  This occurs a few places below (and maybe above, I might have missed some).

As an aside, Starlark really needs the ability to print out warnings and nonfatal errors with the appropriate ornamentation.  Being able to tag rule attributes as "deprecated" may also be a good feature.

pkg/tests/package_naming_aggregate_test.sh
```diff
@@ -0,0 +1,10 @@
+# This test simply checks that when we create packages with package_file_name
```
nacl, line 1: Nits: Copyright?  Shebang line for `/bin/bash`?

nacl (approved):
I like it.

The comments are largely nitpicky or can be deferred.  I don't feel that I need to look at this again unless you want me to.

I'm interested in a common attribute for package names.  If that isn't done now, I'd like to see it done later.

examples/naming_package_files/BUILD
```diff
@@ -74,3 +75,30 @@ pkg_tar(
     package_file_name = "example2-{cc_cpu}-{compiler}-{compilation_mode}.tar",
     package_variables = ":toolchain_vars",
 )
+
+# Let
```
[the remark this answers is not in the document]
aiuto: Fixed.

pkg/distro/BUILD
aiuto: Will do.

examples/naming_package_files/package_upload.bzl
aiuto: Changed

pkg/distro/BUILD
aiuto: I don't recall one or the other. My quick check shows most are private. I'll change to that.

pkg/pkg.bzl
aiuto: Added #289

pkg/pkg.bzl
aiuto: I'm deferring this because it has been in use so long. package_name is probably better.

pkg/pkg.bzl
aiuto: We need a PR just to make the doc strings usable to generate docs. Right now this is unused, so it is sort of moot.
I am waiting for improvements to Stardoc.

pkg/tests/package_naming_aggregate_test.sh
aiuto: Done

aiuto: But I don't bother with the shebang. It is moot since we let the system shell win, which might b zsh on macos.

pkg/pkg.bzl
aiuto: Done.
Agreed on the Stardoc changes.

PR #285, opened 2021-01-29
Extend pkg_tar rule with stamp_mtime attribute

dennis3620:
This change extends the `pkg_tar` rule for stamp support for files' mtime. The attribute `stamp_mtime` contains a "volatile" workspace status command key which allows to set dynamic mtime for files inside tar archives. The value of the key depends on the workspace status command result but only "volatile" keys are provided as input to `pkg_tar` rule.

The motivation behind this change relates to #149 and #192. Files are build with Bazel and added as tar archive to `container_image` rule ([rules_docker](https://github.com/bazelbuild/rules_docker#container_image-1)). These files are served by nginx inside the container. This change should provide proper ETag values for nginx which relies on file's last modification date for caching.


pkg/docs/reference.md
```diff
@@ -244,6 +244,22 @@ Creates a tar file from a list of inputs.
...
+      <td>
+        <code>String, optional</code>
+        <p>
+          Stamp variable from workspace status file.
+          The key's name must not start with <code>STABLE_</code>
+          because only "volatile" keys are read.
+        </p>
+        <p>
+          <code>
+          stamp_mtime = "{STAMP_MTIME}",
```
aiuto, line 258: Why is this using a differen name than BUILD_TIMESTAMP?

aiuto (changes requested):

aiuto:
Let's discuss solutions in an issue first.  I would like to to have as many people as possible weigh in on what is useful.
There are a lot of questions
- do we change the stamp of every file?
- does that include files which came from pre-made archives that have a particular timestamp of their own?
- some people want to put the timestamp into the output filename. While I find that dubious, any stamping solution should consider it.
- Why is stamp_mtime an attribute?  Why not just support the --stamp flag. That seems to be the API of least surprise.
- and.... no solutions that just to .tar.   We need do do the appropriate thing for each of pkg_zip, pkg_deb and pkg_rpm.  But that makes getting the API right before doing any code even more important.

nacl:
Design and requirements gathering is tracked in #287.

pkg/pkg.bzl
```diff
@@ -90,12 +90,28 @@ def _pkg_tar_impl(ctx):
...
+        if ctx.attr.stamp_mtime:
+            fail("You may not set both portable_mtime and stamp_mtime")
         args.append("--mtime=portable")
+    if ctx.attr.stamp_mtime:
+        # Only ctx.version_file for file volatile-status.txt is needed
+        # because change of key/value should not invalid existing build
+        stamp_inputs += [ctx.version_file]
+        stamp_mtime_strip = ctx.attr.stamp_mtime.strip()
+        if not stamp_mtime_strip.startswith("{") or not stamp_mtime_strip.endswith("}"):
+            fail("You set stamp_mtime, but this doesn't contain a valid key variable")
```
aiuto, line 110: This should just use BUILD_STAMP.  We don't need the flexibility to let the user set an arbitrary variable out of the volatile status file. This adds complexity that no one asked for.

pkg/tests/workspace_status.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+stamp_mtime=946684741  # 1999-12-31, 23:58:01
+cat << EOF
+STAMP_MTIME ${stamp_mtime}
```
aiuto, line 4: Should be BUILD_TIMESTAMP

aiuto (commented):
I tried my own variant in #288.
It uses a similar technique, but supports --stamp directly, without having to modify the pkg_tar rule.
I prefer that because it feeds into the `stamp` behavior of the built-in rules like cc_binary. https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary

Let's finish discussion of what we want in #287 and then see what to implement.

aiuto:
I'm opting for the path I have taken in #288.  That will works the same way as the `stamp` attribute in the *_binary rules.

PR #290, opened 2021-02-05
Allow for more structured package mapping rule attributes

nacl:
This change modifies the `attributes` attribute of `pkg_files` and `pkg_mkdirs`
to allow for more structured inputs than a list of strings.

For the lack of a better option, the solution for this right now is to serialize
the data in terms of JSON strings, which will be assembled using the new
`pkg_attributes` macro.  Users should always use `pkg_attributes` for this
purpose.

The JSON is decoded within the mapping rules and passed out through their
providers.

A better option would be to pass Providers full of structured data directly into
rules, but, unfortunately Bazel does not support this at this time.

### DIFF
--- a/pkg/docs/reference.md
+++ b/pkg/docs/reference.md
@@ -23,7 +23,7 @@ These attributes are used in several rules within this module.
 | out               | Name of the output file. This file will always be created and used to access the package content. If `package_file_name` is also specified, `out` will be a symlink.            | String                                                             | required        |                                           |
 | package_file_name | The name of the file which will contain the package. The name may contain variables in the form `{var}`. The values for substitution are specified through `package_variables`. | String                                                             | optional        | package type specific                     |
 | package_variables | A target that provides `PackageVariablesInfo` to substitute into `package_file_name`.                                                                                           | <a href="https://bazel.build/docs/build-ref.html#labels">Label</a> | optional        | None                                      |
-| attributes        | Attributes to set on entities created within packages.  Not to be confused with bazel rule attributes.  See 'Mapping "Attributes"' below                                        | dict of String -> String List                                      | optional        | Varies.  See 'Mapping "Attributes"' below |
+| attributes        | Attributes to set on entities created within packages.  Not to be confused with bazel rule attributes.  See 'Mapping "Attributes"' below                                        | Undefined.                                                         | optional        | Varies.  Consult individual rule documentation for details. |
 
 See
 [examples/naming_package_files](https://github.com/bazelbuild/rules_pkg/tree/main/examples/naming_package_files)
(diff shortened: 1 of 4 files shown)


nacl:
A potential issue with this is that `buildifier` may want to sort named arguments in the order of `(group, mode, user)`.  The most readable ordering for these is `(mode, user, group)`.

We may be able to override it with additions to `buildifier`'s [tables](https://github.com/bazelbuild/buildtools/blob/master/tables/tables.go), but that may not be sufficient.

nacl:
Latest force-push updates documentation and cleaned up some noticed whitespace issues.

aiuto:
Github review is too painful to try incremental things. Let's get #273 in and then look at this.

nacl:
> Github review is too painful to try incremental things. Let's get #273 in and then look at this.

Done!

pkg/docs/reference.md
```diff
@@ -38,31 +38,27 @@ rules such as `pkg_files`, and `pkg_mkdirs`.  These allow fine-grained control
...
-This is a three-element tuple, providing the filesystem mode (as an octal
-string, like above), user, and group.
-
-`"-"` may be provided in place of any of the values in the tuple.  Packaging
-rules will set these to a default, which will vary based on the underlying
-package builder and is otherwise unspecified.  Relying upon values set by it is
-not recommended.
-
-Package mapping rules typically set this to `("-", "-", "-")`.
+There are currently no other well-known attributes known at this time.
```
aiuto, line 61: "currently .... at this time" is redundant. One or the other.

pkg/mappings.bzl
```diff
@@ -66,19 +66,40 @@ strip_prefix = struct(
...
+def pkg_attributes(mode = "0444", user = None, group = None, **kwargs):
+    """Format attributes for use in package mapping rules.
+
+    Args:
+      mode: string: UNIXy octal permissions, as a string.  Defaults to being read-only ("0444")
+      user: string: Filesystem owning user.
+      group: string: Filesystem owning group.
+      **kwargs: any other desired attributes.
+
+    Not providing any of "user", or "group" will result in the package builder
```
aiuto, line 78: Why is mode defaulted, but the rest not?  It might be valid to say
pkg_attributes(user='man') and let the builder default the mode.

pkg/mappings.bzl
```diff
@@ -249,13 +270,15 @@ pkg_files = rule(
...
+        "attributes": attr.string(
             doc = """Attributes to set on packaged files.
 
+            Always use `pkg_attributes()` to set this rule attribute.
+
             Consult the "Mapping Attributes" documentation in the rules_pkg
             reference for more details.
             """,
-            default = {"unix": ["-", "-", "-"]},
+            default = pkg_attributes(),
```
aiuto, line 281: This is subtly different from -,-,- given that mode defaults to 0444. 
If you want the default to be "let the builder pick" then this should be default="" or pkg_attributes should treat mode like owner and group.

aiuto (commented):
Really close. Just one thought, spread into two different comments.

pkg/mappings.bzl
nacl: Thanks for taking a look

Mostly just supposition so far:

- The "sane" default for package managers that produce installable package is to set user/group to an administrative user.
- If you're not producing something that's installed, user/group generally do not matter, as they get replaced with other things.

That being said, you got me thinking again about the default for `mode`.  Virtually all source archives I've encountered have files at UNIX 0644 and directories at 0755.  If you explicitly want to tune permissions to be read only for everyone, then you can do whatever you want.

So, I'm thinking to do the following:

- Set the default attributes to the empty string (or empty json, depending on what works)
- Set default mode for pkg_files to `0644`
- Within each rule, establish defaults like the ones specified in rules now, then overlay the decoded input on top of it.

I think the behavior is easy enough to understand, but it deviates from most things I've seen in bazel so far (attribute always overrides default).  I don't think this will be a problem. WDYT?

nacl: I went for a "set a sane default for mode and leave everything else alone" option, with updated documentation and tests.  PTAL.

pkg/docs/reference.md
nacl: Fixed.  Thanks.

pkg/mappings.bzl
aiuto: I would think about what happens when you do this
```
cc_binary(name='x')
pkg_file(name='files', srcs=[':x', ':y'], ...)
pkg_rpm(name='pkg', srcs=[':files'])
```
The least surprising thing is if `x` is mode 0755 at the end.  I think what you have now will allow that to happen.

aiuto (approved):

aiuto:
If you think you got the 'non surprising' behavior, go ahead and try to merge. Can you? If not, I will.

nacl:
> If you think you got the 'non surprising' behavior, go ahead and try to merge. Can you? If not, I will.

Yeah, I can.  Got sidetracked with a few internal things.

pkg/mappings.bzl
nacl: That's a good point, however I don't think we're quite there.  We'd have to do some introspection magic in `pkg_files` to identify the nature of the dependent rules.  Providers?  Rule names?  This will need to be investigated.  At first brush, this feels potentially error-prone.

I feel that this can be postponed to a later task.  I'll make a ticket for it.

PR #291, opened 2021-02-09
Add an example that shows how to get a user defined build setting into a package name

aiuto:
Uses user defined build setting to get a label
```
bazel build --//:label_from_command_line=bar  :tarball_that_needs_a_name
ls bazel-bin/foo-bar.tar
```


mzeren-vmw:
This PR might not be the best place to record this thought, but here it goes: We have rules internally in vmware that put "change numbers" in file names. One observation from experience is that pulling these files out of the build tree correctly is error prone on CI systems that do incremental builds. Changing file names will cause files to accumulate over time, filling up disk space. Also, if some piece of CI does not know the parameter passed into the build it may have to heuristically choose the "latest" file found with a glob. Which make me :( . We have been batting around the idea of "pkg_install" that would produce an "install script" that would copy (or link) from bazel-out file locations derived from targets to a directory outside of bazel-out, renaming as it goes. Ideally the pkg_install script would be the _only_ place that would need to know about command line parameters that change file names and we would not need this capability on all of the pkg_* rules. Thoughts? @nacl's opinions may differ.

aiuto:
LOL.  I just hacked this together to show a colleague how to get a command line variable into a rule. I didn't expect people to find it yet. 

I agree this PR is not the place to discuss the thought. It is a deep subject. Do you want to start a conversation in the mailing list or create a doc where different companies can bring their needs and solutions.

aiuto:
Let's continue your points in #301.  For now, I like this example because it does not invent anything. It just shows something people have asked for a lot but probably don't know how to do.

examples/naming_package_files/my_package_name.bzl
```diff
@@ -82,3 +82,19 @@ names_from_toolchains = rule(
...
+    return PackageVariablesInfo(values = values)
+
+#
+# Creating this build_setting defines a flag the user can set.
+#
+label_from_command_line = rule(
+    implementation = _label_from_command_line_naming_impl,
+    # Note that the default value comes from the rule instantiation.
+    build_setting = config.string(flag = True),
+)
```
nacl, line 100: What about using bazel_skylib's [`string_flag`](https://github.com/bazelbuild/bazel-skylib/blob/master/docs/common_settings_doc.md#string_flag) rule?

My understanding is that most people should be using it rather than rolling their own rule like this.

examples/naming_package_files/BUILD
```diff
@@ -104,3 +104,22 @@ debian_upload(
     # Refer to the target label, without having to know the final file name.
     package = ":a_deb_package",
 )
+
+#
+# Demonstrate how we can use a value from the command line in a package name.
+#
+# bazel build --//:label_from_command_line=bar  :tarball_that_needs_a_name
+# ls bazel-bin/foo-bar.tar
+label_from_command_line(
```
nacl, line 113: Nit: "label" could be misconstrued in the bazel context.  So could "tag", and "stamp", which are the other synonyms.  Bleh.

Not sure of any other alternatives off the top of my head, though.

nacl (approved):
Sorry for taking so long on this.  Comments below are minor.

I'd say the use of Skylib is the most important of the two.

examples/naming_package_files/my_package_name.bzl
aiuto: The string_flag provides a `BuildSettingInfo`, which is singled valued, so we would have to remap to a `PackageVariablesInfo`.    I think that would introduce a layer of complexity that might distract from the point.

examples/naming_package_files/BUILD
aiuto: I changed it to name_part.  That is sufficient different from label, tag, and stamp so it can't be confused with any of them. It is also sufficiently ugly that people won't cargo cult it.

examples/naming_package_files/my_package_name.bzl
nacl: Oooh.  Right.  Totally missed that.  My bad.

PR #293, opened 2021-02-16
Provide pkg_mklink, pkg_filegroup

nacl:
This change provides the last two pieces of the basic package creation puzzle: a
symbolic link provisioning rule (`pkg_mklink`) and a simple package grouping
rule (`pkg_filegroup`).

As compared to the original rules in `pkg/experimental/pkg_filegroup.bzl`:

- `pkg_mklink` (formerly `pkg_mklinks`) now operates on a single
  source-destination pair.  This is to ease integration of a planned feature
  where you will be able to make symbolic links to targets named by a label.
  The mechanics of this particular feature are not fleshed out at this time, but
  the current feeling is that these will be resolved as a part of the
  `pkg_filegroup` rule.

  To create multiple links, you can use a list comprehension in a BUILD
  file with minimal loss of clarity.

- `pkg_filegroup` is now an example of a "Grouping Rule", which allows for common
  groups of packaged files to be associated under a common name and prefix.  These
  are intended to be inputs to packaging rules (`pkg_rpm`, `pkg_tar`, etc.).

Tests are included for all of the above.

### DIFF
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -480,7 +480,7 @@ pkg_mklink = rule(
             mandatory = True,
         ),
         "attributes": attr.string(
-            doc = """Attributes to set on packaged directories.
+            doc = """Attributes to set on packaged symbolic links.
 
             Always use `pkg_attributes()` to set this rule attribute.
 
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -509,7 +509,7 @@ def _pkg_filegroup_impl(ctx):
                     },
                     attributes = s[PackageFilesInfo].attributes,
                 )
-                files.append(new_pfi)
+                files.append((new_pfi, s.label))
 
                 # dict.values() returns a list, not an iterator like in python3
                 mapped_files_depsets.append(s[DefaultInfo].files)
@@ -519,7 +519,7 @@ def _pkg_filegroup_impl(ctx):
                     dirs = [paths.join(ctx.attr.prefix, d) for d in s[PackageDirsInfo].dirs],
                     attributes = s[PackageDirsInfo].attributes,
                 )
-                dirs.append(new_pdi)
+                dirs.append((new_pdi, s.label))
 
             if PackageSymlinkInfo in s:
                 new_psi = PackageSymlinkInfo(
@@ -527,19 +527,19 @@ def _pkg_filegroup_impl(ctx):
                     destination = paths.join(ctx.attr.prefix, s[PackageSymlinkInfo].destination),
                     attributes = s[PackageSymlinkInfo].attributes,
                 )
-                links.append(new_psi)
+                links.append((new_psi, s.label))
     else:
         # Otherwise, everything is pretty much direct copies
         for s in ctx.attr.srcs:
             if PackageFilesInfo in s:
-                files.append(s[PackageFilesInfo])
+                files.append((s[PackageFilesInfo], s.label))
 
                 # dict.values() returns a list, not an iterator like in python3
                 mapped_files_depsets.append(s[DefaultInfo].files)
             if PackageDirsInfo in s:
-                dirs.append(s[PackageDirsInfo])
+                dirs.append((s[PackageDirsInfo], s.label))
             if PackageSymlinkInfo in s:
-                links.append(s[PackageSymlinkInfo])
+                links.append((s[PackageSymlinkInfo], s.label))
 
     return [
         PackageFilegroupInfo(
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -20,13 +20,15 @@ the following:
 
 - `pkg_files` describes destinations for rule outputs
 - `pkg_mkdirs` describes directory structures
+- `pkg_mklink` describes symbolic links
+- `pkg_filegroup` creates groupings of above to add to packages
 
 Rules that actually make use of the outputs of the above rules are not specified
 here.  TODO(nacl): implement one.
 """
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
-load("//:providers.bzl", "PackageDirsInfo", "PackageFilesInfo")
+load("//:providers.bzl", "PackageDirsInfo", "PackageFilegroupInfo", "PackageFilesInfo", "PackageSymlinkInfo")
 
 _PKGFILEGROUP_STRIP_ALL = "."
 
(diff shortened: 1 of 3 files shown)


nacl:
Latest change modifies the `PackageFilegroupInfo` provider to contain a list of `(child_provider, provider_origin)` tuples, which will aid diagnostics in packaging rules without being too disruptive to other parts of the ruleset.

This seemed to be a reasonable compromise between intrusiveness and clarity, but I'm willing to entertain other options, like embedding the origins directly in the mapping providers (e.g. `PackageFilesInfo`).

pkg/mappings.bzl
```diff
@@ -426,3 +428,166 @@ pkg_mkdirs = rule(
...
+            If the directory structure provided by this attribute is not
+            otherwise created when exist within the package when it is built, it
+            will be created implicitly, much like with `pkg_files`.
+
+            This path may be prefixed or rooted by grouping or packaging rules.
+
+            """,
+            mandatory = True,
+        ),
+        "src": attr.string(
```
aiuto, line 473: How are we going to point to a target with a dynamic name? 
foo_tool ->  foo_tool_1.2-x86-glibc2.1

pkg/mappings.bzl
```diff
@@ -426,3 +428,166 @@ pkg_mkdirs = rule(
...
+    links = []
+    mapped_files_depsets = []
+
+    if ctx.attr.prefix:
+        # If "prefix" is provided, we need to manipulate the incoming providers.
+        for s in ctx.attr.srcs:
+            if PackageFilesInfo in s:
+                new_pfi = PackageFilesInfo(
+                    dest_src_map = {
+                        paths.join(ctx.attr.prefix, dest): src
```
aiuto, line 514: Is paths OS specific or does it just use / as a sep?   I think we need to use / so that we can do the OS specific convert at package write time.

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -592,6 +608,269 @@ def _test_pkg_mkdirs():
...
+        ),
+    },
+)
+
+def _test_pkg_filegroup(name):
+    pkg_files(
+        name = "{}_pkg_files".format(name),
+        srcs = ["foo", "bar"],
+        prefix = "bin",
+        tags = ["manual"],
```
aiuto, line 786: Why the manual tag?

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -592,6 +608,269 @@ def _test_pkg_mkdirs():
...
+
+    # Then test that the results are equivalent to manually specifying the
+    # prefix everywhere.
+    pkg_filegroup_contents_test(
+        name = "{}_contents_prefix_translated".format(name),
+        target_under_test = "{}_prefixed_pfg".format(name),
+        expected_pkg_files = ["{}_pkg_files_prefixed".format(name)],
+        expected_pkg_dirs = ["{}_pkg_dirs_prefixed".format(name)],
+        expected_pkg_symlinks = ["{}_pkg_symlink_prefixed".format(name)],
+        # The origins for everythin will be wrong here, since they're derived
```
aiuto, line 856: everything

pkg/mappings.bzl
nacl: `src` can obviously be a template, however there should be an option to refer to a label.

Picking up names via templates?  That's trickier.  My immediate thought was that we could pass in a label that refers to a mapping provider (e.g. `PackageFilesInfo`), along with something used to key into it, if even needed:

- Source files must be unique within a single `pkg_files` provider, so we can key on a label
- Other symbolic links are simply a one-to-one mapping, so no keying is needed
- Directories are lists of untagged entities.  Not sure about this one.  If there's only one directory, it'll Just Work(TM), but that may not be a practical ask

Another option would be to create a `links` attribute associated with mapping rules (`pkg_files`), and then the individual rules could return a provider that represents symbolic links.  This is not supported by the provider we have currently have, though.  It would also make setting attributes difficult if not impossible.

I'd need to think on this a bit more to figure out how to tie all this together, but I think that we wouldn't have to change too much about what we already have.

pkg/mappings.bzl
nacl: `paths` always uses "/".  Quoth the [source](https://github.com/bazelbuild/bazel-skylib/blob/f80bc733d4b9f83d427ce3442be2e07427b2cc8d/lib/paths.bzl):

> NOTE: The functions in this module currently only support paths with Unix-style
path separators (forward slash, "/"); they do not handle Windows-style paths
with backslash separators or drive letters.

You mentioned in the call the morning regarding whether we needed `paths` from bazel-skylib.  I think we do -- at the very least we use `normalize`, `relativize`, and `join`.

We might be able to get away with vendoring it.  I'm not sure that's a good idea, but it can be considered later.

pkg/tests/mappings/mappings_test.bzl
nacl: It's to prevent the rules from actually being built when you run `bazel test //...`.

I'm not really sure about how analysis tests work, but if the rules aren't completely executed, then the labels used as sources can be bogus.  They'll still be presented to providers in the dependent rules, and the whole thing "just works" and I don't have to create a bunch of extra files.

It's a similar pattern to what's used when you write [negative analysis tests](https://docs.bazel.build/versions/master/skylark/testing.html#failure-testing).

pkg/tests/mappings/mappings_test.bzl
nacl: Yup, that's a typo.  Wasn't sure to what you were referring there earlier; thanks for the in-meeting clarification.

nacl:
@aiuto Ping.  Is there anything else you'd like to see done as a part of this PR?

pkg/mappings.bzl
aiuto: I actually didn't mean using a template. I was talking about refering to a label.  If we do the label, we can pick up the complex file name easily.

nacl: > I actually didn't mean using a template.

Ah, yes.  Regardless template via `PackageVariablesInfo` might be a good option for a lightweight case, or for pointing to dynamic locations outside of the package.

> I was talking about refering to a label. If we do the label, we can pick up the complex file name easily.

Fair enough.  However, what I was alluding to above was that our current provider interfaces may be enough to identify what the actual destination is without additional help:

1) If the provider (from rules_pkg or otherwise) points to multiple potential in-package destinations, the user will need a way to pick the precise one they want

2) If the non-packaging provider (`DefaultInfo`) points to a single source location (source/generated file), the destination may be ambiguous when we get to the grouping/packaging level.  The source location could correspond to multiple in-package destinations.

    The actual destination could be clarified by providing an additional label containing a `PackageFilesInfo` provider: sources within a single `PackageFilesInfo` must be unique.

    This doesn't solve the problem for `PackageDirsInfo`, which can contain multiple destinations with no source.

3) If a packaging provider contains a single destination (like `pkg_mklink`, or `pkg_files` and `pkg_mkdirs` with a single entry) there is no ambiguity.  The rule can calculate the destination based on what's in the provider, but then there could be additional ambiguity if multiple different providers are emitted by a single rule (the implementation currently allows this).

The easiest thing from the an implementation perspective is to enforce 3), but that might be inconvenient for users.

Regardless, I believe this can be added in after the fact.

nacl: Discussion can continue in #313.

aiuto (approved):
It feels like the origin should be in the info, rather than as a tuple, but I'm willing to face that choice after we try to use it for some more complex things. I'm not worried about backwards compatibility of these providers while we are pre 1.0, because only you and I should be using them.

pkg/tests/mappings/mappings_test.bzl
nacl: An interesting idea from @beasleyr-vmw that we've been mulling over internally is an extension to the tagging system where you could embed a reason into a tag (something like `manual:analysis-test`).

In particular, the `manual` tag has many use cases and it's usually not clear what, and comments are hard to query.  Something like this could make intent clearer w/o introducing additional terminology.

Probably worth a ticket against bazel itself if there isn't one already.

PR #303, opened 2021-02-25
Update experimental pkg_rpm rule to use new mapping/grouping providers

nacl:
This change replaces the experimental `pkg_rpm` rule in `pkg/experimental/rpm.bzl`
rule with one that is aware of the grouping rule structure.  This will soon become the
primary `pkg_rpm` rule made available to users.

Changes made to the experimental/ tree include:

- Usage of new providers and different paths everywhere

- Restructuring of how the `%files` metadata is built out.  It was restructured
  for clarity, and now makes use of the new `rpm_filetag` mapping attribute
  instead of `section`, which was since removed.

- Inputs to `pkg_rpm` are now specified with the `srcs` attribute instead of
  `data`, to be consistent between runtime requirements and compile-time sources.

- Conflicting content error detection and emission is now streamlined.

- Test coverage has improved, including new analysis tests for conflict
  detection, and bugs in the RPM python tests.

The old `pkg_filegroup` code in this same directory will be removed in a
separate change.

---

This particular implementation of `pkg_rpm` has some known defects and missing features.  These are tracked in #300, and will be fixed opportunistically.

PR #304, opened 2021-02-25
Provide pkg_filegroup-using pkg_rpm contents/analysis tests

nacl:
This change is built on top of #293 and #303, which are not yet submitted.  The actual work involved with this change is in 146bc2a.

---

This change provides a number of unit tests for the pkg_filegroup-using
`pkg_rpm` rule, meant to ensure that packages are formed as expected and
conflict detection is properly implemented.

The main differences between these and the tests in experimental/tests are:

- Analysis tests for conflict detection
- Fixed bugs WRT contents detection
- Path changes

This is likely not the final destination for these test cases, and this commit may be
combined with the rules implementation, depending on the size of the diff.  See
#303 for more details.

### DIFF
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -480,7 +480,7 @@ pkg_mklink = rule(
             mandatory = True,
         ),
         "attributes": attr.string(
-            doc = """Attributes to set on packaged directories.
+            doc = """Attributes to set on packaged symbolic links.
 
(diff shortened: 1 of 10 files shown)


nacl:
After discussion in a meeting on Monday 2021-03-01, this PR will be discarded in favor of overlaying directly on the `experimental/` tree.

#303 will be updated to contain these tests in a different location.

PR #305, opened 2021-02-27
Implemented package_variables substitution for package_dir attribute.

konste:


pkg/private/util.bzl
```diff
@@ -57,3 +52,28 @@ def setup_output_files(ctx, package_file_name=None):
...
+    Returns:
+      expanded_attribute_value: new value of the attribute after package_variables substitution
+    """
+    if not ctx.attr.package_variables:
+        # Nothing to substitute. Return the attribute value as is.
+        if type(attribute_value) == "string" and attribute_value.find("{") >= 0:
+            fail("package_variables is required when using '{' in attribute value %s" % attribute_value)
+        return attribute_value
+
+    if not attribute_value:
```
aiuto, line 72: I would move this and the next test to the top of the method.
That always catches the error of providing a non-string.
I think the current ordering would silently return a Bool as a value at line 70

pkg/tests/package_naming_aggregate_test.sh
```diff
@@ -21,4 +21,12 @@ declare -r DATA_DIR="${TEST_SRCDIR}/rules_pkg/tests"
 for pkg in test_naming_some_value.deb test_naming_some_value.tar test_naming_some_value.zip ; do
   ls -l "${DATA_DIR}/$pkg"
 done
+
+TAR_STRUCTURE=$(tar -tf tests/test_tar_package_dir_substitution.tar)
```
aiuto, line 25: can you do this in pkg_tar_test.py instead.  Shell tests are fragile.

aiuto (changes requested):
Thanks. Looks mostly great.

examples/naming_package_files/BUILD
```diff
@@ -68,7 +80,7 @@ names_from_toolchains(
 )
 
 pkg_tar(
-    name = "example2",
+    name = "example3",
```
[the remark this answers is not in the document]
konste: I have named my (new) example as "example2" and therefore renamed existing "example2" to "example3".

examples/naming_package_files/BUILD
```diff
@@ -60,6 +60,18 @@ pkg_tar(
     package_variables = ":my_naming_vars",
 )
 
+# This example shows that in the same way as package_file_name inner folder structure of the archive
```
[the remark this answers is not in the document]
konste: Fixed.

pkg/private/util.bzl
konste: Checks rearranged.

pkg/tests/package_naming_aggregate_test.sh
konste: Done.

pkg/.bazelrc
```diff
@@ -0,0 +1,3 @@
+common --enable_platform_specific_config=true
```
aiuto, line 1: I'm worried about all of this.  Is this needed for rules_pkg or is it your own?

pkg/.bazelrc
```diff
@@ -0,0 +1,3 @@
+common --enable_platform_specific_config=true
+build:windows --shell_executable="C:/Program Files/Git/usr/bin/bash.exe"
```
aiuto, line 2: This won't be portable for other people.  Do you need it now that you dropped the other shell test?

pkg/.bazelrc
konste: Indeed after shell test is replaced with Python test .bazelrc is not needed anymore. Removed.

aiuto (commented):
Looks good, but the tests are failing.

aiuto (approved):
Nice PR. 
- tests that work
- end user examples
- sensible refactoring

Thanks.

PR #314, opened 2021-03-05
Correct readme.md for rpm toolchain

GregBowyer:
The readme instructions for the rpm toolchain is incorrect, this should fix it.


nacl (approved):
Thanks for posting this!  Made some other related tweaks I noticed needed doing while in the neighborhood.

PR #318, opened 2021-03-15
experimental/rpm.bzl: use rpmbuild toolchain, re-enable RPM testing in CI

### LINKED ISSUE #183: Tests requiring `rpmbuild(8)` cannot be run in CI
The bazelci images for macOS and Ubuntu images in `bazelci` naturally do not have `rpmbuild(8)` on them, so any tests that run `rpmbuild` on them will fail.

I did a little digging into the docker images, and it looks like the provided centos7 image [doesn't have rpmbuild either](https://github.com/bazelbuild/continuous-integration/blob/e93e893071f90a066b2182c58ef4460ae6eb75eb/buildkite/docker/centos7/Dockerfile).

Fixing this will probably require modifying the CI setup, unless the centos7 image somehow has `rpmbuild(8)` in some way that is not apparent from examining the base docker image and the deplist of the installed packages in the image's dockerfile.

### LINKED ISSUE #299: experimental/rpm.bzl: `pkg_rpm` does not use the rpmbuild toolchain
In #254, a toolchain for `rpmbuild` and related tools was provided to the spec-file-based `pkg_rpm` rule.

This should be propagated to the `pkg_filegroup`-using one, wherever it is.

nacl:
This change modifies the pkg_rpm rule in `pkg/experimental/rpm.bzl` to use the
rpmbuild toolchain currently in use by the existing RPM packaging rule in
`pkg/rpm.bzl`.  Logic was largely copied from `pkg/rpm.bzl`.

Additionally, the RPM tests were removed from the disabled test set -- the
toolchain skipping system should be enough to take care of this on systems that
don't have rpmbuild (everything but CentOS in CI, right now).

Documentation was also updated to reflect bazel version compatibility.  Other
`pkg_filegroup`-related changes require 4.0.0.

Fixes #299, partially resolves #183.

### DIFF
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -10,8 +10,9 @@ common: &common
 tasks:
   centos7:
     platform: centos7
+    working_directory: /workdir/pkg
     test_targets:
-    # This include the rpm tests, whihc are not (yet) supported in the below
+    # This include the rpm tests, which are not (yet) supported in the below
     # cases.
     - "..."
   ubuntu1604:
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -12,7 +12,7 @@ tasks:
     platform: centos7
     working_directory: /workdir/pkg
     test_targets:
-    # This include the rpm tests, which are not (yet) supported in the below
+    # This includes the rpm tests, which are not (yet) supported in the below
     # cases.
     - "..."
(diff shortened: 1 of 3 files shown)


nacl:
Please disregard the CLA message -- I had originally committed using the wrong e-mail.

aiuto (approved):

Issue #269, opened 2021-01-15
Packaging `strip_prefix` does not function on or within TreeArtifacts (directories)


nacl:
`strip_prefix` for rules within rules_pkg removes directory components from paths.  This is very useful for users who wish to maintain parts of directory structures found in their source trees.

A potential misconception, though, is to what it applies.  It only applies to directories that can be known at analysis time.  TreeArtifacts and directories within do not seem to be considered.

Examples to demonstrate this are forthcoming.

This can be resolved by having the packaging rules do this at runtime, or via documentation that explains how to do prefix stripping for such artifacts.

PR #319, opened 2021-03-18
Improve pkg_filegroup-using pkg_rpm TreeArtifact installation

### LINKED ISSUE #292: experimental/rpm.bzl does not properly install TreeArtifacts
The experimental pkg_rpm rule (upon which the new RPM packaging rules are based) uses `cp -r` to install "file" artifacts.  This includes TreeArtifacts.

`cp -r $A $B` will recursively copy `$A` to `$B`.  If `$B` doesn't exist, `$B` the output structure will look like `$A/$B`.  If it does, then it will look like `$A/$B/$B`, which is not desirable.

A known workaround is `renames = {":directory_rule": "."}`.

The best fix for this is not yet clear.

This will be fixed while we migrate the RPM packaging rule to the new providers framework.

nacl:
`pkg_rpm` with `pkg_filegroup` never properly consumed TreeArtifacts (directory
outputs).  This is due to two factors:

- In RPM packages, adding a directory as a file results in the directory being
  owned and every file under it having consistent permissions.  In our
  experience, this has led to confusing behavior and can make RPM
  package content listings harder to understand.

- The command for copying files into place does the wrong thing when the
  target directory already exists for any reason.

This change remedies the above by enumerating TreeArtifact directory structures
during the execution phase (as they are not knowable at analysis time), and
passing the outputs to a command that creates additions to the install scripts
and `%files` manifests.  Files in TreeArtifacts are now installed and owned by
the RPM individually.  Attributes are applied identically to all files in the
TreeArtifact, and TreeArtifact processing is only performed if any are detected.

It is currently not possible to any additional processing to TreeArtifacts,
including excluding files and stripping prefixes (#269).  Until such
time, workarounds such as using `genrule`s or other custom rules should be used.

Regression tests are also provided.

Additional changes include:

- Fixed some uninclusive language.

- The "RPM contents manifest" code for the `pkg_rpm` tests has been factored
  into a library.

- `buildifier` opportunistically applied.

- Removing an unused test symlink

NOTE 1: Another possible alternative implementations involve changing the existing
`PackageFilesInfo` provider and/or creating a new one that represents
TreeArtifacts.  This was not explored, but many of the implementation details
used here can be re-purposed should that be considered.

NOTE 2: The change does not attempt to do conflict checks between TreeArtifact
installations and other TreeArtifact installations, or
files/directories/symlinks already identified at analysis time.  To reduce the
complexity of this otherwise large change, this feature currently deferred.  My
current (untested) understanding is that if this happens, duplicates in `%files`
will be detected, and cause most versions of `rpmbuild` to fail.

Fixes #292.

### DIFF
--- a/pkg/experimental/augment_rpm_files_install.py
+++ b/pkg/experimental/augment_rpm_files_install.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python3
+
+# Copyright 2021 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script takes a pair of Bazel-generated RPM %install scriptlet and %files
+# list, and combined with JSON description of TreeArtifacts, emits copies
+# augmented with the files detected in the materialized TreeArtifacts.
+
+import os
+import sys
+import json
+
+# XXX: Keep this in sync with the same variable in rpm.bzl
+_INSTALL_FILE_STANZA_FMT = """
+install -d %{{buildroot}}/$(dirname {1})
+cp {0} %{{buildroot}}/{1}
+""".strip()
+
+# Cheapo arg parsing.  Currently this script is single-purpose.
+
+# JSON file containing the TreeArtifact manifest info.
+#
+# This is expected to be an array of objects with the following fields:
+#
+# - src: Source file/directory location.
+# - dest: Install prefix
+# - tags: Tags for the %files manifest
+dir_data_path = sys.argv[1]
+
+# Existing files
+existing_install_script_path = sys.argv[2]
+existing_files_path = sys.argv[3]
+
+# Output files
+new_install_script_path = sys.argv[4]
+new_files_path = sys.argv[5]
+
+# Computed outputs to be combined with the originals
+dir_install_script_segments = []
+dir_files_segments = []
+
+with open(dir_data_path, 'r') as fh:
+    dir_data = json.load(fh)
+
+    for d in dir_data:
+        # d is a dict, d["src"] is the TreeArtifact directory to walk.
+        print(d["src"])
+        for root, dirs, files in os.walk(d["src"]):
+            # "root" is the current directory we're walking through.  This
+            # computes the path the source location (the TreeArtifact root) --
+            # the desired install location relative to the user-provided install
+            # destination.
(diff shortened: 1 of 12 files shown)


aiuto:
I'm not going to be able to look at this until Monday. I am right in the
middle of a crunch.

nacl:
> I'm not going to be able to look at this until Monday. I am right in the middle of a crunch.

No problem, thanks!

nacl:
@aiuto Could you look at this when you get a chance?

pkg/experimental/augment_rpm_files_install.py
```diff
@@ -0,0 +1,94 @@
...
+
+# This script takes a pair of Bazel-generated RPM %install scriptlet and %files
+# list, and combined with JSON description of TreeArtifacts, emits copies
+# augmented with the files detected in the materialized TreeArtifacts.
+
+import os
+import sys
+import json
+
+# XXX: Keep this in sync with the same variable in rpm.bzl
```
aiuto, line 25: XXX?

pkg/providers.bzl
```diff
@@ -34,12 +34,12 @@ PackageFilesInfo = provider(
     fields = {
         "attributes": """Attribute information, represented as a `dict`.
 
-        Keys are strings representing attribute identifiers, values are
-        arbitrary data structures that represent the associated data.  These are
-        most often strings, but are not explicity defined.
+Keys are strings representing attribute identifiers, values are
```
aiuto, line 37: Is the dedent on purpose?

aiuto (approved):
Just two formatting nits. 
To be honest, I can't tell if the tree artifact handling is correct, because I am not quite sure how it tree artifacts work myself. But since this is experimental, we can shake it out over time.

nacl:
> To be honest, I can't tell if the tree artifact handling is correct, because I am not quite sure how it tree artifacts work myself. But since this is experimental, we can shake it out over time.

Fair enough.  

However, I'm soon going to be working on migrating the RPM builder out of `experimental/`, so this will not be true for long.  Should this feature be gated on the addition of an attribute in `pkg_rpm` instead?

aiuto:
> > To be honest, I can't tell if the tree artifact handling is correct, because I am not quite sure how it tree artifacts work myself. But since this is experimental, we can shake it out over time.
> 
> Fair enough.
> 
> However, I'm soon going to be working on migrating the RPM builder out of `experimental/`, so this will not be true for long. Should this feature be gated on the addition of an attribute in `pkg_rpm` instead?

I don't think you should wait. Having it in will make it a lot easier to test incrementally.

nacl:
> > > To be honest, I can't tell if the tree artifact handling is correct, because I am not quite sure how it tree artifacts work myself. But since this is experimental, we can shake it out over time.
> > 
> > 
> > Fair enough.
> > However, I'm soon going to be working on migrating the RPM builder out of `experimental/`, so this will not be true for long. Should this feature be gated on the addition of an attribute in `pkg_rpm` instead?
> 
> I don't think you should wait. Having it in will make it a lot easier to test incrementally.

Aye.  Will be submitting this shortly.

pkg/experimental/augment_rpm_files_install.py
nacl: Changed to `NOTE`

PR #320, opened 2021-03-20
Support custom compression program

### LINKED ISSUE #59: pkg_tar gzip would be faster with a concurrent (pigz) mode
pkg_tar uses Python's built-in tarfile module, and will only use a single CPU core for compression. If it worked in parallel (like pigz), it would be much faster.

As part of the Kubernetes build, we make a [300MB .tar.gz file](kubernetes-release/release/v1.6.0/kubernetes-server-linux-amd64.tar.gz) from 1.3GB of inputs. This step takes around 100 seconds.

```
$ pv kubernetes-server-linux-amd64.tar | gzip | wc -c
1.23GB 0:00:38 [32.9MB/s] [===>] 100%            
312612343
$ pv kubernetes-server-linux-amd64.tar | pigz | wc -c
1.23GB 0:00:05 [ 248MB/s] [===>] 100%            
311930961
```

HackAttack:
Resolves #59

### DIFF
--- a/pkg/archive.py
+++ b/pkg/archive.py
@@ -137,22 +137,6 @@ class TarFileWriter(object):
           2000-01-01, which is compatible with non *nix OSes'.
       preserve_tar_mtimes: if true, keep file mtimes from input tar file.
     """
-    if compression in ['bzip2', 'bz2']:
-      mode = 'w:bz2'
-    else:
-      mode = 'w:'
-    self.gz = compression in ['tgz', 'gz']
-    # Fallback to xz compression through xz.
-    self.use_xz_tool = False
-    if compression in ['xz', 'lzma']:
-      if HAS_LZMA:
-        mode = 'w:xz'
-      else:
-        self.use_xz_tool = True
-    self.name = name
-    self.compressor = compressor
-    self.root_directory = root_directory.rstrip('/').rstrip('\\')
-    self.root_directory = self.root_directory.replace('\\', '/')
     self.preserve_mtime = preserve_tar_mtimes
     if default_mtime is None:
       self.default_mtime = 0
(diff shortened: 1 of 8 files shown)


pkg/archive.py
```diff
@@ -422,7 +425,14 @@ def close(self):
     # Close the gzip file object if necessary.
     if self.fileobj:
       self.fileobj.close()
-    if self.use_xz_tool:
+    if self.compressor:
```
aiuto, line 428: This approach wastes time and space because it fully writes the fie uncompress, then calls the compressor.
Instead, this should be done with a passed in fileobj the way we do for gz format.

You could wrap the compressor invocation as a subprocess.run(), where you forward writes to the fileobj to stdin of the process.

aiuto (changes requested):

pkg/archive.py
HackAttack: Isn’t this the exact same approach used by the `xz` code below?

aiuto:
Yes. But the xz code path is also non optimal. It is also not used for most
people, since their python will have lzma support and we will use the
inline compressor.

pkg/archive.py
HackAttack: This is resolved, and the `xz` code migrated to use the same (improved) code path.

pkg/archive.py
```diff
@@ -162,11 +146,35 @@ def __init__(self,
...
 
     self.fileobj = None
-    if self.gz:
-      # The Tarfile class doesn't allow us to specify gzip's mtime attribute.
-      # Instead, we manually re-implement gzopen from tarfile.py and set mtime.
-      self.fileobj = gzip.GzipFile(
-          filename=name, mode='w', compresslevel=9, mtime=self.default_mtime)
+    self.compressor_cmd = compressor.split() if compressor else None
+    if self.compressor_cmd:
+      pass
```
aiuto, line 151: This would be easier to follow if you put lines 169-173 here instead of the pass. 
if compressor_cmd
  ... do that
elif xz|lzma:
   ..
elif ...

pkg/docs/reference.md
```diff
@@ -145,6 +146,20 @@ Creates a tar file from a list of inputs.
         </p>
       </td>
     </tr>
+    <tr>
+      <td><code>compressor</code></td>
+      <td>
+        <code>Label, optional</code>
+        <p>Executable to be used as compression program.</p>
```
aiuto, line 153: maybe   program => filter, so it is obvious that the program does not specify input or output files.

pkg/pkg.bzl
```diff
@@ -76,6 +76,8 @@ def _pkg_tar_impl(ctx):
         "--owner=" + ctx.attr.owner,
         "--owner_name=" + ctx.attr.ownername,
     ]
+    if ctx.executable.compressor:
+        args.append("--compressor=%s %s" % (ctx.executable.compressor.path, ctx.attr.compressor_args))
```
aiuto, line 80: This kind of construction worries me.    How does arg parsing know this arg takes several values.
Just make it two distinct args.  compressor, compressor_args, and the second one is quoted as
--compresser_args='%s'

aiuto (changes requested):
Much nicer. Thanks.

The thing it needs now is some sort of test, or at least an executable example in the /examples tree. Otherwise, this will get inevitably get broken.

pkg/archive.py
HackAttack: Those lines need to run after the if/else chain because the `xz` branch might set `compressor_cmd`. I’ll add a comment to the `pass` unless we find another refactoring (`if not self.compressor_cmd` is more expressive but also not great since it requires an additional level of indentation…maybe a helper function).

pkg/pkg.bzl
HackAttack: These args are interpreted directly by `argparse`, not the shell, so I think quotes would be interpreted as part of the argument. My guess is it does a naive `split('=')[1]`

HackAttack: I confirmed that this works correctly and the value of the argument is a space-delimited string.

aiuto: I'm not convinced, because I know of existing bugs in the way we handle passing args to actions on various platforms. But I'll let it slide now, because I'm going to rewrite this method to use the ctx.action.args rather than a string list. So we'll see what happens then.

aiuto (changes requested):
One last thing, can you add this to .bazelci/test.yaml in the windows section

- "//tests:test_tar_compression"

pkg/pkg.bzl
HackAttack: I added an assertion to validate the arg passing.

HackAttack:
Done. Should this be necessary though, since it is a data dependency of `archive_test`, which is in the list?

aiuto:
Oh.... dang. The name fooled me.   "test_tar_compression" is not a test, though it sounds like one.
The whole 'archive_test' is a mess. I should make it several distinct ones, but that is not your problem.
Thanks for adding this.

aiuto (approved):

PR #322, opened 2021-03-24
Support SOURCE_DATE_EPOCH in experimental/rpm.bzl; enable file clamping in make_rpm.py

nacl:
NOTE: this is a review built on top of the changes in #319.  This will be rebased on #319 when necessary.  The part up for review in this PR is 95af33c.

---

This change allows experimental/rpm.bzl to support setting [`SOURCE_DATE_EPOCH`],
sourced either directly inline in BUILD files or from a rule, as in the existing `pkg_rpm` rule.

The Starlark code in experimental/rpm.bzl is largely identical to the existing code in rpm.bzl.

Additionally, it was discovered that `SOURCE_DATE_EPOCH` only affects the RPM's
build time (not the files) unless you explicitly tell it otherwise.  This seems
to be the `%clamp_mtime_to_source_date_epoch` macro, first available in rpm 4.13+.
It is now set automatically in RPMs when `SOURCE_DATE_EPOCH` is set in both
`pkg_rpm` rules.

Tests are also provided for `experimental/rpm.bzl` only.  Tests for the
existing/legacy `pkg_rpm` rule can be added if so desired.

Additionally:

- Command debugging in `make_rpm.py` was expanded to include printing out the
  environment.

- The test helper routines were made even more generic -- now users can specify
  the exact CSV manifest formats they want.  This is used to check for mtimes for
  this particular use case which are otherwise not checked.

- Documentation fixes where noted.  Some of the TODOs will be tagged with
  tickets when opportune (soon).

[`SOURCE_DATE_EPOCH`]: https://reproducible-builds.org/docs/source-date-epoch/


nacl:
The CI failures are caused by the version of rpmbuild in CentOS 7 being too old to support `SOURCE_DATE_PREFIX`.  I will take a look into this tomorrow.

pkg/rpm.bzl
```diff
@@ -233,7 +233,10 @@ Args:
   changelog: A changelog file to include. This will not be written to the spec
     file, which should only list changes to the packaging, not the software itself.
   source_date_epoch: Value to export as SOURCE_DATE_EPOCH to facilitate reproducible
+    timestamps.  Implicitly sets the `%clamp_mtime_to_source_date_epoch` in the
+    subordinate call to `rpmbuild` to facilitate more consistent in-RPM file
     timestamps.
-  source_date_epoch_file: File containing the SOURCE_DATE_EPOCH value
+  source_date_epoch_file: File containing the SOURCE_DATE_EPOCH value.  Sets
```
aiuto, line 239: Is a file the way most people use SOURCE_DATE_EPOCH?
or should this, instead, to the glue to pass that from the caller environment?
or is this fine for now and you can do the other later?

pkg/providers.bzl
```diff
@@ -34,12 +34,12 @@ PackageFilesInfo = provider(
     fields = {
         "attributes": """Attribute information, represented as a `dict`.
 
-        Keys are strings representing attribute identifiers, values are
-        arbitrary data structures that represent the associated data.  These are
-        most often strings, but are not explicity defined.
+Keys are strings representing attribute identifiers, values are
```
aiuto, line 37: Did you dedent this on purpose?

pkg/experimental/tests/rpm/tree_artifacts/rpm_contents_vs_manifest_test.py
```diff
@@ -0,0 +1,59 @@
...
+/test_dir/f,d41d8cd98f00b204e9800998ecf8427e,root,root,100644,,
+""".strip()
+
+
+class PkgRpmCompManifest(unittest.TestCase):
+    def setUp(self):
+        self.runfiles = runfiles.Create()
+        self.maxDiff = None
+        # Allow for parameterization of this test based on the desired RPM to test.
+        self.rpm_path = self.runfiles.Rlocation(os.path.join(
```
aiuto, line 40: I think you want '/'.join().   What I found when trying to make tests portable to windows is that Rlocation insists on Bazel paths, not os specific ones.  I  had to replace all my carefully thought out os.path uses with /.

pkg/experimental/tests/rpm/tree_artifacts/create_directory_with_contents.py
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+import os
+
+"""
+Creates a directory containing some files with determined contents
```
aiuto, line 21: The style we have been using is """First line of pydoc...\n\n...

pkg/experimental/tests/rpm/source_date_epoch/rpm_contents_vs_manifest_test.py
```diff
@@ -0,0 +1,103 @@
...
+
+def version_to_string(version):
+    return ".".join(str(i) for i in version)
+
+
+class PkgRpmCompManifest(unittest.TestCase):
+    # Support for SOURCE_DATE_EPOCH is only available as of rpm
+    # 4.14: http://rpm.org/wiki/Releases/4.14.0
+    #
+    # TODO(nacl): it would probably be better to ask the rpmbuild(8) binary for
```
aiuto, line 41: I wonder if.....
The rpmbuild toolchain can add the version as a property (or just a field saying "supports_source_date_epoch")
We can have an rpm_test rule which is just a py/sh_test that depends on the toolchain
The test rule can pass the version or support indicator to the test runner. 

On the one hand, that feels more precise than running RPM build to find out if the test can run.
But... it seems like so much more scaffolding than needed.

Having the toolchain provide the data and the test depend on it. Great.
Passing the data into the test runner. Ugly.

Sort of need incompatible target skipping for tests, which maybe exists.

pkg/experimental/tests/rpm/source_date_epoch/BUILD
```diff
@@ -0,0 +1,133 @@
...
+# Test handling the source_date_epoch attribute
+############################################################################
+
+py_test(
+    name = "source_date_epoch_from_file",
+    srcs = ["rpm_contents_vs_manifest_test.py"],
+    data = [":rpm_sde_fromfile_data"],
+    env = {"TEST_RPM": "rpm_sde_fromfile.rpm"},
+    main = "rpm_contents_vs_manifest_test.py",
+    tags = [
```
aiuto, line 101: As I was hinting at in the other comment, maybe an exec_compatible_with might be the thing to use here.
But... also happy to come back to it. That may force a minimum Bazel version and thus be a breaking change.

aiuto (commented):
There is a lot going on here.
My comments are pretty much all nits.
I can approve even if you don't want to change anything.

pkg/experimental/tests/rpm/source_date_epoch/rpm_contents_vs_manifest_test.py
nacl: > I wonder if.....
> The rpmbuild toolchain can add the version as a property (or just a field saying "supports_source_date_epoch")
> We can have an rpm_test rule which is just a py/sh_test that depends on the toolchain
> The test rule can pass the version or support indicator to the test runner.

Hm.  I tried writing a wrapper around python tests similar when I was working on testing TreeArtifact manipulations.  Couldn't quickly determine how to create something that looks like a custom `py_test` from a rule, so I gave up and wrote a macro instead.

The macro approach could be taken here too.  Doesn't seem like it would be too hard to get something saying "supports_source_date_epoch" into a `config_setting`.

> On the one hand, that feels more precise than running RPM build to find out if the test can run.
> But... it seems like so much more scaffolding than needed.

Yup, that was my thought.  There's way too much going on here.

What is present here seems "good enough" for now until we can easily arbitrary versions of rpmbuild with bazel.

> Having the toolchain provide the data and the test depend on it. Great.
> Passing the data into the test runner. Ugly.

Indeed

> Sort of need incompatible target skipping for tests, which maybe exists.

This is what I want to see happen, but it just looks to be highly complex for little gain.

I have with rules_pkg's testing system for things like debian packages -- if the `dpkg` tools are missing from the system, then the tests are skipped within

pkg/experimental/tests/rpm/source_date_epoch/BUILD
nacl: As stated above, I think we should come back to this.  There are some other places where this would be useful in rules_pkg.

And WRT compatibility, we're already requring 4.0 as of #290, which uses the `json` module.  #319 does too.  The README states this as well.

pkg/experimental/tests/rpm/tree_artifacts/create_directory_with_contents.py
nacl: Fixed.

pkg/experimental/tests/rpm/tree_artifacts/rpm_contents_vs_manifest_test.py
nacl: Right, I remember this.  Fixed.

We might want to consider using something like `PosixPath` from [pathlib] to assemble bazely paths in the future.

pathlib: [https://docs.python.org/3/library/pathlib.html#pathlib.PurePosixPath]

pkg/providers.bzl
nacl: Yes.  I spent an annoying amount of time misreading this a few weeks back, so I made it more obvious.

I have no problem reverting this, but the structure of the provider `fields` can look like a large string blob if you aren't reading it closely.  Once upon a time, a colleague of my tried something like this:

```python
    dict(
        attributes = """Attribute information, represented as a `dict`.

        Keys are strings representing attribute identifiers, values are
        arbitrary data structures that represent the associated data.  These are
        most often strings, but are not explicity defined.

        For known attributes and data type expectations, see the Common
        Attributes documentation in the `rules_pkg` reference.
        """,
        ...
    )
```

Github's default syntax highlighting doesn't make this easy to distinguish, but with color styles where identifiers are vastly different this is much easier to read.

WDYT?  Nothing we need to worry about in the here and now, though.

pkg/rpm.bzl
nacl: I legit don't know.  We don't use it anywhere that I know of.

This is primarily to establish parity with changes added during #113.

The most correct way to do this off the top of my head would to pass this in via [`--workspace_status_command`](https://docs.bazel.build/versions/master/user-manual.html#flag--workspace_status_command).

pkg/experimental/tests/rpm/source_date_epoch/rpm_contents_vs_manifest_test.py
aiuto: SGTM

pkg/rpm.bzl
aiuto: OK. I was just asking. I don't know either.

aiuto (approved):

Issue #323, opened 2021-03-26
Support cross-building packages


nacl:
Package managers like `rpm` have considerable support for differentiating between host and target platforms.  Most of the time, packages built targeting a platform are built on that platform, e.g. RPM and Debian packages are more often than not built on Linux for Linux.

There are some use cases where developers may want to cross-build packages.  This will vary per package manager, but the core of it will require packaging rules to hook into Bazel's platform constraint system.

An example of how to query the constraint system from a rule can be found [here](https://github.com/bazelbuild/proposals/blob/master/designs/2019-11-11-target-platform-constraints.md#proposal-a-check-method-that-takes-a-constraint-value-label).

PR #324, opened 2021-03-26
Ensure that pkg_rpm tests can run on macOS if rpm(build) is present

### LINKED ISSUE #255: //experimental/tests/rpm:test_rpm_manifest is linux only
The bash script expects the tool md5sum to be present on the machine. On mac it is md5.  We could make a portable version in python.

nacl:
The experimental `pkg_rpm` tests make use of CSV manifests.  An important part
of this is verifying file consistency.  For this, among other things, we capture
MD5 checksums of expected package contents, and compare them to the outputs from
a formatted RPM query.

The tool initially used for this is `md5sum` from GNU coreutils.  macOS (and
Windows) do not have this utility by default; in particular, macOS has a `md5`
tool that does essentially the same thing.

To allow RPM tests to run on platforms without GNU coreutils, this change adds
in a simple cross-platform MD5 hashing tool written with Python, and replaces
all instances of `md5sum` with it.

Additionally, when running tests locally, I encountered an issue with macOS
symbolic link permissions.  In macOS (and presumably BSDs), symlinks can have a
mode set directly on them, whereas in Linux, symlink permissions are not
applicable and can't even be set.  To fix this, on non-Linux OSes, I assume that
we have the BSD `chmod` tool with the `-h` flag that can set symbolic link
permissions directly.  This is inserted into the `%install` script as a macro;
Linux users of `rpmbuild` will see no practical differences.

The above is likely not true when `rpmbuild` is run in MinGW and Cygwin, which
almost certainly use GNU coreutils, but we'll cross that bridge when we get to
it.  See also #323.

This does not attempt to do anything with allowing RPM tests to run on CI.
Tests were run locally on a system running macOS 11.2 (Big Sur).

Fixes #255.


pkg/experimental/tests/rpm/md5.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
```
aiuto, line 1: This seems useful everywhere. Can you move it to pkg/tests instead of experimental?

pkg/experimental/tests/rpm/BUILD
```diff
@@ -183,7 +183,7 @@ genrule(
     for f in $(location :config_file); do
         (
             echo -n /$$(basename $$f),
```
aiuto, line 185: This might be simpler to understand as.
echo -n /$$(basename $$f),$($(execpath :md5) $$f),root,root,100644,cnm'

Or maybe not. But I think $($(execpath :md5) $$f)  might do what you want without having to fool around with the tr \\n ,

aiuto (approved):

pkg/experimental/tests/rpm/BUILD
nacl: > This might be simpler to understand as.
> echo -n /$$(basename $$f),$($(execpath :md5) $$f),root,root,100644,cnm'
> 
> Or maybe not.

This ended up working for me:

```
echo /$$(basename $$f),$$($(execpath //tests/util:md5) $$f),root,root,100644,cmn, >> $@
```

It's not bad, but it all kinda blurs together when I read it.  The multi-line one works has a similar problem, but it's easier to stick comments in there to explain empty fields or other oddities.

In an ideal world, there would be a rule that would accept `pkg_filegroup`s and emit CSV without having to write a script like this.  It's likely overkill w/o uses outside of `pkg_rpm`, though.

> But I think $($(execpath :md5) $$f) might do what you want without having to fool around with the tr \n ,

Indeed it does.  Thanks for the suggestion.

I'm going to leave this as-is for now.  I'll take a look at adjusting this in a future change.

aiuto (approved):

pkg/experimental/tests/rpm/md5.py
nacl: Indeed.  Done.

PR #330, opened 2021-04-16
Codify how to remove TreeArtifact base directories in `pkg_files` 

### LINKED ISSUE #269: Packaging `strip_prefix` does not function on or within TreeArtifacts (directories)

nacl:
We often have TreeArtifacts (directory artifacts) that are built to be installed
directly onto the filesystem in packages, but their names are often not correct.
They, thus must be renamed.  There are two ways to do this after we submit #319.

1) Rename the TreeArtifact to the destination
2) Rename the TreeArtifact to nothing

I believe that (2) is clearer, as the `prefix` provided to `pkg_files` stands
out more than an entry in `renames`.  This change codifies a workaround we have
tried out internally -- renaming to "nothing".  "Nothing" in this case, is represented by the
constant `REMOVE_BASE_DIRECTORY`, (1) is still allowed.

This change does not attempt to do anything related to updating paths inside
TreeArtifacts.  This will be attempted in a follow-on change.

Tests and documentation updates are also provided.

This change also moves the common analysis test boilerplate to its own location
in `pkg/tests/util/defs.bzl`, where it will be more generally accessible in
future changes.

Partially resolves #269.


pkg/mappings.bzl
```diff
@@ -24,14 +24,16 @@ the following:
...
-here.  TODO(nacl): implement one.
+here.
 """
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//:providers.bzl", "PackageDirsInfo", "PackageFilegroupInfo", "PackageFilesInfo", "PackageSymlinkInfo")
 
 _PKGFILEGROUP_STRIP_ALL = "."
 
+REMOVE_BASE_DIRECTORY = ""
```
aiuto, line 35: Does this all still work if we made this an impossible value, like  chr(0)

aiuto (approved):
I'm sorry I missed this one.

pkg/mappings.bzl
nacl: Not without modifications elsewhere.  Let me see how much it would take.

nacl:
> I'm sorry I missed this one.

No worries, thanks for taking a look!

pkg/mappings.bzl
nacl: It turned out to be pretty easy.  PTAL.

aiuto (approved):

PR #335, opened 2021-04-29
Ensure tree_artifact test RPMs have consistent permissions

nacl:
The recently created tree_artifact experimental pkg_rpm tests assumed that the
default `user:group` for files in the RPM is `root:root`.  When run on macOS (I
have a local copy of `rpm`), it ended up being `root:wheel`.

Fix this by setting `attributes` in all `pkg_files`s in
`pkg/experimental/tests/rpm/tree_artifacts/BUILD`.


aiuto (approved):

nacl:
Thanks!

Issue #307, opened 2021-03-02
pkg_rpm supports pkg_filegroup framework (graduate from experimental)


nacl:
PR #303 updates the `pkg_rpm` rule to use the providers in`pkg/providers.bzl` via rules in `pkg/mappings.bzl`.  This signals adequate code readiness, but a problem remains in how to ensure compatibility for users moving on to 1.0.

Tasks include:

- [x] Completing #303 
- [x] Creating a compatibility macro that replaces the current `pkg_rpm` rule.  It will choose which `pkg_rpm` rule to call based on the presence of `spec_file` attribute
- [x] Move the current `pkg_rpm` rule (and its tests) to a new "legacy" package
- [ ] Move the experimental `pkg_rpm` rule (and its tests) to the root package

PR #337, opened 2021-04-29
Deprecate pkg/rpm.bzl and move it to pkg/legacy/rpm.bzl

nacl:
The `pkg_filegroup` framework has evolved considerably since its creation, and
the `rpm` packaging rule that uses it (in `pkg/experimental/rpm.bzl`) has as
well.  We believe it is time to move it out of `experimental` and into general
use.

This change prepares for this by:

- Moving the existing `pkg/rpm.bzl` module and its associated tests to new
  package in `pkg/legacy`.

- Replacing the contents `pkg/rpm.bzl` with a compatibility macro that allows
  for easily switching between the two (the interfaces are significantly
  different).  The macro switches between the two based on mandatory attributes
  that are available in one instance of `pkg_rpm` but not the other.

CI, packaging, and test targets were also updated as needed.

Part of #307.


aiuto (approved):
Works for me!

nacl:
Awesomeeeee

PR #338, opened 2021-04-29
Graduate experimental pkg_rpm to mainline 

### LINKED ISSUE #307: pkg_rpm supports pkg_filegroup framework (graduate from experimental)

nacl:
The `pkg_filegroup` framework has evolved considerably since its creation, and
the `rpm` packaging rule that uses it (in `pkg/experimental/rpm.bzl`) has as
well.  We believe it is time to move it out of `experimental` and into general
use.

This change moves `pkg/experimental/rpm.bzl` to `pkg/rpm_pfg.bzl` for direct
access, and the compatibility macro in `pkg/rpm.bzl` is also modified to use it.
All tests have been updated and moved to subdirectories of `pkg/tests/rpm`.
`pkg/legacy/rpm.bzl` was also updated to indicate its deprecated status.  Distro
packaging and CI configuration were updated as well.

`@rules_pkg` has been opportunistically removed from some `load`s in test BUILD
files, and some pathing was also fixed for compatibility with windows.  There may
few other places where both of these would need to be cleaned up, and were not
investigated.

Resolves #307.


aiuto (approved):
I guess this needs the rebase first.

nacl:
> I guess this needs the rebase first.

Done.  Will submit after testing is complete.

nacl:
Thanks for taking a look at this!

PR #342, opened 2021-05-04
Remove the capability to independently name the Debian .changes file

### LINKED ISSUE #289: Decide if deb_pkg needs the capability to have a .changes file named differently than the package.
Right now, you can name the output package with packge_file_name. This can pull in variables and expand in interesting ways. A correspondingly named .changes file will be created as the default.
There is also an attribute to let you explicitly name the .changes file.

Is that attribute of any value or does it add complexity for no reason? That is, should we remove it before 1.0?

aiuto:
Remove the capability to have the Debian .changes file have a different naming pattern than the .deb file

Closes #289

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -190,17 +190,7 @@ def _pkg_deb_impl(ctx):
         package_file_name = package_file_name,
     )
 
-    # If the user does not provide the changes file, compute it.
-    # TODO(https://github.com/bazelbuild/rules_pkg/issues/289): Should the
-    # user even be allowed to provide it? That is, is there any value in
-    # ability to have the basename of the .deb and the .changes file be
-    # different?
-    if ctx.outputs.changes:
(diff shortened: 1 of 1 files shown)


aiuto:
Friendly ping

dannysullivan (commented):
I don't see anything in particular that concerns me, but this is a very large diff, including lots of new files (were they all moved from `experimental`?). I'd be interested in seeing a more detailed summary of where these changes came from/what we expect them to do. For example, this PR contains a bunch of changes related to time stamping - I'm not sure how that connects to the name of the PR.

aiuto:
WTF..... the merge history is botched. This was like a tiny change. I will fix it.

aiuto:
Got it. Now it is down to 1 file again.

dannysullivan (approved):
Awesome, that makes way more sense - LGTM.

aiuto:
Thanks!

PR #345, opened 2021-05-04
Fix compression detection from tar file extension.

aiuto:
Fix compression detection from tar file extension.
    - and add some tests to see that we really do compress the right way.
    - fixes bug where user could specify `extnsion=".tar.gz"` and get the wrong thing


pkg/pkg.bzl
```diff
@@ -132,12 +132,16 @@ def _pkg_tar_impl(ctx):
...
-        elif ctx.attr.extension == "tgz":
-            args += ["--compression=gz"]
+        compression = None
+        dotPos = ctx.attr.extension.rfind(".")
+        if dotPos >= 0:
+            compression = ctx.attr.extension[dotPos+1:]
+        else:
+            compression = ctx.attr.extension
+        if compression == "tgz":
+            compression = "gz"
```
nacl, line 142: If you're looking for completeness, you could also do `tbz2` and `txz` for `bzip2` and `xz`-compressed-archives.

Consider also adding `.tbz2` and `txz` to `tar_filetype` at the top of the file if you do this.  Might be worth a ticket if you don't here.

pkg/pkg.bzl
```diff
@@ -132,12 +132,16 @@ def _pkg_tar_impl(ctx):
...
+        if compression and compression in SUPPORTED_TAR_COMPRESSIONS:
+            args += ["--compression=%s" % compression]
```
nacl, line 144: Shouldn't this `fail` if an unsupported compression scheme is used?

pkg/tests/BUILD
```diff
@@ -536,3 +536,42 @@ filegroup(
     ],
     visibility = ["//visibility:public"],
 )
+
```
nacl, line 539: This file needs some visible division to make it easier to browse.  Perhaps a subject of a future change.

pkg/tests/is_compressed_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+  def setUp(self):
+    super(IsCompressedTest, self).setUp()
+    self.data_files = runfiles.Create()
+
+
+  def is_zip_compressed(self, file_name):
+    """Returns true if file_name is zip compressed."""
+    with open(self.get_file_under_test(file_name), 'rb') as inp:
+      content = inp.read()
```
nacl, line 38: Consider only reading the first two bytes here (`inp.read(2)`).  Ditto with the first 7 of the `bzip2` one.

pkg/tests/is_compressed_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+  def get_file_under_test(self, file_name):
+    """Get the file path to a generated zip in the runfiles."""
+
+    return self.data_files.Rlocation(
+        "rules_pkg/tests/" + file_name
+    )
+
+  def setUp(self):
+    super(IsCompressedTest, self).setUp()
```
nacl, line 31: Is this necessary?

pkg/tests/is_compressed_test.py
```diff
@@ -0,0 +1,62 @@
...
+from bazel_tools.tools.python.runfiles import runfiles
+
+
+class IsCompressedTest(unittest.TestCase):
+
+  def get_file_under_test(self, file_name):
+    """Get the file path to a generated zip in the runfiles."""
+
+    return self.data_files.Rlocation(
+        "rules_pkg/tests/" + file_name
```
nacl, line 27: Consider `os.environ['WORKSPACE_NAME']` instead of `rules_pkg` directly for portability.

pkg/tests/is_compressed_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+import unittest
+
+from bazel_tools.tools.python.runfiles import runfiles
+
+
+class IsCompressedTest(unittest.TestCase):
+
+  def get_file_under_test(self, file_name):
+    """Get the file path to a generated zip in the runfiles."""
```
nacl, line 24: Nit: `s/zip/archive`

pkg/tests/BUILD
```diff
@@ -31,8 +31,8 @@ filegroup(
         ":test_tar_package_dir_file",
     ],
     visibility = [
-        "//tests:__subpackages__",
         "//legacy/tests:__subpackages__",
+        "//tests:__subpackages__",
```
nacl, line 35: Did you do this with `buildifier`, or by hand?  Might be worth calling out `buildifier` in your commit message if done with `buildifier`.

nacl (approved):
LGTM other than minor comments.

Could do a little additional detection on filenames if you want, but that can be deferred.

pkg/pkg.bzl
aiuto: I don't understand the landscape of file naming to know if I want completeness.
My personal belief has always been that you have one extension, not two.  That is .tgz is correct and .tar.gz is an abomination. Then I met people with other beliefs. I'ld like to defer adding more until we get requests.

pkg/pkg.bzl
aiuto: Whoops. Yes. Done

pkg/tests/is_compressed_test.py
aiuto: Good idea.

pkg/tests/is_compressed_test.py
aiuto: Not really. Removed

PR #348, opened 2021-05-05
Make `find_system_rpmbuild` repo rule depend on `PATH`

### LINKED ISSUE #344: Autodetecting `rpmbuild` toolchain is not updated after first use
I have a custom copy of `rpmbuild` which is not always in my PATH, and I often forget to add it in before calling `bazel`.  The autodetecting toolchain correctly notices its absence:

```
DEBUG: /private/var/tmp/_bazel_$USER/297e6fd0ec15ec7901f2a2bf25d17005/external/rules_pkg/toolchains/rpmbuild_configure.bzl:35:16: No system rpmbuild found.
```

However, even if I update `PATH` to contain the `rpmbuild` binary, bazel will produce an identical log message.  I have to `clean` the tree to get it to acquire the correct toolchain.

Theoretically, this could be fixed by adding `PATH` to the [`environ`](https://docs.bazel.build/versions/4.0.0/skylark/lib/globals.html#repository_rule) attribute of the `_find_system_rpmbuild` repository rule in pkg/toolchains/rpmbuild_configure.bzl.

nacl:
This change modifies `find_system_rpmbuild`'s underlying repository rule to be
aware of modifications to `PATH`.  Without this change, `find_system_rpmbuild`
will only run once, and cannot be run again without `clean`ing the tree.

Fixes #344.

### DIFF
--- a/pkg/toolchains/rpmbuild_configure.bzl
+++ b/pkg/toolchains/rpmbuild_configure.bzl
@@ -39,6 +39,7 @@ _find_system_rpmbuild = repository_rule(
     implementation = _find_system_rpmbuild_impl,
     doc = """Create a repository that defines an rpmbuild toolchain based on the system rpmbuild.""",
     local = True,
+    environ = ["PATH"],
     attrs = {
         "verbose": attr.bool(
             doc = "If true, print status messages.",


aiuto (approved):
Cool

PR #353, opened 2021-05-11
Move build_tar, build_zip and helpers to private

aiuto:
This PR changes no behavior. It moves build_tar and build_zip to //private to emphasize that people should not depend on the interface. There will be extensive changes to the tools to support `pkg_files` so I am making this declaration of privacy now.

In followup PRs we should move make_deb and make_rpm.


pkg/private/BUILD
```diff
@@ -53,3 +53,56 @@ py_library(
...
+py_library(
+    name = "archive",
+    srcs = [
+        "__init__.py",
+        "archive.py",
+    ],
+    srcs_version = "PY3",
+    visibility = [
+        "//:__pkg__",
+        "//experimental:__pkg__",
```
nacl, line 66: `//experimental` is gone now

pkg/make_deb.py
```diff
@@ -31,7 +31,7 @@
 
 # list of debian fields : (name, mandatory, wrap[, default])
 # see http://www.debian.org/doc/debian-policy/ch-controlfields.html
-from helpers import GetFlagValue
+from rules_pkg.private.helpers import GetFlagValue
```
nacl, line 34: Is the `rules_pkg` part derived from the workspace name?

This would make it more difficult for users to change the workspace name on import.  Perhaps `imports = [".."]` could be added to the corresponding `py_library`s?  Modules could then be imported relative to `private`. This may end up being a largely moot point when all the `make_*` scripts are moved to `pkg/private`, though.

Speaking of changing the workspace name, I found that in order to do this, we need to rename all uses of the `rpmbuild` toolchain.  Not sure if there's anything we want to do or reasonably can do about this.

nacl (approved):
Cool.  Mostly nits and a question.

pkg/private/BUILD
aiuto: Removed

pkg/make_deb.py
aiuto: I removed the rules_pkg part.

aiuto:
I don't think we can do much about the toolchain naming problem. It feels
like a basic Bazel bug to me.

PR #358, opened 2021-05-28
Code clean up. Move all pkg_tar compression logic to the same code block

aiuto:
No functional changes. It just makes it easier to understand how we set the compression up.

cc: @HackAttack


dannysullivan (approved):

aiuto:
thanks

PR #359, opened 2021-06-02
Change compute_data_path to use ctx as an input instead of output_file.

aiuto:
This is a refactoring before the PR where we introduce pkg_files* support for pkg_tar & pkg_zip. The intent for that PR is that there will be a helper method like

```
add_pkg_srcs(ctx, manifest, ctx.attr.srcs)
```

That will call comput_data_path and then use that per file in ctx.attr.srcs to compute the destination path. Using ctx instead of an arbitrary output file feels cleaner.

This was more broadly tested in the work in this branch, https://github.com/aiuto/rules_pkg/tree/tar_pkg_files.
I have most of the final PR there, but pulling this out makes an easier review.

Bonus side effect: Since I need ctx to call the function, I had to turn part of path_test.py into an analysistest instead of the very hacky python test that reads starlark as python.


pkg/path.bzl
```diff
@@ -61,8 +61,9 @@ def dest_path(f, strip_prefix, data_path_without_prefix = ""):
         return f_short_path[len(strip_prefix):]
     return f_short_path
 
-def compute_data_path(out, data_path):
+def compute_data_path(ctx, data_path):
```
nacl, line 64: Consider providing a docstring here.  If you do, it would be worth mentioning the requirements of `ctx`.

pkg/tests/path_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+        "expected_path": attr.string(mandatory = True),
+    },
+)
+
+def _test_compute_data_path():
+    pkg_mkdirs(
+        name = "dummy",
+        dirs = [],
+        tags = ["manual"],
+    )
```
nacl, line 47: Is this strictly necessary?  If it is, it might be worth a comment saying it's unused and something needs to be passed into `target_under_test`.

pkg/tests/path_test.bzl
```diff
@@ -0,0 +1,86 @@
...
+
+"""Tests for path.bzl"""
+
+load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts", "unittest")
+load("//:mappings.bzl", "pkg_mkdirs")
+load("//:path.bzl", "compute_data_path")
+
+##########
+# Test compute_data_path
+##########
```
nacl, line 23: I get why you're using analysis tests here, but it feels like much of this could instead be a unit test instead.

Might be worth a comment, but it's not essential.

nacl (approved):
LGTM other than nits (mostly docs)

pkg/path.bzl
aiuto: Done

pkg/tests/path_test.bzl
aiuto: Hmmm.. I wonder if I can do that.  I will come back to that later this week if I have time.
For now, I want to get this refactoring in so that I can mail you the PR that adds pkg_files support to pkg_tar.
I pulled this out of that.

PR #361, opened 2021-06-08
Stop importing single methods from modules. Use the entire thing.

aiuto:
That is
```
from private.helpers import Foo
...
x = Foo()
```
becomes
```
from private import helpers
...
x = helpers.Foo()
```

This is for consistency of usage and it makes it easier to write [copybara](https://github.com/google/copybara) scripts to import the code.


nacl (approved):
Naturally, I'm not familiar with copybara.  LGTM regardless.

aiuto:
https://github.com/google/copybara

PR #362, opened 2021-06-09
Update BUILD for PY versions and licenses=notice

aiuto:


aiuto:
Doing some tidying before a release.

PR #363, opened 2021-06-09
Fix some things that Google's strident buildifier warns about.

aiuto:


pkg/legacy/rpm.bzl
```diff
@@ -200,6 +200,7 @@ pkg_rpm = rule(
     toolchains = ["@rules_pkg//toolchains:rpmbuild_toolchain_type"],
 )
 
+# buildifier: disable=no-effect
```
nacl, line 203: This could perhaps instead be fixed by including it in the `doc` attribute of `pkg_rpm`.   I always wondered why this string was here -- was it from a time when there was no `doc` argument in `rule`?

nacl (approved):
One small thing to consider, but LGTM otherwise.

pkg/legacy/rpm.bzl
aiuto: Yes. I wanted to do that. But I was not sure quite where to put it. I want to defer moving it until we do the global documentation overhaul. For now, I am just trying to smooth import into Google, so I can  start testing our migration to this package.

PR #364, opened 2021-06-09
Enable more tests on windows

aiuto:
Turned on most of them.
Fixed an Rlocation bug in one test.


nacl (approved):
LGTM.

In a future change, it might be a good idea to use something like `exec_compatible_with` on the incompatible targets to exclude them from running on Windows instead of doing so here.

aiuto:
exec_compatible_with should work.  Since we now have committed to requiring bazel 4.x for 1.0 we can probably start using it. I'll keep that in mind.

Issue #287, opened 2021-02-03
Packaging rules should support stamping


nacl:
_Originally posted by @aiuto in https://github.com/bazelbuild/rules_pkg/issues/285#issuecomment-770153075_

Let's discuss solutions in an issue first.  I would like to to have as many people as possible weigh in on what is useful.
There are a lot of questions
- do we change the stamp of every file?
- does that include files which came from pre-made archives that have a particular timestamp of their own?
- some people want to put the timestamp into the output filename. While I find that dubious, any stamping solution should consider it.
- Why is stamp_mtime an attribute?  Why not just support the --stamp flag. That seems to be the API of least surprise.
- and.... no solutions that just to .tar.   We need do do the appropriate thing for each of pkg_zip, pkg_deb and pkg_rpm.  But that makes getting the API right before doing any code even more important.

nacl:
AFAIK, my particular use case does not have a preference WRT preserving mtimes, but some things that immediately come to mind are:

- Some rules support [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) (e.g. `pkg_rpm`), others do not.  It also applies to more than mtimes, so that may be out of scope here.

- `--stamp` feels like the tool of choice here, but `BUILD_TIMESTAMP` may also not be the right value to use.  Bazel provides it for free.

aiuto:
The least surprising behavior is that all the top level pkg rules implement `stamp` as done in cc_binary.  I consider that a minimal requirement.

**stamp**
- Integer; optional; default is -1
- Enable link stamping. Whether to encode build information into the binary. Possible values:
  - stamp = 1: Stamp the build information into the binary. Stamped binaries are only rebuilt when their dependencies change. Use this if there are tests that depend on the build information.
  - stamp = 0: Always replace build information by constant values. This gives good build result caching.
  - stamp = -1: Embedding of build information is controlled by the --[no]stamp flag.

#288 demonstrates how the default behavior could be done.  That is, I can get the build time for stamping if --stamp is set on the command line. I didn't do the 0 and 1 cases, nor implement for every rule. 

An orthogonal feature to that is if we change the time stamps of files expanded out of archives (if I am reading #265 right). That is it's own debate. The request I have seen in the past is along the lines of
- I have a bundle of static files that will be served by a web service
- they have a set of creation dates.
- I bundle them into a zip file and check into my source tree
- I want to include them in a .deb package, but individually not as one zip
[...]

PR #365, opened 2021-06-10
Add --stamp support to pkg_zip.

aiuto:
Advances #287 so that pkg_tar and pkg_zip now support --stamp in the same way that various embedded *_binary rules do.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -509,6 +509,11 @@ def _pkg_zip_impl(ctx):
     args.add("-d", ctx.attr.package_dir)
     args.add("-t", ctx.attr.timestamp)
     args.add("-m", ctx.attr.mode)
+    inputs = []
+    if ctx.attr.stamp == 1 or (ctx.attr.stamp == -1 and
+                               ctx.attr.private_stamp_detect):
+        args.add("--stamp_from", ctx.version_file.path)
+        inputs.append(ctx.version_file)
 
     data_path = compute_data_path(ctx, ctx.attr.strip_prefix)
     for f in ctx.files.srcs:
@@ -520,8 +525,8 @@ def _pkg_zip_impl(ctx):
 
     ctx.actions.run(
         mnemonic = "PackageZip",
-        inputs = ctx.files.srcs,
-        executable = ctx.executable.build_zip,
+        inputs = ctx.files.srcs + inputs,
+        executable = ctx.executable._build_zip,
         arguments = [args],
         outputs = [output_file],
         env = {
@@ -559,9 +564,14 @@ pkg_zip_impl = rule(
             doc = "See Common Attributes",
             providers = [PackageVariablesInfo],
         ),
+        "stamp": attr.int(default = 0),
+
+        # Is --stamp set on the command line?
+        # TODO(https://github.com/bazelbuild/rules_pkg/issues/340): Remove this.
+        "private_stamp_detect": attr.bool(default = False),
 
         # Implicit dependencies.
-        "build_zip": attr.label(
+        "_build_zip": attr.label(
             default = Label("//private:build_zip"),
             cfg = "exec",
             executable = True,
@@ -592,5 +602,9 @@ def pkg_zip(name, **kwargs):
     pkg_zip_impl(
         name = name,
         out = archive_name + "." + extension,
+        private_stamp_detect = select({
+            _stamp_condition: True,
+            "//conditions:default": False,
+        }),
         **kwargs
     )
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -369,7 +369,13 @@ pkg_tar_impl = rule(
             doc = "See Common Attributes",
             providers = [PackageVariablesInfo],
         ),
-        "stamp": attr.int(default = 0),
+        "stamp": attr.int(
+            doc = """Enable file time stamping.  Possible values:<ul>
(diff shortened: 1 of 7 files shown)


pkg/pkg.bzl
```diff
@@ -559,9 +564,14 @@ pkg_zip_impl = rule(
             doc = "See Common Attributes",
             providers = [PackageVariablesInfo],
         ),
+        "stamp": attr.int(default = 0),
```
nacl, line 567: Document this in the reference?

pkg/tests/stamp_test.py
```diff
@@ -13,17 +13,52 @@
 # limitations under the License.
 """Test time stamping in pkg_tar"""
```
nacl, line 14: And now `pkg_zip`.  Perhaps make this docstring generic?

pkg/tests/stamp_test.py
```diff
@@ -13,17 +13,52 @@
...
 from bazel_tools.tools.python.runfiles import runfiles
 
 # keep in sync with archive.py
 PORTABLE_MTIME = 946684800  # 2000-01-01 00:00:00.000 UTC
 
-class PkgTarTest(unittest.TestCase):
-  """Testing for pkg_tar rule."""
+
+class StampTest(unittest.TestCase):
+  """Test for time stamps in packages."""
```
nacl, line 29: This implies this might also test the normal case where artifacts have a consistent time.  This doesn't seem to happen here.

It might be worth clarifying this point, and perhaps also stating somewhere the test for the default epoch is elsewhere.

pkg/tests/stamp_test.py
```diff
@@ -13,17 +13,52 @@
...
+  ALLOWED_DELTA_FROM_NOW = 10000  # seconds
+
+  def check_mtime(self, mtime, file_path, file_name):
+    """Checks that a time stamp is reasonable.
+
+    This checks that a timestamp is not 0 or any of the well known EPOCH values,
+    and that it is within some delta from the current time.
+
+    Args:
+      mtime: timestamp in seconts
```
nacl, line 43: Nit: "seconts" -> "seconds"

pkg/tests/zip_test.py
```diff
@@ -12,18 +12,27 @@
...
 LOREM_CRC = 2178844372
 EXECUTABLE_CRC = 342626072
 
 
+# The ZIP epoch date: (1980, 1, 1, 0, 0, 0)
+_ZIP_EPOCH_DT = datetime.datetime(1980, 1, 1, 0, 0, 0, tzinfo=datetime.timezone.utc)
+_ZIP_EPOCH_S = int(_ZIP_EPOCH_DT.timestamp())
+
+def seconds_to_ziptime(s):
+  dt = datetime.datetime.utcfromtimestamp(s)
```
nacl, line 32: Python's upstream recommendation is to use something like:

```python
datetime.datetime.fromtimestamp(s, datetime.timezone.utc)
```

Mostly b/c leap seconds.  This is probably fine for this use case, though.

[Documentation](https://docs.python.org/3.5/library/datetime.html#datetime.datetime.utcfromtimestamp).

pkg/tests/stamp_test.py
```diff
@@ -13,17 +13,52 @@
...
+           file_path, file_name))
+    if mtime == StampTest.ZIP_EPOCH:
+       self.fail('Archive %s contains file %s with ZIP epoch' % (
+           file_path, file_name))
+    if ((mtime < self.target_mtime - StampTest.ALLOWED_DELTA_FROM_NOW)
+        or (mtime > self.target_mtime + StampTest.ALLOWED_DELTA_FROM_NOW)):
+       self.fail(
+           'Archive %s contains file %s with mtime:%d, expected:%d +/- %d' % (
+               file_path, file_name, mtime, self.target_mtime,
+               StampTest.ALLOWED_DELTA_FROM_NOW))
```
nacl, line 61: Consider using asserts instead of this conditional logic -- `self.assertNotEqual` and `self.assertAlmostEqual` come to mind.

[Documentation](https://docs.python.org/3/library/unittest.html).

nacl (approved):
LGTM other than small documentation nits and minor suggestions.

Sorry for the delays in looking at this.

pkg/tests/stamp_test.py
aiuto: Done

pkg/tests/stamp_test.py
aiuto: Done

pkg/tests/stamp_test.py
aiuto: Right.  I changed the comment to indicate non-epochness.

pkg/tests/zip_test.py
aiuto: I'm inclined to keep this as is because it is the same code used in the build_zip tool.

pkg/tests/stamp_test.py
aiuto: AlmostEqual is probably too precise.  If our tests get big enough, we could build the files minutes before using them in a test, so the timestamp could as far off of "now()" as the total length of test time. I chose this as a reasonable hedge against variability vs time to build super robust tests if we don't need the extra robustness.

aiuto: I added these comments into the code.

aiuto:
Thanks for the review.

PR #366, opened 2021-06-10
Add a rule to make artifacts that mimic binaries

aiuto:
`fake_artifact` is a new rule in the test infrastructure.  It creates an output with the ability to set it as executable, and to specify runfiles. It is designed only to be able to mimic complex objects like a cc_binary, without having to depend on the cc_toolchain.

In this specific PR, it is used to replace a sh_binary that was not really a runnable binary. It just looked enough like a binary so we could build a tests with it as data. Using a `fake_artifact` makes it more obvious that the target has no behavior itself.

### DIFF
--- a/pkg/tests/mappings/mappings_test.bzl
+++ b/pkg/tests/mappings/mappings_test.bzl
@@ -53,10 +53,6 @@ def _pkg_files_contents_test_impl(ctx):
     expected_dests = {e: None for e in ctx.attr.expected_dests}
     n_found = 0
     for got in target_under_test[PackageFilesInfo].dest_src_map.keys():
-        # skip .exe files because they are unexpected extra outputs of
-        # sh_binary on Windows.
-        if got.endswith(".exe"):
-            continue
         asserts.true(
             got in expected_dests,
             "got <%s> not in expected set: %s" % (got, ctx.attr.expected_dests))
@@ -133,7 +129,7 @@ def _test_pkg_files_contents():
     # Used in the following tests
     fake_artifact(
         name = "testdata/test_script",
-        runfiles = ["testdata/a_script.sh"],
+        files = ["testdata/a_script.sh"],
         tags = ["manual"],
     )
 
@@ -448,7 +444,7 @@ def _test_pkg_files_rename():
     # Used in the following tests
     fake_artifact(
         name = "test_script_rename",
-        runfiles = ["testdata/a_script.sh"],
+        files = ["testdata/a_script.sh"],
         tags = ["manual"],
     )
 
--- a/pkg/tests/mappings/mappings_test.bzl
+++ b/pkg/tests/mappings/mappings_test.bzl
@@ -27,7 +27,12 @@ load(
     "strip_prefix",
     "REMOVE_BASE_DIRECTORY",
 )
-load("//tests/util:defs.bzl", "directory", "generic_base_case_test", "generic_negative_test")
+load("//tests/util:defs.bzl",
+     "directory",
+     "fake_artifact",
+     "generic_base_case_test",
+     "generic_negative_test"
+)
 
 ##########
 # Helpers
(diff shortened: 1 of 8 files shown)


pkg/tests/util/defs.bzl
```diff
@@ -58,6 +58,44 @@ Paths containing directories will also have the intermediate directories created
...
+def _fake_artifact_impl(ctx):
+    out_file = ctx.actions.declare_file(ctx.attr.name)
+    content = ['echo ' + rf.path for rf in ctx.files.runfiles]
+    ctx.actions.write(
+        output = out_file,
+        content = '\r\n'.join(content),
+        is_executable = ctx.attr.executable,
+    )
+    return DefaultInfo(
+        files = depset([out_file]+ctx.files.runfiles),
```
nacl, line 70: Nit: should be whitespace around the `+`

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -126,12 +131,9 @@ def _test_pkg_files_contents():
     )
 
     # Used in the following tests
-    #
-    # Note that since the pkg_files rule is never actually used in anything
-    # other than this test, nonexistent_script can be included with no ill effects. :P
-    native.sh_binary(
+    fake_artifact(
```
nacl, line 134: Might be worth explaining what `fake_artifact` replaces in the build in your commit message.

pkg/tests/util/defs.bzl
```diff
@@ -58,6 +58,44 @@ Paths containing directories will also have the intermediate directories created
...
+depend on a large toolchain.
+    """,
+    implementation = _fake_artifact_impl,
+    attrs = {
+        "deps": attr.label_list(
+            doc = "Dependencies to trigger other rules, but are then discarded.",
+            allow_files = True,
+        ),
+        "runfiles": attr.label_list(
+            doc = "Deps which are passed in DefaultInfo as runfiles.",
```
nacl, line 89: It doesn't look like the implementation matches the documentation.  When I think of `runfiles`, I would expect to see `ctx.runfiles` be used in the rule and it be passed to the `runfiles` argument in `DefaultInfo`.  As it stands, I would just associate these as files that need to be passed to consumers.

Perhaps to make this distinction, there should be both `files` and `runfiles`, with the `files` being the what's currently in the rule?

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -27,7 +27,12 @@ load(
     "strip_prefix",
     "REMOVE_BASE_DIRECTORY",
 )
-load("//tests/util:defs.bzl", "directory", "generic_base_case_test", "generic_negative_test")
+load("//tests/util:defs.bzl",
+     "directory",
+     "fake_artifact",
+     "generic_base_case_test",
+     "generic_negative_test"
+)
```
nacl, line 36: (Github won't let me add a comment in the relevant location)

On lines 51-52/56-57, there's a comment that mentions `.exe` files and `sh_binary`.  This comment should be updated to remove the mention of `sh_binary` if it's still needed, or the logic immediately below should be removed if possible.

nacl (changes requested):
Cool.  Some small things I'd like clarified/fixed, though.

pkg/tests/mappings/mappings_test.bzl
aiuto: Done

pkg/tests/util/defs.bzl
aiuto: Good catch. I split files and runfiles to distinct attributes.

pkg/tests/mappings/mappings_test.bzl
aiuto: Done.

pkg/tests/util/defs.bzl
nacl: LGTM

nacl (approved):
Thanks for addressing the comments!

PR #368, opened 2021-06-22
Update load location for private/util.bzl

jin:
It's now located at https://github.com/bazelbuild/rules_pkg/blob/main/pkg/private/util.bzl, not //private:util.bzl


aiuto:
I'm going to guess that you tried to use the source rather than the release and saw the load fail.

Would it help if we added documentation about how to vendor it into your own third_party and set up a local repository?

jin:
That's right :) 

I figured out how to use strip_prefix to also include the `pkg` dir. It would be great to have instructions on using the source zips (or warnings not to)

PR #371, opened 2021-07-06
Send symlinks from pkg_tar to build_tar via the manifest instead of the command line

aiuto:
Send symlinks from pkg_tar to build_tar via the manifest instead of the command line.

- This required changing some test which were file order dependent. The
  manifest files now get added earlier
- Remove the unneeded flags from build_tar
- Ran buildifier (which changed too much), sorry.


nacl (approved):
Looks good at a glance, but it looks like there's still gaps in the testing for `pkg_filegroup` framework integration, specifically ensuring that a symlink requested by `pkg_mklink` or a specially constructed manifest does the right thing.

Thanks for running `buildifier`.

aiuto:
I'm waiting to get the manifest golden test in place.
That will make it easier to see that complex combinations of things produce the right manifest.
Then, for a single manifest of all the bizarre things, we can reuse that against build_*.py to ensure that they can write the complex bits.

To put it another way. I am trying to make # of tests O(# pkg_file features) + O(# archive types) and not O(features * types).

PR #372, opened 2021-07-06
In pkg_tar, use manifest for empty_file instead of command line.

aiuto:
- create new helper to add empty files to the manifest
- remove obsolete --empty_file flag from build_tar
- also remove unused --empty_dir flag


pkg/private/pkg_files.bzl
```diff
@@ -202,6 +203,28 @@ def add_directory(content_map, dir_path, origin, mode=None, user=None, group=Non
         group = group,
     )
 
+def add_empty_file(content_map, dest_path, origin, mode=None, user=None, group=None):
+    """Add a single file to the content map.
```
nacl, line 207: Perhaps this docstring should say "Add a single empty file to the content map"?

nacl (approved):
LGTM, given the situation with `pkg_tar`.  Comment is a nit.

A question though: how long should this `EMPTY_FILE` manifest entry type live?  It  seems to be something one could just implement using `pkg_files`es.  Should there be a `pkg_*` rule and an associated provider for doing this as well?

pkg/private/pkg_files.bzl
aiuto: Done.

aiuto:
> A question though: how long should this `EMPTY_FILE` manifest entry type live? It seems to be something one could just implement using `pkg_files`es. Should there be a `pkg_*` rule and an associated provider for doing this as well?

Great question. pkg_files could add it, or pkg_empty_file() could be added.  I included it in pkg_tar for backwards compatibility, but I don't see it as a priority feature though. I don't ever recall seeing a need for an installer to create an empty file sitting somewhere. Usually, it at least has some text in it to say why it exists.

I'm willing to defer any thought on it until someone comes up with a well reasoned use case.

nacl:
> > A question though: how long should this `EMPTY_FILE` manifest entry type live? It seems to be something one could just implement using `pkg_files`es. Should there be a `pkg_*` rule and an associated provider for doing this as well?
> 
> Great question. pkg_files could add it, or pkg_empty_file() could be added. I included it in pkg_tar for backwards compatibility, but I don't see it as a priority feature though. I don't ever recall seeing a need for an installer to create an empty file sitting somewhere. Usually, it at least has some text in it to say why it exists.
> 
> I'm willing to defer any thought on it until someone comes up with a well reasoned use case.

This all sounds great to me.  Thank you for the clarification.

Issue #309, opened 2021-03-02
pkg_zip supports pkg_filegroup framework


nacl:

aiuto:
Still need support for symlinks

aiuto:
Still need tree_artifact support.

aiuto:
Closed by #537

PR #373, opened 2021-07-07
Initial support for pkg_files* in pkg_zip

aiuto:
- This is the minimal PR to switch to the newer style. That is, only plain files.
- Followup CLs will add things previously not supported, such as links.
- Advances #309


pkg/pkg.bzl
```diff
@@ -579,10 +579,43 @@ def _pkg_zip_impl(ctx):
...
+                    # Tree artifacts need a name, but the name is never really
+                    # the important part. The likely behavior people want is
+                    # just the content, so we strip the directory name.
+                    dest = '/'.join(d_path.split('/')[0:-1])
+                    add_tree_artifact(content_map, dest, f, src.label)
+                else:
+                    # Note: This extra remap is the bottleneck preventing this
+                    # large block from being a utility method as shown below.
+                    # Should we disallow mixing pkg_files in srcs with remap?
+                    # I am fine with that if it makes the code more readable.
```
nacl, line 611: I'd vote for disallowing this -- having both being usable simultaneously feels potentially confusing IMO.

Might be worth doing in a follow-on change, though, as this could make the tests a bit more complicated.

pkg/private/build_zip.py
```diff
@@ -70,24 +125,13 @@ def main(args):
...
-
-      dst_path = _combine_paths(args.directory, dst_path)
-
-      entry_info = zipfile.ZipInfo(filename=dst_path, date_time=ts)
-
-      if default_mode:
-        entry_info.external_attr = default_mode << 16
-
-      entry_info.compress_type = zipfile.ZIP_DEFLATED
+  with zipfile.ZipFile(args.output, mode='w') as zip_file:
```
nacl, line 128: I've been writing all of my Python with 4-space indents, and much of the other python code is using 2-space indents.  Should we standardize on one or the other?

pkg/private/build_zip.py
```diff
@@ -60,6 +72,49 @@ def parse_date(ts):
...
+  if entry_type == ENTRY_IS_FILE:
+    entry_info.compress_type = zipfile.ZIP_DEFLATED
+    with open(src, 'rb') as src:
+      zip_file.writestr(entry_info, src.read())
+  elif entry_type == ENTRY_IS_DIR:
+    entry_info.compress_type = zipfile.ZIP_STORED
+    # Set directory bits
+    entry_info.external_attr |= (UNIX_DIR_BIT << 16) | MSDOS_DIR_BIT
+    zip_file.writestr(entry_info, '')
+  # TODO(aiuto): All the rest
```
nacl, line 117: Mention a ticket here?

nacl (approved):
You weren't kidding about this size of this.  A great start indeed.  Comments are meta or minor.

It feels like there should be some more testing here, but this LGTM otherwise.  Feel free to submit now if you want to disallow using `pkg_*` along with normal files in `srcs` in a later change.

aiuto:
Now waiting for #384

nacl:
> Now waiting for #384

You should be good to go.

pkg/private/build_zip.py
aiuto: Years of Google 2-space indent vs. other people's 4 space have left me ill-suited to choose. This code all started out as 2-space, and didn't change when we released it.
I'm fine with picking 4 space, but first let's pick a linter that warns of the indent and
then do the change in one big pass that has no functional changes.
Actually, maybe two passes. One to re-indent but not change line lengths, so we can compare before and after with diff -R -w and see that they are identical. And a second pass to allow the linter to rewrap the, now longer, lines.

pkg/private/build_zip.py
aiuto: Done.  #309

pkg/pkg.bzl
aiuto: Whoops. This comment is vestigial from a copy/paste. There is no remap_paths in pkg_zip, so the comment is moot. I removed it and the commented out line that referenced (the non-existent) remap_paths.

PR #378, opened 2021-07-12
minor doc fixes

aiuto:
- Fixes for links that moved.
- add rule to generate pkg_files docs

### DIFF
--- a/pkg/README.md
+++ b/pkg/README.md
@@ -4,7 +4,7 @@
 -   [Overview](#overview)
 -   [Roadmap](#roadmap)
 -   [Reference](docs/reference.md)
--   [Examples](examples/readme.md)
+-   [Examples](/examples/readme.md)
 
 <a name="notes"></a>
 ## Release Notes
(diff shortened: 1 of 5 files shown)


pkg/README.md
```diff
@@ -4,7 +4,7 @@
 -   [Overview](#overview)
 -   [Roadmap](#roadmap)
 -   [Reference](docs/reference.md)
```
nacl, line 6: Should this start with a slash as well?

pkg/README.md
```diff
@@ -21,21 +21,10 @@ and debian package.
...
-    name = "rules_pkg",
-    urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.4.0/rules_pkg-0.4.0.tar.gz",
-        "https://github.com/bazelbuild/rules_pkg/releases/download/0.4.0/rules_pkg-0.4.0.tar.gz",
-    ],
-    sha256 = "038f1caa773a7e35b3663865ffb003169c6a71dc995e39bf4815792f385d837d",
-)
-load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
-rules_pkg_dependencies()
-```
```
nacl, line 36: It might be worth leaving something here for how to use rules_pkg at HEAD, but it's not essential.

nacl (approved):
LGTM.  Comments are nitty or non-actionable.

pkg/README.md
aiuto: I could, but it would be /pkg/docs/reference.md.  docs is a sibling of this file, examples is a level up

PR #384, opened 2021-07-14
Modularize manifest python code

nacl:
The manifest code used in pkg_tar has uses outside of pkg_tar.  Let's modularize
it to reduce the number of places it has to be duplicated (it needs to be
duplicated in `pkg/private/pkg_files.bzl` to be loaded) into starlark.

As a part of this, we create an entry type stringifier for ManifestEntry, and a
`collections.namedtuple` representing the fields in the entry.

`pkg/private/build_tar.py` was also opportunistically modified to use all of the
above.

### DIFF
--- a/pkg/private/BUILD
+++ b/pkg/private/BUILD
@@ -116,3 +116,10 @@ py_binary(
         ":helpers",
     ],
 )
+
+py_library(
+    name = "manifest",
+    srcs = ["manifest.py"],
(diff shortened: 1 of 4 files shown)


aiuto (approved):
Thanks for the cleanup!
I love the PRs that are trivial to review and make future PRs smaller.
LMK if you are going to merge or you want me to.

PR #390, opened 2021-07-26
Add architecture_file parameter to pkg_deb

tnarg:
RELNOTES: Add `pkg_deb(architecture_file)`  to provide a way to set the Debian package architecture from the content of a file created at build time.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -271,9 +271,18 @@ def _pkg_deb_impl(ctx):
         "--changes=" + changes_file.path,
         "--data=" + ctx.file.data.path,
         "--package=" + ctx.attr.package,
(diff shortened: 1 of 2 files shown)


aiuto:
Is there a compelling need for this and an example of how you would use this?

At first glance, what I would prefer is a way to get the architecture dynamically, from something like --cpu?
Could you start an issue to describe the need, and we can come up with the best design.

tnarg:
I've opened https://github.com/bazelbuild/rules_pkg/issues/391 to provide more context around this change

aiuto (approved):

aiuto:
Thanks for filing the issues. It will provide great context for an example of how Go interacts with Bazel and packaging. If you would care to write one for the examples tree, that would be appreciated.

Issue #392, opened 2021-07-27
Runfiles are not included in pkg_tar.


kangnak:
It seems `include_runfiles = True` in pkg_tar() doesn't work after [the PR](https://github.com/bazelbuild/rules_pkg/pull/367), at least with py_binary. The resulting tar file doesn't include runfiles. And the release version 0.5.0 has the same issue.

kangnak:
The default behaviour from the version 0.4.0 is to flatten all files by removing directories:
```
./my_prog
./foo
```

This is suboptimal and doesn't work in many reasons. So we have to use `remap_paths` and `strip_pefix` to keep the directory structure:
```
pkg_tar(
    name = "my_prog_pkg",
    srcs = [":my_prog"],
    include_runfiles = True,
    remap_paths = {
        "my_app/my_prog.py": "my_prog.runfiles/<repo name>/my_app/my_prog.py",
        "my_app/my_prog": "my_prog",
        "my_app": "my_prog.runfiles/<repo name>/my_app",
        "external": "my_prog.runfiles",
    },
    strip_prefix = "/",
)
```
So I hope `include_runfiles = True` should not flatten files in the first place. However this might affect other people who are using the current default behaviour.

aiuto:
It sounds like what you want is to mimic the runfiles layout that Bazel produces.
That should be a distinct option, but it insufficient for the only choice, since it is unique to Bazel and does not match the expectations of any other packaging.

I have a PR ready that mimics the old behavior. That can go in after I add tests.

For the mimic bazel case, what I would do is move the functionality to a new rule, so it can be reused by all pkg_* rules.
```
pkg_expand_runfiles(
    name = "my_prog_runfiles",
    srcs = [":my_prog"],
)
pkg_tar(
    name = "my_prog_tar",
    srcs = ":my_prog_runfiles",
)
pkg_zip(
    name = "my_prog_zip",
    srcs = ":my_prog_runfiles",
)
```

kangnak:
Your PR(pkg_expand_runfiles) sounds good to me. Thanks for your works!

limdor:
@aiuto we cannot update to 0.5.0 because of this issue, do you have a plan on when 0.6.0 would be created? Is there any ticket that I could follow? Thanks

aiuto:
Can you try at the sources at head and see if they work for you?
There is a chance I can make a 0.5.1 patch over the next few days, but not
today.

limdor:
[...]

PR #393, opened 2021-07-28
Parameterize the manifest golden tests.

aiuto:
This was split out from #376 to make that review easier.
I want to get this in place before adding a fix for #392.
The UTF-8 tests are lower priority.

### DIFF
--- a/pkg/tests/mappings/manifest_test_lib.py
+++ b/pkg/tests/mappings/manifest_test_lib.py
@@ -32,10 +32,10 @@ class ContentManifestTest(unittest.TestCase):
     """
     e_file = ContentManifestTest.run_files.Rlocation(
         'rules_pkg/tests/mappings/' + expected)
-    with open(e_file, 'r') as e_fp:
+    with open(e_file, mode='rb') as e_fp:
       expected = json.load(e_fp)
     g_file = ContentManifestTest.run_files.Rlocation(
         'rules_pkg/tests/mappings/' + got)
-    with open(g_file, 'r') as g_fp:
+    with open(g_file, mode='rb') as g_fp:
       got = json.load(g_fp)
     self.assertEquals(expected, got)
--- a/pkg/tests/mappings/mappings_test.bzl
+++ b/pkg/tests/mappings/mappings_test.bzl
@@ -16,6 +16,7 @@
 
 load("@bazel_skylib//lib:new_sets.bzl", "sets")
 load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts", "unittest")
+load("@rules_python//python:defs.bzl", "py_test")
 load("//:providers.bzl", "PackageDirsInfo", "PackageFilegroupInfo", "PackageFilesInfo", "PackageSymlinkInfo")
 load(
     "//:mappings.bzl",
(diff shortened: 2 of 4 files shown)


pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+    implementation = _gen_manifest_test_main_impl,
+    attrs = {
+        "_template": attr.label(
+            default = Label("//tests/mappings:manifest_test_main.py.tpl"),
+            allow_single_file = True,
+        ),
+        "out": attr.output(mandatory = True),
+        "expected": attr.string(),
+        "target": attr.string(),
+        "test_name": attr.string(),
```
nacl, line 1003: Consider making these `mandatory` as well, given their use in `manifest_golden_test`.

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+    },
+)
+
+
+def manifest_golden_test(name, target, expected):
```
nacl, line 1008: Please add a docstring for this.

pkg/tests/mappings/manifest_test_lib.py
```diff
@@ -32,10 +32,10 @@ def assertManifestsMatch(self, expected, got):
     """
```
nacl, line 32: While you're in the neighborhood, consider setting `self.maxDiff` to something large (or `None`) in `setUp` to make any large differences more apparent.

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+def _gen_manifest_test_main_impl(ctx):
+    ctx.actions.expand_template(
+        template = ctx.file._template,
+        output = ctx.outputs.out,
+        substitutions = {
+            "${EXPECTED}": ctx.attr.expected,
+            "${TARGET}": ctx.attr.target,
+            "${TEST_NAME}": ctx.attr.test_name,
+        },
+    )
```
nacl, line 988: I've been meaning to publish something WRT https://github.com/bazelbuild/bazel-skylib/issues/191, which would obviate this setup, but never got around to it.

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+    return [
+        DefaultInfo(files = depset([ctx.outputs.out])),
+    ]
```
nacl, line 991: Since `out` is a predeclared output, isn't this unnecessary?

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+
+_gen_manifest_test_main = rule(
+    implementation = _gen_manifest_test_main_impl,
+    attrs = {
+        "_template": attr.label(
+            default = Label("//tests/mappings:manifest_test_main.py.tpl"),
+            allow_single_file = True,
+        ),
```
nacl, line 999: Shouldn't this be at the end of the list?

pkg/tests/mappings/mappings_test.bzl
```diff
@@ -974,3 +975,53 @@ def mappings_unit_tests():
...
+    )
+    return [
+        DefaultInfo(files = depset([ctx.outputs.out])),
+    ]
+
+_gen_manifest_test_main = rule(
+    implementation = _gen_manifest_test_main_impl,
+    attrs = {
+        "_template": attr.label(
+            default = Label("//tests/mappings:manifest_test_main.py.tpl"),
```
nacl, line 997: Couldn't this just be:

```
default = Label("manifest_test_main.py.tpl")
```

Or something similar?  The template file is in the same package as the rule, after all.

nacl (approved):
Nice incremental progress.  Suggestions are minor.

pkg/tests/mappings/mappings_test.bzl
aiuto: Done

pkg/tests/mappings/manifest_test_lib.py
aiuto: I want to change this to a more useful semantic diff. So I'll defer on that.

pkg/tests/mappings/mappings_test.bzl
aiuto: ack.

pkg/tests/mappings/mappings_test.bzl
aiuto: Moved

pkg/tests/mappings/mappings_test.bzl
aiuto: No. You need an absolute path label in this case.

pkg/tests/mappings/mappings_test.bzl
aiuto: Yes, but I think rules are more readable when they are explicit about their returned providers. Someone can see the implementation and know what the returns are, without having to go to the definition.

Issue #252, opened 2019-05-08
Option to suppress "Duplicate file in archive picking first occurrence" in pkg_tar.


u-quark:
### Description of the problem / feature request:

Option to suppress "Duplicate file in archive picking first occurrence" from pkg_tar

### Feature requests: what underlying problem are you trying to solve with this feature?

When creating an archive with `pkg_tar` or similar the warning for duplicate files "Duplicate file in archive <ARCHIVE> picking first occurrence" can be very noisy. It would be nice if there was a way to silence this. In our case all the duplicated files are the same, but there is no easy way to generate disjoint tars.

### Bugs: what's the simplest, easiest way to reproduce this bug? Please provide a minimal example if possible.

```
$ mkdir tmp
$ touch tmp/somefile
$ tar cf a.tar tmp/
$ tar cf b.tar tmp/
$ cat BUILD.bazel 
load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")

pkg_tar(
  name = "foo",
  deps = [":a.tar", ":b.tar"]
)
$ bazel build //:foo
INFO: Invocation ID: cea310a5-0315-4f88-b838-02b632df4378
INFO: Analysed target //:foo (1 packages loaded, 3 targets configured).
INFO: Found 1 target...
INFO: From PackageTar foo.tar:
Duplicate file in archive: ./tmp/somefile, picking first occurrence   <<<<<<<<<<<<<<
Target //:foo up-to-date:
  bazel-bin/foo.tar
INFO: Elapsed time: 1.383s, Critical Path: 0.69s
INFO: 1 process: 1 linux-sandbox.
INFO: Build completed successfully, 4 total actions
```

### What operating system are you running Bazel on?

```
$ uname -a
Linux 4.15.0-45-generic bazelbuild/bazel#48-Ubuntu SMP Tue Jan 29 16:28:13 UTC 2019 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

### What's the output of `bazel info release`?

```
INFO: Invocation ID: 9afcbd24-d8de-4caa-820a-13cbdd735116
release 0.22.0- (@non-git)
```

### If `bazel info release` returns "development version" or "(@non-git)", tell us how you built Bazel.

[...]

Issue #385, opened 2021-07-14
Package JSON manifest should support generic attributes


nacl:
I like the idea of the JSON manifest format -- it allows us to create backends for a generic packaging frontends.

I would like to adopt this for `pkg_rpm` at some point, but in order to do this, the manifest would need to support a generic "attributes" format to support file tags (like `%config`).  This will also be needed for supporting packages in Windows.

So, instead of separate `mode`, `user`, and `group` fields, perhaps we should have an `attributes` object that can contain arbitrary information, mirroring `pkg_attributes` in the `pkg_filegroup` framework?

aiuto:
We can do whatever we want.  Since there is, so far, 1 method that writes the manifest, and one that decodes an entry, and they are both private interfaces, changing the format is fair game as we need it.

Another reason for moving to that now might be that we may have to add more user/owner types for #370 . Rather than add more fields at the top level, add them to the embedded struct.

aiuto:
One more requirement: We need operators around this
- combine - which combines two attribute sets. For example, one has owner + group and the other has mode, yielding one with owner, group, & mode.
- compare for equality. - when we have the same file put in an archive twice, that is fine if the modes are the same. If the modes conflict, it should raise an error.

nacl:
Been thinking about this on and off.  Initial thoughts?

- The manifest is a JSON array containing a series of objects.  Each object represents an individual installable entity, as is known by Bazel.  Each key represents an attribute of the entity, each value represents the value of that entity.
- Each entity type is represented by a `type` field (a string or enumerated value).  This type dictates how the "core" attributes are interpreted.
- "Core" attributes include things like sources, in-package destinations, basic UNIX-style permissions (either octal or strings), and UNIX-style ownership.
- "Extended" attributes are related to specific entities (e.g. properties of UNIX-style device nodes), or specific packaging rules (e.g. Windows installers, or custom file tags in `pkg_rpm`)

So, something like this:

```jsonc
[
  {
    "label": "//:foo-config",
    "type": "FILE",
    "source": "bazel-out/k8-linux-fastbuild/foo", // or whatever bazel provides us
    "destination": "{PREFIX}/etc/foo", // Templated destination
    "mode": "rw-r--r--", // or 0644
    "user": "root",
    "group": "root",
    "rpm_filetag": "%config" // Custom attribute for pkg_rpm
    "zip_compression": "STORE" // Theoretical custom attribute for pkg_zip
[...]

PR #398, opened 2021-08-03
Restore `include_runfiles` support for pkg_tar.

### LINKED ISSUE #392: Runfiles are not included in pkg_tar.

aiuto:
Fixes #392

This does not address the deeper issues with include_runfiles not
being convenient to use. It just gets back to parity with the legacy
feature.

Also fix: Do not complain on duplicates that are safe.
Addresses part of #252

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -61,9 +61,14 @@ _DestFile = provider(
     },
 )
 
-def _check_dest(content_map, dest, origin):
-    if dest in content_map:
-        fail("Duplicate output path: <%s>, declared in %s and %s" % (
+def _check_dest(content_map, dest, src, origin):
+    old_entry = content_map.get(dest)
+    # TODO(aiuto): This is insufficient but good enough for now. We should
+    # compare over all the attributes too. That will detect problems where
+    # people specify the owner in one place, but another overly broad glob
+    # brings in the file with a different owner.
+    if old_entry and old_entry.src != src:
+        print("Duplicate output path: <%s>, declared in %s and %s" % (
             dest,
             origin,
             content_map[dest].origin,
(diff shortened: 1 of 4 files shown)


pkg/tests/BUILD
```diff
@@ -600,6 +601,20 @@ pkg_tar(
     extension = ".bz2",
 )
 
+fake_artifact(
+    name = "a_program",
+    files = ["//tests:testdata/executable.sh"],
+    runfiles = [":BUILD"],
```
nacl, line 607: Preceding colon is not necessary because "BUILD" is a source file, not a generated one.

pkg/private/pkg_files.bzl
```diff
@@ -61,9 +61,14 @@ _DestFile = provider(
...
 
-def _check_dest(content_map, dest, origin):
-    if dest in content_map:
-        fail("Duplicate output path: <%s>, declared in %s and %s" % (
+def _check_dest(content_map, dest, src, origin):
+    old_entry = content_map.get(dest)
+    # TODO(aiuto): This is insufficient but good enough for now. We should
+    # compare over all the attributes too. That will detect problems where
+    # people specify the owner in one place, but another overly broad glob
+    # brings in the file with a different owner.
```
nacl, line 69: Please create and mention a ticket for tracking this issue.  This does seem good enough for now, but this could needlessly surprise users.

Perhaps also mention that you made this change in the commit's summary line.

nacl (approved):
LGTM given the premise, and modulo comments.

Please also mention that you ran buildifier in the commit message.

pkg/tests/BUILD
aiuto: done

pkg/private/pkg_files.bzl
aiuto: I folded that requirement into #385

PR #399, opened 2021-08-04
Normalize import statements and make tests more robust to vendoring

aiuto:
These are non-functional changes to aid users who import the library to their own repo.

The correct way to import is to list the package, relative to the repo: E.g. `from private import helpers`

- none should use 'import rules_pkg.*', that breaks if you rename the repo
- raw imports are harder to find and automatically transform

And fix some tests so they can be imported into other places.

Tested by importing the repot into Google. Not every test builds and runs, but we are getting closer.


pkg/private/build_tar.py
```diff
@@ -19,10 +19,10 @@
...
 import tempfile
 
-import archive
-import helpers
-import build_info
-import manifest
+from private import archive
+from private import helpers
+from private import build_info
+from private import manifest
```
nacl, line 25: This is to make it easier for tools to figure out where paths are, given WORKSPACE-relative paths, right?

I'm theoretically fine with this, but it would seem odd to people who aren't used to this sort of scheme, myself included.  Consider documenting it somewhere.

nacl (commented):
I'm fine with the idea, but I think that this rule and the overall reasoning should be documented somewhere to reduce surprise during reviews.  Could this be enforced somehow?

There are others that might be missing here.  This one: https://github.com/bazelbuild/rules_pkg/blob/f7f2bd3c0fc7e891e6bd7c5503dd0bb95670a642/pkg/tests/rpm/BUILD#L286-L291 came to mind as I was browsing around.

pkg/private/build_tar.py
aiuto: Yes. It is mostly for people who import the code into their WORKSPACE. They might land it at `/third_party/bazel/rules_pkg` and want to normalize the paths so their Python code only has a single import path at the top.

I'm torn between a local guide to this or the overall "blaze ecosystem" style guide. I have an AI to take a crack at the latter.

aiuto:
> I'm fine with the idea, but I think that this rule and the overall reasoning should be documented somewhere to reduce surprise during reviews. Could this be enforced somehow?
> 
> There are others that might be missing here. This one:
> 
> https://github.com/bazelbuild/rules_pkg/blob/f7f2bd3c0fc7e891e6bd7c5503dd0bb95670a642/pkg/tests/rpm/BUILD#L286-L291
> 
> came to mind as I was browsing around.

Right... the imports=[.] is a particular problem for me. It does not exist in Blaze py_library. I have to map it out on import.

pkg/private/build_tar.py
nacl: Tracked in #412

nacl (approved):
I don't want to block your importing into Google's repos.  Feel free to submit.

Regardless, the import style should be documented and (preferably) enforced soon (#412).

PR #400, opened 2021-08-04
Split mappings tests that use an external repo from the others

aiuto:
Split mappings tests that use an external repo from the others.
This make it easier to exclude those tests from a `bazel test ...` pattern when used in a repo that does not allow WORKSPACE file customizations.

### DIFF
--- a/pkg/tests/mappings/mappings_external_repo_test.bzl
+++ b/pkg/tests/mappings/mappings_external_repo_test.bzl
@@ -1,0 +1,183 @@
+# Copyright 2021 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


pkg/tests/mappings/mappings_external_repo_test.bzl
```diff
@@ -0,0 +1,183 @@
...
+    _pkg_files_contents_test_impl,
+    attrs = {
+        # Other attributes can be tested here, but the most important one is the
+        # destinations.
+        "expected_dests": attr.string_list(
+            mandatory = True,
+        ),
+        "expected_attributes": attr.string(),
+    },
+)
```
nacl, line 71: Consider putting these functions in a local .bzl file sourced by both the local and external repository tests.

nacl (approved):
LGTM other than a potentially helpful code centralization thing

PR #409, opened 2021-08-14
Add version numbers to title pages of versioned docs

aiuto:
Also, remove robots meta tag from custom header. It should be only on the versioned sets, not on the top.

https://bazelbuild.github.io/rules_pkg/


nacl:
> Also, remove robots meta tag from custom header. It should be only on the versioned sets, not on the top.

This didn't seem to show up in the commit.  Is this because of the conflicts?

Doc pages are looking pretty good so far :)

aiuto:
Yes. Incomplete merging.  Pushed now.

nacl (approved):
> Yes. Incomplete merging. Pushed now.

I'm still not seeing the precise change.  It looks like there never was a "robots" section in `docs/_includes/head-custom.html`, but there is still one one in `docs/_includes/head.html`.

I know very little about Jeckyll.  Either the desired outcome has already been achieved (indexing only main), or nothing is being indexed.

Regardless, I don't want to block you further here.

aiuto:
The difference is small.  It puts 0.5.0 in the title of the page.

PR #418, opened 2021-09-12
fix multiarch deb package generation

motiejus:
When 2 deb packages are defined with the same name and version, but
different arch (e.g. `amd64` and `arm64`), the `changes` file is
declared on both. Excerpt from the attached test case, but unpatched
`pkg.bzl`:

```
ERROR: file 'tests/test_deb_multiarch_1.changes' is generated by these conflicting actions:
```

This is because the name of the changes file is calculated without
taking the dots into account, which are frequently part of the version.

The patch takes the dots in version into account, and changes the
filename only by the last dot, of the presumed `.deb`.


aiuto:
This is not really about multi-arch. It's about not making the right name for the .changes file if the version has a . in it.  I filed bug #419 for that.

Can you devise a test that the filename is really of the expected format, and update the title and description to reflect that instead of multi-arch.  I think you might be able to do that with an analysis test that looks at a pkg_deb rule and examines the `OutputGroupInfo` to get the `changes` section. No rush, I am going to be unavoidably unavailable much of this week, so I won't have a chance to review until the 17th at the earliest.

aiuto (changes requested):
This is not really about multi-arch. It's about not making the right name for the .changes file if the version has a . in it.  I filed bug #419 for that.

Can you devise a test that the filename is really of the expected format, and update the title and description to reflect that instead of multi-arch.  I think you might be able to do that with an analysis test that looks at a pkg_deb rule and examines the `OutputGroupInfo` to get the `changes` section. No rush, I am going to be unavoidably unavailable much of this week, so I won't have a chance to review until the 17th at the earliest.

motiejus:
Thank you for your feedback.

> Can you devise a test that the filename is really of the expected format, and update the title and description to reflect that instead of multi-arch. I think you might be able to do that with an analysis test that looks at a pkg_deb rule and examines the OutputGroupInfo to get the changes section. No rush, I am going to be unavoidably unavailable much of this week, so I won't have a chance to review until the 17th at the earliest.

Looks like I'll need to create `analysistest` for `pkg_deb`, which there are none at the moment. This sounds non-trivial (this is the first time I have been reading about those), so it will take some time.

motiejus:
> > Can you devise a test that the filename is really of the expected format, and update the title and description to reflect that instead of multi-arch. I think you might be able to do that with an analysis test that looks at a pkg_deb rule and examines the OutputGroupInfo to get the changes section. No rush, I am going to be unavoidably unavailable much of this week, so I won't have a chance to review until the 17th at the earliest.
> 
> Looks like I'll need to create `analysistest` for `pkg_deb`, which there are none at the moment. This sounds non-trivial (this is the first time I have been reading about those), so it will take some time.

Heads up, I am not able to spend more time on the new test case in the foreseeable future.

aiuto:
I made #425 to refactor out the pkg_deb specific behaviors. Any tests should be done in tests/deb after that PR goes in.

aiuto (approved):

aiuto:
The checks are failing for known reasons not caused by this PR.

PR #420, opened 2021-09-14
Enable nested pkg_filegroups

### LINKED ISSUE #415: Nesting of `pkg_filegroup` is not possible
I am using `pkg_filegroup` on sources that are a `pkg_filegroup` again.
This fails with
```
in srcs attribute of pkg_filegroup rule //:codeql-cpp: '//odasa-buildutils:dist-generic' does not have mandatory providers: 'PackageFilesInfo' or 'PackageDirsInfo' or 'PackageSymlinkInfo'.
```

Is there any specific reason that this is not allowed?
Using `pkg_files` everywhere (except at the top-level) is legal starlark, but needs very careful `strip_prefix` handling to not have too many/the wrong prefixes (or I don't understand `strip_prefix` well enough).

beasleyr-vmw:
This commit enables users to feed a `pkg_filegroup` into another
`pkg_filegroup`.  Starlark tests included.

Resolves #415.

Testing Done:
- `bazelisk test tests/mappings/...`

### DIFF
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -250,10 +250,11 @@ def _pkg_files_impl(ctx):
                     "REMOVE_BASE_DIRECTORY as a renaming target for non-directories is disallowed.",
                     "renames",
                 )
+
             # REMOVE_BASE_DIRECTORY results in the contents being dropped into
             # place directly in the prefix path.
             src_dest_paths_map[src_file] = ctx.attr.prefix
-            
+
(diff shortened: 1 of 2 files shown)


nacl:
@aiuto could you take a peek at this?  I looked at this in an internal review, and it looked fine to me.

aiuto (approved):
Is there any chance you could extend the rich_structure example to show how this works?

Issue #216, opened 2020-08-07
Python 3.8 changed default tar format from GNU to PAX. Respond appropriately.


aiuto:
python3.8 changed the default for tarfile from GNU to PAX.
1. this breaks pkg_deb, because dpkg-info expects GNU format.
1. this has implications with long file names in general tarballs.

Fixing #1 is easy, just change the write mode for that instance.
Fixing #2 is less clear.

aiuto:
Oh.. That could do it.

aiuto:
The fix I am going to go with is:
1. changing the default for pkg_tar back to GNU
2. adding a capability to specify PAX format in pkg_tar.

That's a breaking change, but for the better. Most people won't care, and it fixes the broken debian packages when the file names are long.

PR #422, opened 2021-09-24
[pkg_tar] specify tar format

motiejus:
Python's tarfile.DEFAULT_FORMAT changed from GNU to Pax from Python 3.8
onwards ([doc][1]):

```
tarfile.DEFAULT_FORMAT
The default format for creating archives. This is currently PAX_FORMAT.
```

Hardcode *a particular format* (the old one, for wider compatibility with what's out there), so the produced tars are more hermetic against the differences in the host system.

This is of particular interest when tarballs are combined with `pkg_tar` rule:

```
pkg_tar(
  name = "other",
  srcs = [...],
)

pkg_tar(
  name = "final",
  deps = [
      "//src/...", # assume it was generated by a tool which emits a gnu-tarball
      ":other",
  ]
)
```

Since the archives of `;final` are not repacked, just appended, the resulting concatenated tar contains archives of different types, which, for exampe, `dpkg` is complaining about. A tar with mixed format is better to be avoided, esp. that the messages are quite cryptic:

```
dpkg: error processing archive /<...>.deb (--install):
 corrupted filesystem tarfile in package archive: unsupported PAX tar header type 'x'
```

[1]: https://docs.python.org/3/library/tarfile.html#tarfile.DEFAULT_FORMAT

### DIFF
--- a/pkg/private/archive.py
+++ b/pkg/private/archive.py
@@ -177,7 +177,8 @@ class TarFileWriter(object):
     self.root_directory = root_directory.rstrip('/').rstrip('\\')
     self.root_directory = self.root_directory.replace('\\', '/')
 
-    self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj)
+    self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj,
+                            format=tarfile.GNU_FORMAT)
     self.members = set([])
     self.directories = set([])
 


aiuto:
I've been tied up this week on a project. I can look soon.

But, the failure of the integration tests makes me suspect there is a subtle behavior change to that should be tested in some way.

motiejus:
> But, the failure of the integration tests makes me suspect there is a subtle behavior change to that should be tested in some way.

The same test is failing for #418, another PR of mine. The stack trace says:

```
UnicodeEncodeError: 'ascii' codec can't encode character '\u0161' in position 186: ordinal not in range(128)
```

`\u0161` is `š`, which is in in my last name. I strongly suspect the integration tests refuse to accept my surname in the git log.

aiuto:
I am not inclined to take this PR without addressing #216.  
That is, any solution must prove that we do the right thing w.r.t. various Python versions.
This will probably mean more tests, or at least a good analysis of why a change is correct.

nacl:
I agree with @aiuto.  #216 really needs to be solved properly first.  Perhaps the tarball format should be an attribute passed to `pkg_tar`, perhaps defaulting to GNU?

Then, we can properly solve the issue with `pkg_deb`.  IMO, `pkg_deb` should no longer give users the option to provide their own data tarball, and it instead should create the tarball from `pkg_filegroup`s and friends with the GNU tarball format.

> That is, any solution must prove that we do the right thing w.r.t. various Python versions.

One thing we could do to help with this is start integrating Ubuntu 20.04 into our CI -- it uses Python 3.8, whereas Ubuntu 18.04 has Python 3.6.

aiuto:
> > That is, any solution must prove that we do the right thing w.r.t. various Python versions.
> > This will probably mean more tests, or at least a good analysis of why a change is correct.
> 
> I can write a couple of Dockerfiles (with different Python versions) that build and install a package. However, I don't see how it could be added to CI. Would that be sufficient?

No. I want to understand how this impacts users with different python versions.  There are still a lot of people at python <3.8 and I want to understand that any solution works for them too.

aiuto:
Closing as obsolete. This was done in a different PR and the code exists in a different place noe.

PR #428, opened 2021-10-04
add recurse option to pkg_zip to maintain directory structure

hborawski:
Need to maintain the zip directory structure for zipping cocoapods as a part of the build, otherwise the glob wont match from the podspec file (or i'd end up matching test files from a test_spec as well as source files)

```ruby
...
s.source_files = "SomePod/Sources/**/*.swift"
s.test_spec do |tests|
  tests.source_files = "SomePod/Tests/**/*.swift"
end
```

`pkg_zip` with `glob(["SomePod/Sources/**/*.swift"])` will produce a zip with all swift files in a flat directory

I can solve this simple example with `package_dir = 'SomePod/Sources'` but looking for a longer term solution because i want to use this on pods with multiple subspecs

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -593,8 +593,6 @@ def _pkg_zip_impl(ctx):
     args.add("-d", ctx.attr.package_dir)
     args.add("-t", ctx.attr.timestamp)
     args.add("-m", ctx.attr.mode)
-    if ctx.attr.recurse:
-        args.add("-r")
     inputs = []
     if ctx.attr.stamp == 1 or (ctx.attr.stamp == -1 and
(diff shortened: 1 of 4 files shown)


aiuto:
Can you create an issue that describes the broader problem?
My inclination is to not even review this. The main reason is that it is a special hack to pkg_zip, and not reusable across all the other package types.  This should be implemented as extensions to pkg_files, if anywhere.

aiuto (changes requested):
Needs discussion first.

hborawski:
issue created: https://github.com/bazelbuild/rules_pkg/issues/434

hborawski:
Closing due to resolution in #434

PR #430, opened 2021-10-05
Refactor presubmit tests for LTS & head.

aiuto:


.bazelci/examples_naming.yml
```diff
@@ -1,19 +1,16 @@
 common: &common
+  working_directory: ../examples/naming_package_files
   build_targets:
   - "..."
 
 tasks:
-  centos7:
+  centos:
```
nacl, line 7: IMO, calling out the version is helpful, especially if we start testing on, e.g., multiple versions of Ubuntu.

Analogous changes are also not done in multiple places below.

.bazelci/integration.yml
```diff
@@ -1,10 +1,23 @@
...
   build_targets:
   - "distro:*"
 
 tasks:
-  ubuntu1804:
+  rolling_ubuntu:
+    name: rolling_distro
+    <<: *rolling
+    <<: *common
+  lts_ubuntu:
```
nacl, line 19: Amusingly, Ubuntu 18.04 could also be considered "LTS ubuntu", so perhaps the name could be clarified a bit?

It's clear to me what's going on (and buildkite looks fine), so I don't feel too strongly about it.

PR #436, opened 2021-10-15
Allow tarball_name as parameter

### LINKED ISSUE #435: `print_rel_notes` should support customizing `tarball_name`
The `tarball_name` is currently hard coded in `print_rel_notes` to follow a strict pattern of `repo` name and `version`. This does not support the case where different name is used for the tar file. The call/build then fails with missing input file.

guw:
Fixes #435

### DIFF
--- a/pkg/distro/BUILD
+++ b/pkg/distro/BUILD
@@ -40,6 +40,7 @@ pkg_tar(
         "//private:standard_package",
         "//releasing:standard_package",
         "//toolchains:standard_package",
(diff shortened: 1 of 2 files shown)


aiuto (changes requested):
Sorry I didn't ask that right.  What I  meant was why not package_file_name as the attribute?

Whatever we do here, tarball is the wrong name because we should allow packages of other types, not just tar.   Before updating this CL, let's explore the correct solution in the bug.

PR #438, opened 2021-10-15
Make label reference absolute

### LINKED ISSUE #437: `git_changelog` cannot be used from outside
If you want to use git_changlog outside of this reop it fails with a missing target

eg. `BUILD`:
```
load("@rules_pkg//releasing:git.bzl", "git_changelog")
...
git_changelog(
    name = "changelog",
    out = "changelog.txt",
)
```

`WORKSPACE`:
```
# rules_pkg dused to produce the release artifacts.
http_archive(
    name = "rules_pkg",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
        "https://github.com/bazelbuild/rules_pkg/releases/download/0.5.1/rules_pkg-0.5.1.tar.gz",
    ],
    sha256 = "a89e203d3cf264e564fcb96b6e06dd70bc0557356eb48400ce4b5d97c2c3720d",
)
load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
rules_pkg_dependencies()

# Needed for making our release notes
load("@rules_pkg//toolchains/git:git_configure.bzl", "experimental_find_system_git")
experimental_find_system_git(
    name = "my_git",
    workspace_file = "//:WORKSPACE", 
    verbose = True,
)
```

It fails with:
```
DEBUG: .../external/rules_pkg/toolchains/git/git_configure.bzl:34:18: Found git at '/usr/bin/git'
DEBUG: .../external/rules_pkg/toolchains/git/git_configure.bzl:42:14: Found WORKSPACE in ...
ERROR: Analysis of target '//dist:relnotes' failed; build aborted: no such package 'toolchains/git': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /path/to/my/local/bazel_workspace/toolchains/git
```

guw:
Fixes #437

### DIFF
--- a/pkg/releasing/git.bzl
+++ b/pkg/releasing/git.bzl
@@ -89,6 +89,7 @@ def git_changelog(name, **kwargs):
         name = name,
         # This requires bazel 4.x
         target_compatible_with = select({
+            # Force label resolution to be rules_pkg, instead of my repo.
             str(Label("//toolchains/git:have_git")): [],
             "//conditions:default": ["//:not_compatible"],
         }),


pkg/releasing/git.bzl
```diff
@@ -89,7 +89,7 @@ def git_changelog(name, **kwargs):
         name = name,
         # This requires bazel 4.x
         target_compatible_with = select({
-            "//toolchains/git:have_git": [],
+            "@rules_pkg//toolchains/git:have_git": [],
```
aiuto, line 92: Won't this break people who rename the repo to a new name on import?
Have you tried doing that?

aiuto (changes requested):

pkg/releasing/git.bzl
guw: Pretty sure it will break. It looks like a flaw in Bazel tbh. I thought that all the refs/labels should be resolved within the repo - which would be the external (imported) repo. But apparently it's not. It tries to resolve those from within the host workspace, which seem ood.

aiuto: @@ -89,7 +91,7 @@ def git_changelog(name, **kwargs):
         name = name,
         # This requires bazel 4.x
         target_compatible_with = select({
-            "//toolchains/git:have_git": [],
+            str(Label("//toolchains/git:have_git")): [],
             "//conditions:default": ["//:not_compatible"],
         }),

pkg/distro/BUILD
```diff
@@ -40,6 +40,7 @@ pkg_tar(
         "//private:standard_package",
         "//releasing:standard_package",
         "//toolchains:standard_package",
+        "//toolchains/git:standard_package",
```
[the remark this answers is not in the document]
aiuto: I guess so.  But I am actually dubious about the whole thing.
The hoops I jump through to get bazel to run git in the source tree instead of a sandbox are a little strange. OTOH, even if I toss out the toolchain and change it to provide tools that run git manually, standard_package would still be part of it.

PR #442, opened 2021-10-19
Add "file" field to PackageArtifactInfo

### LINKED ISSUE #433: PackageArtifactInfo should contain the named file (File), not the name of the file (string)
@aiuto Do you recall why `PackageArtifactInfo` only contains a string containing the name of the file, and not the File itself?

https://github.com/bazelbuild/rules_pkg/blob/8d542763a3959db79175404758f46c7f3f385fa5/pkg/providers.bzl#L17-L23

When working with this internally, I found I needed to jump through a few hoops to get the named artifact out of the `PackageArtifactInfo` using the (string) file name.  If the value was a `File` instead, acquiring the named artifact would be instantaneous.

This can be done without a breaking change by (in documentation) tagging `file_name` as deprecated.

I have a prototype locally that does something like this, and it seems to work well.  WDYT?

nacl:
This change adds the `file` field to the `PackageArtifactInfo` provider.  This
allows consuming rules to directly access the named artifact file, as opposed to
having to search the artifact target's `DefaultInfo` for it.

`PackageArtifactInfo.file_name` is now deprecated.  The same information can be
acquired from `PackageArtifactInfo.file`.

Basic tests for the validity of this field have been added to `pkg_rpm`.  I can look into adding tests for the others if desired.

This doesn't solve all the problems with `PackageArtifactInfo`, but it's a step in the right direction.

Fixes #433.

### DIFF
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -577,9 +577,9 @@ def _pkg_rpm_impl(ctx):
             files = depset(outputs),
         ),
         PackageArtifactInfo(
-            label = ctx.label.name,
             file = output_file,
             file_name = output_name,
+            label = ctx.label.name,
         ),
     ]
 


aiuto:
```
'workspace_stanza': workspace_stanza,
--
  | UnicodeEncodeError: 'ascii' codec can't encode character '\u2026' in position 1420: ordinal not in range(128)
```
Cool. Unicode problems. I hope this is because I just added the non latin1 filenames to the test data

aiuto:
I'm merging even though the integration tests are failing. That is clearly the non-ASCII in the changelog.

PR #445, opened 2021-10-22
Move pkg_deb documentation from .md file to .bzl

aiuto:
I took most of the words from here: https://github.com/bazelbuild/rules_pkg/blob/main/pkg/docs/reference.md#pkg_deb
and moved them into pkg.bzl, so you can review them at
https://github.com/bazelbuild/rules_pkg/pull/445/files?short_path=e1fc4a7#diff-e1fc4a7da21e1351984b72ad49949ca6e2c3fe7487ff3cabb8de9fc095a53ecd

What I really miss is the collapsing of the  foo + foo_file structure that the hand-written docs can have.
See  architecture/architecture_file  vs.  built_using/built_using_file for an example.  I can live with the stardoc way for now, but I am seriously considering building my own markdown renderer so I can do better than the current one.

- The output in docs/pkg_deb.md will not be included in the actual commit.  It is just to make the review easier.
- Other reformatting in pkg.bzl is the result of a buildifier run. They are not behavior changing.


pkg/pkg.bzl
```diff
@@ -503,68 +504,171 @@ def pkg_tar(name, **kwargs):
 pkg_deb_impl = rule(
     implementation = _pkg_deb_impl,
     attrs = {
```
nacl, line 506: Maybe add a comment with `# @unsorted-dict-items` to prevent aggressive `buildifier` runs from resorting this?

pkg/pkg.bzl
```diff
@@ -503,68 +504,171 @@ def pkg_tar(name, **kwargs):
...
         # Common attributes
-        "out": attr.output(mandatory = True),
-        "package_file_name": attr.string(doc = "See Common Attributes"),
+        "out": attr.output(
+            doc = """See Common Attributes""",
+            mandatory = True
+        ),
+        "package_file_name": attr.string(
+            doc = """See Common Attributes.
+            Default: "%{package}-%{version}-%{architecture}.deb""",
```
nacl, line 663: The [reference](https://github.com/bazelbuild/rules_pkg/blob/main/pkg/docs/reference.md#common-attributes) suggests that the `%`'s shouldn't be here, perhaps.

nacl:
> What I really miss is the collapsing of the foo + foo_file structure that the hand-written docs can have.
See architecture/architecture_file vs. built_using/built_using_file for an example. I can live with the stardoc way for now, but I am seriously considering building my own markdown renderer so I can do better than the current one.

Hah, yeah, that's fair.  It might also be nice if there was `mutually_exclusive_with` flag that allows a documentation system to automatically specify this.

pkg/pkg.bzl
```diff
@@ -503,68 +504,171 @@ def pkg_tar(name, **kwargs):
...
+        ),
+        "description": attr.string(
+            doc = """The package description. Must not be used with `description_file`.""",
+        ),
+        "description_file": attr.label(
+            doc = """The package description. Must not be used with `description`.""",
+            allow_single_file = True
+        ),
+        "distribution": attr.string(
+            doc = "TBD",
```
nacl, line 550: These TBD docstrings should probably be set somewhere, methinks

nacl (commented):
LGTM other than some nits.

nacl (approved):
Actually approve

pkg/pkg.bzl
aiuto: Done

pkg/pkg.bzl
aiuto: Removed

pkg/pkg.bzl
aiuto: I pointed to the debian docs.

PR #446, opened 2021-10-24
refactor pkg_deb

aiuto:
Refactor debian building so it can be more easily owned by others.

blocked by #443

cc: @alexeagle


aiuto:
This is ready again, now that the big reorganization PR is merged.

CODEOWNERS
```diff
@@ -1,4 +1,11 @@
 * @aiuto @nacl
 
-/deb_packages/  @MarkusTeufelberger @adragomir
-/pkg/rpm/ @nacl
+/deb_packages/  @aiuto
+/docs		@aiuto @nacl @floriographygoth
+
+/pkg/rpm/	@nacl
+/pkg/tests/rpm/	@nacl
+legacy/tests/rpm	@nacl
```
nacl, line 8: So, what's the exact policy on approvers here?  It seems that not having multiple CODEOWNERS could be risky if a change must be reviewed by a CODEOWNER.

distro/BUILD
```diff
@@ -39,6 +39,7 @@ pkg_tar(
         "//:standard_package",
         "//pkg:standard_package",
         "//pkg/private:standard_package",
+        "//pkg/private/deb:standard_package",
```
nacl, line 42: What about inverting the structure a bit: `//pkg/deb/private`?  Would reduce the complexity of CODEOWNERS somewhat.

This is fine as-is though; just wondering your thoughts.

distro/testdata/BUILD.tpl
```diff
@@ -1,4 +1,4 @@
-load("@not_named_rules_pkg//:pkg.bzl", "pkg_tar")
+load("@not_named_rules_pkg//pkg:pkg.bzl", "pkg_tar")
```
nacl, line 1: Was this missed in #443?

pkg/README.md
```diff
@@ -40,8 +40,8 @@ find_system_rpmbuild(name = "rules_pkg_rpmbuild")
 This example is a simplification of the debian packaging of Bazel:
 
 ```python
-load("@rules_pkg//pkg:pkg.bzl", "pkg_tar", "pkg_deb")
-
+load("@rules_pkg//pkg:pkg.bzl", "pkg_tar")
+load("@rules_pkg//pkg:pkg_deb.bzl", "pkg_deb")
```
nacl, line 44: I wonder if this file could just be be called "deb.bzl" instead of "pkg_deb.bzl".

IMO, it's a bit redundant to have `pkg` said twice in the label.  As an example of prior art,`rpm.bzl` and `rpm_pfg.bzl` contain `pkg_rpm`.

nacl:
We don't happen to have tests that exercise the distro tarball, would we?  A simple way to do this would be something that would build the distro, extract it somewhere, plant the tests inside the extracted location, and then run the tests.

nacl (changes requested):
I think we need to confirm the policy on CODEOWNERs first so we prevent shooting ourselves in the foot.

LGTM otherwise, will approve once we come to a consensus on it and naming.

aiuto:
> We don't happen to have tests that exercise the distro tarball, would we? A simple way to do this would be something that would build the distro, extract it somewhere, plant the tests inside the extracted location, and then run the tests.

Yes. The tests under distro do do that. They make sure we can build some things when using the tarball, and under a different repository name. 
We could probably add finer grained test in the future.

pkg/README.md
aiuto: Good point. I'll rename.

distro/testdata/BUILD.tpl
aiuto: Yes. But becaused of the backwrads compatibility, it still worked.
I tighented it up here.

distro/BUILD
aiuto: Too much in my head to think about that right now. That is time for another day.

CODEOWNERS
aiuto: I think the policy works, in that you can't approve yourself.  But I don't actually know.
But.... once we do it, we can observe the behavior and correct it.

aiuto: I find the only way to figure out some of the github interactions is to iteratively submit and see.

aiuto:
> I think we need to confirm the policy on CODEOWNERs first so we prevent shooting ourselves in the foot.

What particular foot shooting are you thinking of?  
The current situation is that you and I are the only codeowners, and with the current branch protection policy, that means a PR from either of us is blocked on the other.

Adding extra owners will allow a domain expert to review and approve a change. If it is strictly to RPM or DEB packaging (or maybe even tar & zip) I welcome the ability to have someone else be on the hook to review things.
That's what this will do.

nacl:
> What particular foot shooting are you thinking of? The current situation is that you and I are the only codeowners, and with the current branch protection policy, that means a PR from either of us is blocked on the other.
> 
> Adding extra owners will allow a domain expert to review and approve a change. If it is strictly to RPM or DEB packaging (or maybe even tar & zip) I welcome the ability to have someone else be on the hook to review things. That's what this will do.

I was mostly concerned regarding exclusivity to owners -- that if I were the only owner of the RPM code and I were unavailable, then a change couldn't be approved without editing the CODEOWNERs files.

Reading more closely, I see that this is more of an overlay and not really a problem.  Will approve momentarily.

distro/BUILD
nacl: Regardless of our current decision, given that this is an implementation detail, we can decide this later if we find it doesn't scale well.

distro/testdata/BUILD.tpl
nacl: Cool, thanks.

nacl (approved):
My concerns are addressed in comments and code updates.  LGTM.

PR #448, opened 2021-10-25
turn on use of target_compatible_with for make_rpm.

aiuto:
Let's see what happens.
Note: This would have smaller impact if we refactored rules so rpm were distinct.

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -58,11 +58,10 @@ py_binary(
     srcs = ["make_rpm.py"],
     python_version = "PY3",
     srcs_version = "PY3",
-    # TODO(aiuto): Enable this for bazel 4.x
-    #target_compatible_with = select({
-    #    "//toolchains:have_rpmbuild": [],
-    #    "//conditions:default": [":not_compatible"],
-    #}),
+    target_compatible_with = select({
+        "//toolchains:have_rpmbuild": [],
+        "//conditions:default": [":not_compatible"],
+    }),
     visibility = ["//visibility:public"],
     deps = [
         "//private:archive",


pkg/BUILD
```diff
@@ -58,11 +58,10 @@ py_binary(
...
     srcs_version = "PY3",
-    # TODO(aiuto): Enable this for bazel 4.x
-    #target_compatible_with = select({
-    #    "//toolchains:have_rpmbuild": [],
-    #    "//conditions:default": [":not_compatible"],
-    #}),
+    target_compatible_with = select({
+        "//toolchains:have_rpmbuild": [],
+        "//conditions:default": [":not_compatible"],
+    }),
```
nacl, line 64: This doesn't _quite_ do what we want it to, but it's a definite improvement.  Without a local copy of `rpmbuild`, I get a bunch of skipped tests, but I get this for a few of them that depend on the outputs of the action that calls `rpmbuild`:

```console
> USE_BAZEL_VERSION=rolling  bazelisk test //tests/rpm:analysis_tests_naming_with_different_name_rpm
DEBUG: /private/var/tmp/_bazel_apsaltis/297e6fd0ec15ec7901f2a2bf25d17005/external/rules_pkg/toolchains/rpmbuild_configure.bzl:35:16: No system rpmbuild found.
DEBUG: /private/var/tmp/_bazel_apsaltis/297e6fd0ec15ec7901f2a2bf25d17005/external/rules_pkg/toolchains/git/git_configure.bzl:34:18: Found git at '/opt/local/bin/git'
DEBUG: /private/var/tmp/_bazel_apsaltis/297e6fd0ec15ec7901f2a2bf25d17005/external/rules_pkg/toolchains/git/git_configure.bzl:42:14: Found WORKSPACE in /Users/apsaltis/workspace-git/rules_pkg/pkg
ERROR: Target //tests/rpm:analysis_tests_naming_with_different_name_rpm is incompatible and cannot be built, but was explicitly requested.
Dependency chain:
    //tests/rpm:analysis_tests_naming_with_different_name_rpm
    //:make_rpm   <-- target platform (@local_config_platform//:host) didn't satisfy constraint //:not_compatible
INFO: Elapsed time: 14.556s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (50 packages loaded, 645 targets configured)
```

pkg/BUILD
```diff
@@ -58,11 +58,10 @@ py_binary(
     srcs = ["make_rpm.py"],
     python_version = "PY3",
     srcs_version = "PY3",
-    # TODO(aiuto): Enable this for bazel 4.x
-    #target_compatible_with = select({
-    #    "//toolchains:have_rpmbuild": [],
-    #    "//conditions:default": [":not_compatible"],
-    #}),
+    target_compatible_with = select({
```
nacl, line 61: I feel this really should be `exec_compatible_with`, but that's apparently not a configurable attribute.  Might be worth adding a comment here?

I suppose this goes back to the point of being able to define platform constraints for the `exec` configuration dynamically.

nacl (approved):
I think this is a good thing, but raises a few more questions.

pkg/BUILD
aiuto: Well. Maybe I can make that work.   I'll to some playing with it.

nacl:
@aiuto, have you had a chance to look at my comments above?

PR #454, opened 2021-10-28
Revert "Remove uses of `incompatible_use_toolchain_transition` now th…

katre:
…at it is enabled by (#452)"

This is still needed for bazel 4.x compatibility, and it a no-op in
Bazel 5.x, so it should be kept.

The API will be removed in Bazel 6.0.

Part of bazelbuild/bazel#14127.

This reverts commit e203d5b32bb0b8d1040543514d421f7f9cd944f5.


nacl (approved):
LGTM; thanks!

Issue #456, opened 2021-11-04
pkg_rpm should accept mapping providers (from pkg_files, etc.) directly 


nacl:
My original implementation of `pkg_rpm` takes a strict approach to `pkg_filegroup`s -- specifically, it only accepts `pkg_filegroup`s or targets with the associated provider.

Other rules implementations have gone for accepting both grouping providers (`PackageFilegroupInfo`) and mapping providers (`PackageFilesInfo`, etc).

I've grown to finding the more relaxed approach highly effective, so let's make this the case globally.

PR #458, opened 2021-11-04
Consolidate pkg_rpm input processing loop

### LINKED ISSUE #191: pkg/experimential/rpm.bzl: Consolidate dupe check and install script provider loops
@aiuto mentioned his [review of pkg_mklinks](https://github.com/bazelbuild/rules_pkg/pull/175/files/ec802488edc1a0594b019fadf1794f13b434b365#r430594653) that the loops for duplicate checks and install script creation could be consolidated to be more performant and provide better information.

The main reason for separating this originally was to make the loops cleaner, and it kinda grew organically from there.

Any solution to this problem should try to prevent the loops from being too long and overall reducing code duplication.  It may be time to introduce some metaprogramming with `getattr` and `hasattr` and friends.

nacl:
The original `pkg_filegroup`-using `pkg_rpm` design had separate loops for
conflict detection and contents processing.  This change consolidates the two
loops.  The combined loop is a bit long, but the code around this point needs to
be adjusted for #456 anyway, so the length problem will be dealt with then.

Much of this code can be eliminated when we move pkg_rpm to use the JSON
manifests.

Fixes #191.

### DIFF
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -354,57 +354,12 @@ def _pkg_rpm_impl(ctx):
 
     files += ctx.files.srcs
 
-    #### Consistency checking
+    #### Consistency checking; input processing
 
     # Ensure that no destinations collide.  RPMs that fail this check may be
     # correct, but the output may also create hard-to-debug issues.  Better to
     # err on the side of correctness here.
     dest_check_map = {}
-    for dep in ctx.attr.srcs:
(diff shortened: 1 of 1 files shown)


aiuto (approved):
I'm going to assume this is right, based on the extensive tests continuing to pass.

nacl:
> I'm going to assume this is right, based on the extensive tests continuing to pass.

Yeah, that's effectively all the testing I did for this change.  Consolidated the loop in chunks, reran tests.

Issue #459, opened 2021-11-04
pkg_rpm should use the JSON manifest system


nacl:
pkg_rpm was the prototype for using pkg_filegroups for inputs.  Other packaging rules have moved on to a common JSON-style manifest, but pkg_rpm still uses its own processing loop.

We should be able to use the same manifest everywhere, but `pkg_rpm` cannot use it until we resolve #385.

PR #460, opened 2021-11-04
Allow pkg_rpm to accept mapping providers directly

### LINKED ISSUE #456: pkg_rpm should accept mapping providers (from pkg_files, etc.) directly

nacl:
`pkg_rpm` initially took a strict stance on the `pkg_filegroup` framework.  In
particular, it would only accept `PackageFilegroupInfo`s, and not any of the
other providers.

It was found that direct access to the mapping
providers (e.g. `PackageFilesInfo`) was highly beneficial.  This change
implements this for `pkg_rpm`.  A large amount of the provider-handling
functionality is now abstracted into functions.

The corresponding error message handling logic was improved, and tests were
provided.

Some reordering was also done by `buildifier`.

Depends on #458.

Resolves #456.


pkg/rpm_pfg.bzl
```diff
@@ -129,6 +142,72 @@ def _make_absolute_if_not_already_or_is_macro(path):
     # this can be inlined easily.
     return path if path.startswith(("/", "%")) else "/" + path
 
+#### Input processing helper functons.
+
+# TODO(nacl, #459): These are redundant with functions and structures in
+# pkg/private/pkg_files.bzl.  We should really use the infrastructure provided
+# there, but as of writing, it's not quite ready.
```
aiuto, line 149: What's the delta?  Is it worth changing there first before this?

nacl: The main reason for doing this now is I am currently in need this internally to ease migration to the latest and greatest rules_pkg.  I was able to whip this up and fully test it in 90 minutes, and to fully do #459 (which would supplant this change altogether) would take significantly longer.

It is mostly blocked by #385, but even though I could perhaps start without, we would have to rework and/or rewrite much of `pkg_rpm` and `make_rpm.py`.  This change gives us closer API compatibility and reduces user surprise until we can find time to implement the desired backend changes.

aiuto: SGTM

aiuto (approved):

pkg/rpm_pfg.bzl
nacl: Thanks!

PR #461, opened 2021-11-08
Split pkg_zip implementation out of pkg.bzl

aiuto:
Another refactoring in the model of pkg_deb.


pkg/private/BUILD
```diff
@@ -57,7 +57,7 @@ py_library(
     ],
     srcs_version = "PY3",
     visibility = [
-        "//pkg:__pkg__",
+        "//:__subpackages__",
```
nacl, line 60: Could this just be `//pkg:__subpackages__` instead?

I'm not opposed to this change as-is, though.

pkg/private/BUILD
```diff
@@ -70,7 +70,7 @@ py_library(
...
```
nacl, line 73: Ditto.

nacl (approved):
LGTM other than a nitpick.

There are some other zipfiles being specified inside `tests/BUILD`.  Is the intent that package-specific tests go into `tests/$PACKAGE_NAME`, and more general ones remain in `tests/` directly?

pkg/private/BUILD
aiuto: I think the second one is too broad. Unless we want the internals visible to things like doc geneation.

aiuto:
I'm not sure yet.  There is a pattern where we have some tests that we want to make sure work the same way for all packages. But perhaps the right thing to do is have them all distinct. I don't have the willpower to think hard about that tonight. I just want to get each of the packages mostly separated, so it is clear where they go.

PR #464, opened 2021-11-09
Stop using //tests/testdata/*.ar for legacy rpm smoke test input files.

aiuto:
- There was no need to use the .ar files as input.
- They are meaningful data for other tests.
- I would like to move those tests and their input to another place to properly disentangle pkg_tar tests from others. #463

Doing this before #463 makes that a little smaller.


nacl (approved):
LGTM.

We use these files in the [new `pkg_rpm` tests](https://github.com/bazelbuild/rules_pkg/blob/fa8ddc9f3ab0d5a215b5e6b893e34b216a312a36/tests/rpm/BUILD#L48-L59) too.  Should they be removed?  They seemed like good sample data at the time.

aiuto:
> LGTM.
> 
> We use these files in the [new `pkg_rpm` tests](https://github.com/bazelbuild/rules_pkg/blob/fa8ddc9f3ab0d5a215b5e6b893e34b216a312a36/tests/rpm/BUILD#L48-L59) too. Should they be removed? They seemed like good sample data at the time.

Yes. They should. I'm leaving that for another PR because my goal here is to mainly to remove the link from //pkg/... out to //tests, which would not fly with the restructured source tree.

PR #466, opened 2021-11-09
Added `//:bzl_lib` target so users can generate stardoc docs

UebelAndre:
As a consumer of `rules_pkg` I want to be able to generate docs via [stardoc](https://github.com/bazelbuild/stardoc). This requires the tree of `.bzl` source files to be consumable publicly. This change adds `@rules_pkg//:bzl_lib` which is a [bzl_library](https://github.com/bazelbuild/bazel-skylib/blob/1.1.1/bzl_library.bzl). With this publicly visible target, users are able to generate docs.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
 exports_files(
     ["WORKSPACE"],
     visibility = ["//visibility:public"],
@@ -33,4 +35,21 @@ filegroup(
         "LICENSE",
     ],
     visibility = ["//distro:__pkg__"],
+)
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
+        "//pkg:bzl_srcs",
+        "//toolchains:bzl_srcs",
+    ],
+)
+
+bzl_library(
+    name = "bzl_lib",
+    srcs = [
+        ":bzl_srcs",
+        "@bazel_skylib//lib:paths",
+    ],
+    visibility = ["//visibility:public"],
 )
--- a/toolchains/BUILD
+++ b/toolchains/BUILD
@@ -74,3 +74,11 @@ toolchain(
     toolchain = ":no_rpmbuild",
     toolchain_type = ":rpmbuild_toolchain_type",
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
+        "//toolchains/git:bzl_srcs",
+    ],
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -83,3 +83,14 @@ py_binary(
     python_version = "PY3",
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
+        "//pkg/legacy:bzl_srcs",
+        "//pkg/private:bzl_srcs",
+        "//pkg/releasing:bzl_srcs",
+        "//pkg/rpm:bzl_srcs",
+    ],
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/legacy/BUILD
+++ b/pkg/legacy/BUILD
@@ -23,3 +23,9 @@ filegroup(
     ],
     visibility = ["//pkg:__pkg__"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/private/BUILD
+++ b/pkg/private/BUILD
@@ -111,3 +111,12 @@ py_library(
     srcs_version = "PY3",
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
+        "//pkg/private/deb:bzl_srcs",
+        "//pkg/private/zip:bzl_srcs",
+    ],
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/private/deb/BUILD
+++ b/pkg/private/deb/BUILD
@@ -48,3 +48,9 @@ py_binary(
         "//pkg/private:helpers",
     ],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/private/zip/BUILD
+++ b/pkg/private/zip/BUILD
@@ -53,3 +53,9 @@ py_binary(
         "//pkg/private:manifest",
     ],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//:__subpackages__"],
+)
--- a/pkg/releasing/BUILD
+++ b/pkg/releasing/BUILD
@@ -67,3 +67,9 @@ py_binary(
     # TODO(https://github.com/bazelbuild/bazel/issues/7377): Make this private.
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//visibility:public"],
+)
--- a/pkg/rpm/BUILD
+++ b/pkg/rpm/BUILD
@@ -41,3 +41,9 @@ py_binary(
     srcs = ["augment_rpm_files_install.py"],
     visibility = ["//visibility:public"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//:__subpackages__"],
+)
--- a/toolchains/git/BUILD
+++ b/toolchains/git/BUILD
@@ -71,3 +71,9 @@ toolchain(
     toolchain = ":no_git",
     toolchain_type = ":git_toolchain_type",
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]),
+    visibility = ["//:__subpackages__"],
+)


distro/BUILD
```diff
@@ -108,16 +108,7 @@ genrule(
...
-        "//pkg:mappings.bzl",
-        "//pkg:package_variables.bzl",
-        "//pkg:path.bzl",
-        "//pkg:pkg.bzl",
-        "//pkg:providers.bzl",
-        "//pkg:rpm.bzl",
-        "//pkg/private:pkg_files.bzl",
-        "//pkg/private:util.bzl",
-        "//pkg/private/deb:deb.bzl",
+        "//:bzl_srcs",
```
aiuto, line 111: Why not just make a filegroup of the files listed originally here and reuse that.

I don't particularly like the public export of the .bzl file everywhere.  That is going to conflict with trying to lock out the private stuff when bzl file visibility becomes available soon (we hope).

BUILD
```diff
@@ -34,3 +34,11 @@ filegroup(
     ],
     visibility = ["//distro:__pkg__"],
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
```
aiuto, line 40: If we go with this,  these should all
- not use glob
- restrict visibility to the places that can consume it. That might be broader for pkg.bzl, but certainly locked down for distro.

Can you file issues for this kind of thing first so we can discuss the right solutions first.

distro/BUILD
UebelAndre: `.bzl` files in `private` packages will need to be included in this filegroup. It doesn't matter where the `.bzl` file is, if a top level file loads a anything from it, it will be required by `stardoc`. This doesn't alter anything about the `bzl-visibility` warnings, which is why I opened https://github.com/bazelbuild/rules_pkg/pull/467 originally, to try and catch visibility violations, but unfortunately, because loads can still occur in any package, it's required that all `.bzl` source files are included in the globl

BUILD
UebelAndre: The use of `glob` pairs well with the release pattern here of creating a `.tar` file with select files. It's intentional that these targets broadly include everything in the package to allow the `package_contents` targets to remain the source of truth.

Based on my experience, there aren't many other solutions. This pattern of filegroups is used in `rules_python`, `rules_docker, `rules_nodejs`, `rules_cc`, `rules_rust` and surely others. If the glob pattern is a large issue for you, that can be changed but I'd warn that it's highly vulnerable to breakages with explicit packages. If a single file is missing, the ability to generate docs is broken and would require a PR and a new release.

UebelAndre:
@aiuto I do not have bandwidth to participate in a large discussion about this. I think this is a very common pattern and didn't think it'd break anything so wanted to try and open the PR for the next release. If you're against this change then that makes me sad but will respect your decision.

UebelAndre:
@aiuto https://docs.bazel.build/versions/4.2.1/visibility.html#visibility-of-bzl-files is the unfortunate reality which in my experience makes the pattern in this PR the most effective and safest from regressions.

achew22:
Another option here would be to run [Gazelle](https://github.com/bazelbuild/bazel-gazelle) with the [Skylib](https://github.com/bazelbuild/bazel-skylib/tree/main/gazelle/bzl) plugin installed and generate `bzl_library` targets. If this is something the project would be interested in having, I'd be happy to send a PR that sets up Gazelle and a separate one that generates the `bzl_library` targets per that generator.

What do you think?

UebelAndre:
If I may, I personally don't care what the implementation is. I'd like to generate docs and my only reservation with Gazelle is I don't want to be forced to introduce it as a dependency in my workspace. If both points are satisfied I'm a happy camper

nacl:
I could use some elaboration on the use case here.

We have targets to generate (admittedly only some) of the stardoc documentation in `pkg/distro/BUILD`.  Users can generate documentation by invoking those targets, and perhaps those targets should be expanded or modified to produce more of the documentation.  Will those not work for you?

UebelAndre:
This is fixing a regression since release `0.2.1` which is the version I'm currently using. After upgrading, I'm no longer able to generate docs as stardoc does not have access to some transitive `.bzl` files in the `private` package. The source file is not a public target so I cannot add it to my `bzl_library`.

UebelAndre:
@nacl @aiuto friendly ping here. Any other questions I can answer to help advance a decision?

nacl:
So, we have some `stardoc` targets in the distro/ directory:

https://github.com/bazelbuild/rules_pkg/blob/ce8c20ac803fecdb99111a6d7b579e5f51146d1a/distro/BUILD#L126-L173

I was thinking you could use them (hence my question), but then I realized they weren't exported in the release tarball.  Is that part of your concern here?

Regardless, perhaps we should export them and/or move them somewhere else?  Would this satisfy your requirements?  @aiuto, WDYT?

UebelAndre:
> So, we have some `stardoc` targets in the distro/ directory:
> 
> https://github.com/bazelbuild/rules_pkg/blob/ce8c20ac803fecdb99111a6d7b579e5f51146d1a/distro/BUILD#L126-L173
> 
> I was thinking you could use them (hence my question), but then I realized they weren't exported in the release tarball. Is that part of your concern here?
> 
> Regardless, perhaps we should export them and/or move them somewhere else? Would this satisfy your requirements? @aiuto, WDYT?

Exporting stardoc targets doesn't help. Those are docs you (`rules_pkg`) are generating. If anything the `bzl_library` might be useful but only if it includes any source files users are able to load without `bzl-visibility` violations.

A good test of what I want would be to take all the public (`bzl-visibility` compliant) files and define a stardoc target to consume them. I should be able to load any of these sources in my modules and continue to use stardoc.
```
rules_pkg-0.5.1
├── deps.bzl
├── install.bzl
├── legacy
│   └── rpm.bzl
├── mappings.bzl
├── package_variables.bzl
├── path.bzl
├── pkg.bzl
├── providers.bzl
├── releasing
│   ├── defs.bzl
│   └── git.bzl
├── rpm.bzl
├── rpm_pfg.bzl
├── toolchains
│   ├── rpmbuild.bzl
│   └── rpmbuild_configure.bzl
└── version.bzl
```

As of today, that's not possible but fixing that in any way would satisfy my needs.

achew22:
> Exporting stardoc targets doesn't help. Those are docs you (rules_pkg) are generating. If anything the `bzl_library` might be useful but only if it includes any source files users are able to load without bzl-visibility violations.

That is exactly what `bzl_library` allows doing. It will provide to stardoc all the internal files for stardoc generation purposes without exposing the actual file contents.

UebelAndre:
> That is exactly what `bzl_library` allows doing. It will provide to stardoc all the internal files for stardoc generation purposes without exposing the actual file contents.

Right, I don't think it makes a significant difference if the exposed target is a `bzl_library` or `filegroup`. They both need to include all sources. Maybe I'm misinterpreting something but it sounds like the issue is about having multiple globs of filegroups (publicly visible or otherwise) and that explicit lists of sources are used instead. This approach is fine but only if you have integration tests to ensure your `bzl_library`/`filegroup` works for generating any public `.bzl` file. Otherwise you'll run into regressions.

BUILD
aiuto: Sure, but the ability to generate docs is not a release breaker.
The doc generation code is not part of the released tarball.

The bigger problem is that I do not want to make a commitment that the stardoc rules will always work, or be useful.  I have a PR out #474 that starts to move the process towards post-processing the stardoc to generate a unified reference page.
The intention is that we'll add code in the future to generate inter rule links, turn markers into version specific stamps and several other things that we've been thinking about in the overall bazel docs rewrite.

Let's say you could point at @rules_pkg//pkg:bzl_files and get a set of every .bzl file.  I am not about to make a commitment that they can be consumed for any other purpose than using them in rules.  You might get stardoc to produce something, but that won't necessarily be good docs.

Here's an example we already have which shows the problem:  All the common attributes are documented once. I won't accept PRs that expand the text into the rule declarations in the .bzl files..  But, we could go with a path that post processes stardoc output and expands the boilerplate text there.  Either way, the raw stardoc generated text is insufficient.  The promise to users that I want to make is that we delivere high quality docs alongside the distribution. I won't promise that our source tree is arranged in such a way that people can bring their own tools to produce docs.

aiuto:
> @aiuto I do not have bandwidth to participate in a large discussion about this. ...

Discussion is the way we have made all the decisions that impact the way we work. @nacl and I regularly debate things in PRs, and discuss style changes and refactorings in the monthly engineering meeting before making them.  The recent PRs to restructure the code base were discussed before we began them, specifically so that the PR reviews would be tactical rather than strategic.

BUILD
UebelAndre: I don't think we're on the same page. I have rules/macros that use `rules_pkg`. I want to generate documentation using `stardoc` so employees of `${WORKPLACE}` have some documentation they can read that explains how to use the rules/macros without having to read the code itself. This works fine with `rules_pkg-0.2.1` but in updating to `rules_pkg-0.5.1`, I can no longer generate my own docs because of a change to some `.bzl` file that includes a private `.bzl` file. For me, it would be ideal to not need to build a `bzl_library` of every individual file `rules_pkg` loads transitively. Instead, I want a single target that has everything I need. That is what this pull-request provides in the form of a `filegroup`. If the `bzl_library` in [@rules_pkg//distro](https://github.com/bazelbuild/rules_pkg/blob/ce8c20ac803fecdb99111a6d7b579e5f51146d1a/distro/BUILD#L108-L124) were made public and contained all the `*.bzl` files, I'd be happy to use that.

UebelAndre:
> > @aiuto I do not have bandwidth to participate in a large discussion about this. ...
> 
> Discussion is the way we have made all the decisions that impact the way we work. @nacl and I regularly debate things in PRs, and discuss style changes and refactorings in the monthly engineering meeting before making them. The recent PRs to restructure the code base were discussed before we began them, specifically so that the PR reviews would be tactical rather than strategic.

I totally get and respect that. I thought this pull request would be a pretty easy and straight-forward change since it's such a common pattern among rules in the `bazelbuild` org so I thought I'd help out by submitting the change. Ultimately, I just want to be able to generate my own docs again.

nacl:
Given my understanding of the above, I'd feel more comfortable with exposing our sources as `bzl_library` targets instead of `filegroup`s.

UebelAndre:
> Given my understanding of the above, I'd feel more comfortable with exposing our sources as `bzl_library` targets instead of `filegroup`s.

I'd be happy with that. Again, just wanna make sure everyone understands https://docs.bazel.build/versions/4.2.1/visibility.html#visibility-of-bzl-files. The existence of this target has no impact on what users can or cannot load in their projects. This *just* fixes workflows involving stardoc.

UebelAndre:
If there's an agreement to use `bzl_library`, I'm happy to update the filegroups to be that or whatever the change should be. A bullet list of the action items would help a ton.

UebelAndre:
Friendly ping on this PR 😅

nacl:
Bah.  I forgot to mention this in our meeting today.  @aiuto, thoughts on exposing `bzl_library` targets to allow for this use case?  It sounds good to me.

nacl:
@aiuto friendly ping on this.  Any opinions on proceeding with `bzl_library`?

BUILD
aiuto: I get the desire to generate your own docs.
On my side, I don't want to be hamstrung by an expectation that we'll keep the docs stardoc compliant. I've recently moved doc generation out of distro and into doc_build, where I can build a unified docs page to publish.  See https://bazelbuild.github.io/rules_pkg/0.5.0/reference.html for an example.

I am only using stardoc today because that is what is available.  What I would like to be able to do is add annotations that we can expand in post processing the stardoc output. What I can do is make doc_build available (which is equivalent to your alternative of making distro available).  I just have to figure out the implications of not forcing users to take a stardoc dependency. I could just say don't do  `bazel test @rules_pkg//...`, or maybe I can make it optional in the deps.

UebelAndre: I think there's a massive amount of confusion here and I don't know how to resolve it. I do not think these rules should have any reliance on `stardoc`. Users should be able to `bazel build @rules_pkg//...` from their repo and never be hit with `@io_bazel_stardoc is missing`. The issue is, I use stardoc for my own docs and after version ~2 I'm not able to generate docs anymore because of some change that added a `*.bzl` file that is no longer visible to *my* stardoc target. All I'm asking for is a filegroup (or `bzl_library`) that has a all `*.bzl` files in the distro package. Which I think is a non-invasive change that puts almost no burden on maintainers (only burden being if a new package is added then the relevant target be made and wired up with the top level `*.bzl` glob).

My situation is I have some wrapper rules around `rules_pkg` and want to generate documentation for them. I cannot because certain `*.bzl` files are not visible to me. What can I do to fix this?

UebelAndre:
@nacl @aiuto I've updated the PR to only expose a `bzl_library` target. 

Also wanna say thanks the time you've already taken to review this and sorry that it's been such a confusing and frustrating experience.

BUILD
aiuto: I'm not a fan of the bzl_srcs recursive thing, but it is the path of least resistance for getting  all the .bzl file in one place.
I can take this PR, but I don't want to switch doc_build over to that, so I will revert that part and try to merge.

toolchains/BUILD
```diff
@@ -74,3 +74,11 @@ toolchain(
     toolchain = ":no_rpmbuild",
     toolchain_type = ":rpmbuild_toolchain_type",
 )
+
+filegroup(
+    name = "bzl_srcs",
+    srcs = glob(["**/*.bzl"]) + [
+        "//toolchains/git:bzl_srcs",
```
aiuto, line 81: Just a note that the git toolchain is undocumented by design, because it is too finicky to use.

UebelAndre: The `bzl_srcs` and `bzl_lib` targets do not impact what does and does not get documented. These targets are only intended to produce a tree of `.bzl` files. In order to document this file, the specific file itself would need to be exposed as a target so it could be passed to something like `stardoc`. The recursive glob adds a convenient way to track `.bzl` srcs without requiring users to maintain explicit lists (which are very often forgotten about).

BUILD
aiuto: Actually, I am still looking for somehting else
Can we just make the sources available at the top level.
Ideally, I would drop an dependency on bzl_skylib, Aside from paths (which can be eliminated), the only reason for it is for bzl_library.

UebelAndre: That's fine. Though I don't see a reason why you wouldn't use the top level `bzl_library`. `stardoc` has no problem with `bzl_library` targets including more sources that what's used. I've found this to be easier to manage but I may misunderstand the intent of `bzl_library`. Happy learn more about it though

UebelAndre:
> I can take this PR, but I don't want to switch doc_build over to that, so I will revert that part and try to merge.

@aiuto I believe you have write permissions to this PR but let me know if you need me to click anything

UebelAndre:
@aiuto @nacl friendly ping 😅 is this ready to merge?

UebelAndre:
@aiuto @nacl one more ping 😅 it would be nice to close this out

UebelAndre:
@aiuto @nacl another ping. Could this be merged to finally close this out? Is there any more work needed on this?

UebelAndre:
@aiuto @nacl sorry to keep pinging. But I noticed some `0.6.0` related changes and would very much like to have this change in the upcoming release. Is there something I can do to help get this merged? 🙏

UebelAndre:
Kinda sad to see the new release without this. Should I just close the PR?

aiuto:
Sorry to necropost.  Functional changes were higher on my priority than
structural ones.
I think I have something I like that avoids a duplicate, parallel 'include
my subdirs' tree.
I have a change ready for PR which does something like.

bzl_sources(
   name = 'bzl",
   srcs = "//:standard_package",
)

That takes all the files which go in the distribution and filters them to
just the .bzl ones.
That becomes suitable for a bzl_library to inject into stardoc.  When I get
back from
vacation I can send it around.

Related, for those who have carefully read PR descriptions, I have taken
the stand that the
way to improve stardoc is to build post-processors on top of it to
implement the features
I want and file bugs against Stardoc.  For example, I have a PR outstanding
that fixes
the output to work around problems with <center> tags and lists in
attributes.
IMO, raw stardoc output is not usable yet, so even having bzl_srcs
available doesn't
help people as much as the good docs at
https://bazelbuild.github.io/rules_pkg/

PR #469, opened 2021-11-09
Fix use of name parameter to pkg_tar

kwc:
The legacy branch of pkg_tar attempts to fetch the name parameter using kwargs. As name is an explicit parameter, this results in a build error. The PR changes this branch to use 'name' instead.


kwc:
@googlebot I signed it!

aiuto (approved):
Thanks for this fix. I'm surprised no one hit this problem before.

PR #478, opened 2021-11-25
compute_data_path should add the workspace_root

BoleynSu:
pkg_tar rule will produce a tarbal with files
named exteranl/workspace_name/path_to_file
even if `strip_prefix = "."` currently. This commit
fixes it.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -65,3 +65,12 @@ local_repository(
     name = "mappings_test_external_repo",
     path = "tests/mappings/external_repo",
 )
+
+# This repository is used to test pkg_tar rule itself.
+local_repository(
+    name = "rules_pkg_test_workspace",
+    path = ".",
+    repo_mapping = {
+        "@rules_pkg": "@",
+    },
+)
--- a/pkg/path.bzl
+++ b/pkg/path.bzl
@@ -68,7 +68,10 @@ def compute_data_path(ctx, data_path):
       ctx: rule implementation ctx.
       data_path: path to a file, relative to the package of the rule ctx.
     """
-    build_dir = ctx.label.package
+    build_dir = ctx.label.workspace_root
+    if ctx.label.workspace_root and ctx.label.package:
+        build_dir += "/"
+    build_dir += ctx.label.package
     if data_path:
         # Strip ./ from the beginning if specified.
         # There is no way to handle .// correctly (no function that would make
--- a/tests/BUILD
+++ b/tests/BUILD
@@ -420,6 +420,15 @@ pkg_tar(
 )
 
 pkg_tar(
+    name = "test-tar-strip_prefix-external_workspace",
+    srcs = [
+         "BUILD",
+    ],
+    strip_prefix = ".",
+    visibility = ["//visibility:public"],
+)
+
+pkg_tar(
     name = "test-tar-files_dict",
     files = {
         ":etc/nsswitch.conf": "not-etc/mapped-filename.conf",
@@ -493,6 +502,7 @@ py_test(
         ":test-tar-strip_prefix-etc.tar",
         ":test-tar-strip_prefix-none.tar",
         ":test-tar-strip_prefix-substring.tar",
+        "@rules_pkg_test_workspace//tests:test-tar-strip_prefix-external_workspace",
         ":test_tar_package_dir_substitution.tar",
         ":test-tar-long-filename",
         ":test-tar-repackaging-long-filename.tar",
--- a/tests/pkg_tar_test.py
+++ b/tests/pkg_tar_test.py
@@ -39,7 +39,10 @@ class PkgTarTest(unittest.TestCase):
     """
     # NOTE: This is portable to Windows. os.path.join('rules_pkg', 'tests',
     # filename) is not.
-    file_path = runfiles.Create().Rlocation('rules_pkg/tests/' + file_name)
+    if file_name[0] == '/':
+      file_path = runfiles.Create().Rlocation(file_name[1:])
+    else:
+      file_path = runfiles.Create().Rlocation('rules_pkg/tests/' + file_name)
     with tarfile.open(file_path, 'r:*') as f:
       i = 0
       for info in f:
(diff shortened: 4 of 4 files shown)


BoleynSu:
@aiuto @nacl PTAL

aiuto (changes requested):
Please describe the problem behavior in an issue first, with before and after examples.
From what it looks like, 
Als, any behavior changing PR must include tests.

BoleynSu:
@aiuto I will open an issue and add more tests for the PR later. Will ping you then.

BoleynSu:
Fixes #479 
Hi @aiuto, I have added the issue and the test. PTAL

nacl:
Thanks for the PR!

Without looking at this review in any particular detail, do you get the desired results (removal of the workspace root) if you pass a `pkg_files` target like this to `pkg_tar` (see `pkg/mappings.bzl`):

```starlark
# Needed new loads
load("@rules_pkg//pkg:mappings.bzl", "pkg_files", "strip_prefix")

pkg_files(
    name = "your-files",
    srcs = ["your", "files", "here"],
    strip_prefix = strip_prefix.from_pkg(""),
)
```

This is the direction we're leaning towards in the long-term: `pkg_files` is intended to be a package-type-independent means for mapping bazel targets into packaging outputs, and is intended to replace all path remapping functionality in packaging rules (like `strip_prefix` in `pkg_tar`).

That being said, I'm not sure about the long-term fate of `pkg/path.bzl` -- most of the functionality needed for `pkg_files` is already included in `pkg/mappings.bzl`.  @aiuto may have a more concrete opinion on this than I.

BoleynSu:
> Thanks for the PR!
> 
> Without looking at this review in any particular detail, do you get the desired results (removal of the workspace root) if you pass a `pkg_files` target like this to `pkg_tar` (see `pkg/mappings.bzl`):
> 
> ```python
> # Needed new loads
> load("@rules_pkg//pkg:mappings.bzl", "pkg_files", "strip_prefix")
> 
> pkg_files(
>     name = "your-files",
>     srcs = ["your", "files", "here"],
>     strip_prefix = strip_prefix.from_pkg(""),
> )
> ```
> 
> This is the direction we're leaning towards in the long-term: `pkg_files` is intended to be a package-type-independent means for mapping bazel targets into packaging outputs, and is intended to replace all path remapping functionality in packaging rules (like `strip_prefix` in `pkg_tar`).
> 
> That being said, I'm not sure about the long-term fate of `pkg/path.bzl` -- most of the functionality needed for `pkg_files` is already included in `pkg/mappings.bzl`. @aiuto may have a more concrete opinion on this than I.

I verified that pkg_files also has the same issue, i.e. the following won't pass the test.

```python
pkg_files(
    name = "your-files",
    srcs = ["BUILD"],
    strip_prefix = strip_prefix.from_pkg(""),
)
pkg_tar(
    name = "test-tar-strip_prefix-external_workspace",
    srcs = [
         "your-files",
         #"BUILD",
    ],
    strip_prefix = ".",
)
```

nacl:
I see.  I checked out your change locally.  When using `pkg_files` like I described above, I found that your new test passed.  Code is available in [topic/i479-compute_data_path](https://github.com/nacl/rules_pkg/tree/topic/i479-compute_data_path) in my personal fork.

Please take a look; I could be missing something.

BoleynSu:
Hi @nacl, do you mean the pkg_tat failed but pkg_files passed? Would you
mind also pushing yout changes in tests/BUILD to your repo?

BoleynSu:
Hi @nacl, I just saw your commit. The only difference is that in my local
test I kept the strip_prefix attr in pkg_tar. Would you mind try that?

nacl:
I readded `strip_prefix` to `pkg_tar` as you had it above, and your test still passed

`strip_prefix` on `pkg_tar` should only interact with file inputs.  It should have no interaction with inputs from `pkg_files` and similar rules.

BoleynSu:
> I readded `strip_prefix` to `pkg_tar` as you had it above, and your test still passed in my environment.
>
> What's supposed to happen is that `pkg_tar`'s `strip_prefix` attribute should only interact with file inputs. It should not interact with `pkg_files` and similar rules.


I have uploaded my working example at https://github.com/BoleynSu/rules_pkg/tree/demo. FYI, I am using Bazel 4.2.1.

```text
Build label: 4.2.1
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Mon Aug 30 15:17:47 2021 (1630336667)
Build timestamp: 1630336667
Build timestamp as int: 1630336667
```

Please take a look.

BoleynSu:
I just realize I am doing it wrongly. I should also use an external pkg_tar when I am using an external pkg_files. Otherwise, the provider won't match.

So, the conclusion is that pkg_files works as expected.

@nacl should I close the issue as WONT FIX or we still wants to fix pkg_tar?

nacl:
> I just realized I was doing it wrongly. I should also use an external pkg_tar when I am using an external pkg_files. Otherwise, the provider won't match.

Huh.  That's odd.  It's weird that this is the case -- IIRC providers are just structures in the global namespace. and where they are declared shouldn't really matter, but I could easily be mistaken.  @aiuto, do you know what might be happening here?

Regardless, I don't think this is a problem in rules_pkg (if it is a problem at all).  Thank you for double-checking.

> So, the conclusion is that pkg_files works as expected.

Great to hear.  I'll close this out.

For additional context, I'd be far more willing to accept this if it weren't a breaking change.  Even though this behavior is bad, changing this would impact anything that uses `path.bzl` to create paths, including rules like `pkg_zip`, and would require users to migrate.  It's not a big deal, but we'd rather people just migrate directly to using `pkg_files` and friends.

If you need something like this in tar files and the `pkg_filegroup` framework (`pkg_files`, etc.), I think `remap_paths` will do the right thing.

FWIW, this functionality for `pkg_files` has [tests](https://github.com/bazelbuild/rules_pkg/blob/3c520a1becd011aa3c578a5edef27c180936adea/tests/mappings/mappings_external_repo_test.bzl).  I'd be happy to accept additions to the test framework if you have any ideas.

BoleynSu:
For the provider issue, it is kinda like

a.py:

import b
assert(x!=b.x)

b.py: symbolic link of a.py

Would you mind documenting the deprecation of strip_prefix functionality of
pkg_tar/pkg_whatever? So that other people could avoid my situation in the
first place.

aiuto:
The location of the bzl file declaring a provider matters. The package
becomes the namespace for it to prevent symbol clashes.

BoleynSu:
Just in case you missed it, @aiuto, the issue still exists for pkg_tar but
not pkg_files. However, the fix results in a breaking change ( since the
old behavior is buggy). Do you think we should fix it? It should only
affect those who used it in a buggy way.

PR #484, opened 2021-12-10
Add example of how to depend one pkg_deb .changes file

### LINKED ISSUE #477: Unable to use changes output file from `pkg_deb` in further rules
Previously i was able to tar deb and changes files created by `pkg_deb` using `pkg_tar`

I did this by manually listing expected outputs as srcs for the pkg_tar rule, eg:

```

debs = (
     "envoy.deb"
)

changes = (
    "envoy_%s.changes" % version,
)

pkg_tar(
       name = "base_debs",
       extension = "tar",
       package_dir = "deb",
       srcs = [":%s" % f for f in debs + changes],
 )
 
```

this no longer seems to work

if i build the deb/s correctly i can see the changes files, but they are no longer available to subsequent rules

aiuto:
- Improve the example about finding output file paths to show how to
  get the implicit outputs like the changes file.

Fixes #477

cc/ @phlax


phlax:
thanks for this @aiuto i have tested it and, as suggested, using `filegroup` to pull out the debs and changes files, it works a treat

nacl:
I will find time to review this today or Friday.

aiuto:
@phlax Was the example clear enough?

phlax:
> Was the example clear enough?

yep. also, happy to relate use...

first, i tried to just use the `pkg_deb` rule i have as a `src` for a `pkg_tar` rule

this gave me the files i wanted, but it also added unwanted `rulename.deb.deb` files

so i then added a filegroup with for each `output_group` of `deb` and `changes` which gave me the specific files i wanted when added to my `pkg_tar`

the only mistake i initially made was in thinking that it required a `pkg_filegroup` rather than a `filegroup` but not sure if that actually needs to be disambiguated, just my confusion

hope this helps

tests/deb/deb_tests.bzl
```diff
@@ -15,6 +15,25 @@
 
 load("@bazel_skylib//lib:unittest.bzl", "analysistest", "asserts")
 
+def assert_contains(env, expected, got, msg = None):
+    """Asserts that the given `expected` occurs in a set of things.
```
nacl, line 19: Nit: "set" could refer to skylib [sets](https://github.com/bazelbuild/bazel-skylib/blob/main/docs/new_sets_doc.md), which are `struct`s.

Maybe say:

```markdown
Asserts that the given `expected` occurs in iterable (`list`, `dict`) `got`.
```

tests/deb/deb_tests.bzl
```diff
@@ -37,16 +56,58 @@ def _package_naming_test_impl(ctx):
...
 
-
 package_naming_test = analysistest.make(
     _package_naming_test_impl,
     attrs = {
         "expected_name": attr.string(),
     },
 )
+
+def _all_files_in_default_info_impl(ctx):
```
nacl, line 72: Another way to implement this would be to convert everything to use skylib sets, and assert that that everything you want [is a subset of what's expected](https://github.com/bazelbuild/bazel-skylib/blob/1.0.3/docs/new_sets_doc.md#setsis_subset).

What you have here may also be easier to follow.  It's fine either way.

tests/deb/deb_tests.bzl
```diff
@@ -15,6 +15,25 @@
...
+          If omitted, a default will be used.
+    """
+    if expected in got:
+        return
+    expectation_msg = "Expected %s in (%s)" % (expected, got)
+    if msg:
+        full_msg = "%s (%s)" % (msg, expectation_msg)
+    else:
+        full_msg = expectation_msg
+    fail(env, full_msg)
```
nacl, line 35: How does all this look using something like [asserts.true](https://github.com/bazelbuild/bazel-skylib/blob/1.0.3/docs/unittest_doc.md#assertstrue)?

If that doesn't look good, then perhaps this at the bottom should be `unittest.fail()` or `analysistest.fail()` instead?  Documentation in [here](https://github.com/bazelbuild/bazel-skylib/blob/1.0.3/docs/unittest_doc.md).

nacl (commented):
The documentation updates are very useful, but I don't think we've ever quite come to a consensus on what should be in `DefaultInfo`.  See the large comment for a starter discussion. on that.

I think what you have specified here is the right thing to do, but I am not completely sold on it.  The code looks good modulo the below small suggestions in the test code.

examples/where_is_my_output/README.md
```diff
@@ -64,3 +64,34 @@ def format(target):
...
+from pkg_deb is an example. If we want another rule to depend on an
+implicitly created file, we can do that with a filegroup that
+specifies the specific output group containing that file.
+
+In the example below, `:deb` is a rule producing an explicit .deb output
+and an implict .changes output. We refer to the .changes file using
+the `filegroup` and specifying the desired output group name. Then,
+any rule can use this `filegroup` as an input.
+
+```python
```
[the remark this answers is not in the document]
aiuto: I would rather go with the more commonly known (in the markdown world) one.

tests/deb/deb_tests.bzl
aiuto: done

tests/deb/deb_tests.bzl
aiuto: It seems a waste to use is_subset for a singleton item.

tests/deb/deb_tests.bzl
aiuto: swtiched to analysistest.fail

examples/where_is_my_output/BUILD
```diff
@@ -39,3 +39,18 @@ pkg_deb(
...
     version = "3.14",
 )
+
+# We can also depend just on the .changes file
+
+filegroup(
+    name = "the_changes_file",
+    srcs = [":deb"],
+    output_group = "changes",
+)
```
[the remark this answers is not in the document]
aiuto: ack

examples/where_is_my_output/README.md
```diff
@@ -64,3 +64,34 @@ def format(target):
...
 requires some knowledge of how to write custom Bazel rules. See the Bazel
 documentation for more information.
+
+## Using an implicit output as input to another rule.
+
+Sometimes a rule will create an implicit output that the user does not
+explicitly specify as an attribute of the target. The .changes file
+from pkg_deb is an example. If we want another rule to depend on an
+implicitly created file, we can do that with a filegroup that
+specifies the specific output group containing that file.
```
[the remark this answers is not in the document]
aiuto: I think DefaultInfo has to include both the real .deb file and the symlink to it. That will allow the common use case of 
- depend on a pkg_deb(name=foo, package=bar, pkg_file_name='something with expansions')
- have a script that blindly opens the (symlink) file bar.deb
I think that is the least surprising one for lazy users who just want a script around the build.

So, the question to me is if we want the changes file as a default output or not?
Do people consider the changes file an important artifact or not?
Is there a comparable extra output from RPM building?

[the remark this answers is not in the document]
phlax: i think changes files are important as they are required for signing, to my knowledge there is not a comparable artefact for rpms, and signing happens on the binary itself

whether that means it should be part of the default output im not sure - as long as its possible to get it from the `pkg_deb` rule one way or another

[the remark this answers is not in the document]
aiuto: nacl@ and I discussed it at the eng meeting. We'll drop it from DefaultInfo, so that the expectation of a single output file remains the same.

aiuto:
@nacl If you could do a quick review again, I would like to get this in for the example, as it appears relevant to https://github.com/bazelbuild/bazel/issues/7977

examples/where_is_my_output/README.md
```diff
@@ -64,3 +64,34 @@ def format(target):
 A full explanation of why this works is beyond the scope of this example. It
 requires some knowledge of how to write custom Bazel rules. See the Bazel
 documentation for more information.
+
+## Using an implicit output as input to another rule.
+
+Sometimes a rule will create an implicit output that the user does not
+explicitly specify as an attribute of the target. The .changes file
+from pkg_deb is an example. If we want another rule to depend on an
```
nacl, line 72: Nit: markdownify?

nacl (approved):
LGTM.  Please update/reword the commit message to indicate what actually is happening here -- just a documentation/examples update.

examples/where_is_my_output/README.md
aiuto: I reflowed it.     `!}fmt -80`

PR #485, opened 2021-12-13
Add MODULE.bazel file

nacl:
[bzlmod] is the upcoming package manager for Bazel's external dependencies.

While it's already in the [central registry], it's apparently also helpful to
have such a file in the repository, in case, you need to use overrides.

This change adds such a file, specifically the one from the central repository
with the version bumped.

I can't quite get the tests to work yet; I think there's a bug in how runfiles
trees are created (the rules_python directory for the runfiles lib contains the
version in addition to the name).  This was tested using the latest 5.0 RC; I
currently cannot test at bazel HEAD due to https://github.com/bazelbuild/bazel/issues/14403.

[bzlmod]: https://docs.bazel.build/versions/main/bzlmod.html
[central registry]: https://github.com/bazelbuild/bazel-central-registry

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,22 @@
+# Copyright 2021 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+module(
+    name = "rules_pkg",
+    version = "0.6.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_python", version = "0.4.0")
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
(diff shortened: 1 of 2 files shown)


MODULE.bazel
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+module(
+    name = "rules_pkg",
+    version = "0.6.0",
+    compatibility_level = 1,
+)
+
+bazel_dep(name = "rules_python", version = "0.4.0")
+bazel_dep(name = "bazel_skylib", version = "1.0.3")
```
aiuto, line 22: Can we make this float to head, or whatever whomever depends on rules_pkg wants to use?

aiuto:
Any update?

nacl:
BTW these problems might not affect users of rules_pkg, but they definitely prevent us from using it ourselves.

aiuto:
Can we check in the module definition but not use it in our CI? I don't see any need to use it ourselves, but we can give others a leg up on trying it. Perhaps they will shake out any issues in the declaration.

nacl:
> Can we check in the module definition but not use it in our CI? I don't see any need to use it ourselves, but we can give others a leg up on trying it. Perhaps they will shake out any issues in the declaration.

Absolutely.  I'll see about doing this when I have an opportunity.

MODULE.bazel
aiuto: Are we able to say we depend on bazel 5.x?

aiuto:
I've taken this up in #625. 
bzlmod is not quite there yet for running our own tests, but  that PR does show a use case where a user of rules_pkg can depend on it through the module.

aiuto:
Closing this since I got an initial MODULE set up. There is more to learn, but we can work from that.

PR #487, opened 2021-12-14
pkg_rpm: Don't have source_date_epoch apply by default; test modularity cleanup

### LINKED ISSUE #486: `pkg_rpm` produces RPMs with clamped timestamps even when not requested
We've been doing work to port our old copies of `pkg_rpm` to use the rules from upstream, an internal user noticed that the RPMs were installing files with "0" mtimes.

This can wreak havoc with some update processes.  Root cause looks to be the conditional that configures SOURCE_DATE_EPOCH behavior in `make_rpm.py` seeing a string `"0"` (which is truthy) instead of integer `0` (which is falsey).  This was introduced in 06a2a231761b707bd04e89bd15f42bf181d1c3ae.

nacl:
The logic for `pkg_rpm`'s `source_date_epoch` attribute should apply only when
set.  This change updates the logic so that a negative value for
`source_date_epoch` results in no `--source_date_epoch` being passed to
`make_rpm.py`.  `None` does not seem to be a valid option.

The default is now a negative number, thus disabling this feature unless
explicitly requested.

A regression test was also provided.  While writing it, I noticed a few quirks
in `rpm` that impacted test execution.  These were also addressed, although
perhaps not in the best way.

For more context, see #486.

NOTE: this does not address the similar issue impacting the legacy `pkg_rpm`
rule.  This is mentioned in #488.

Fixes #486

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -373,7 +373,8 @@ class RpmBuilder(object):
         'LANG': 'C',
         'RPM_BUILD_ROOT': buildroot,
     }
-    if self.source_date_epoch:
+
+    if self.source_date_epoch is not None:
       env['SOURCE_DATE_EPOCH'] = self.source_date_epoch
       args += ["--define", "clamp_mtime_to_source_date_epoch Y"]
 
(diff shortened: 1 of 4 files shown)


aiuto (approved):

nacl:
Thanks for the review!

PR #493, opened 2021-12-20
Make JSON package manifests more human-readable and generic

### LINKED ISSUE #385: Package JSON manifest should support generic attributes

nacl:
NOTE: this is not ready for review.  It is mostly being presented here as a (very clumsy) PoC.  It needs a whole lot of polish, especially on the starlark side.

It's mostly being posted for curiosity, and to make sure there aren't any CI-related surprises later on.

--

This is a change that demonstrates (most of) what's needed to transition to a more verbose and extensible JSON manifest type like the one specified in https://github.com/bazelbuild/rules_pkg/issues/385#issuecomment-974709601.  The rule-facing Starlark APIs have changed little.  I'll probably make the tracing capability more like what's currently in `pkg_rpm` when `pkg_rpm` transitions to this functionality in a follow-on change.

Fixes #385.


aiuto:
Nice. I won't be able to take a good look until January.

pkg/private/build_tar.py
```diff
@@ -278,34 +277,32 @@ def add_tree(self, tree_top, destpath, mode=None, ids=None, names=None):
...
-    if entry.entry_type == manifest.ENTRY_IS_LINK:
-      self.add_link(entry.dest, entry.src, **attrs)
-    elif entry.entry_type == manifest.ENTRY_IS_DIR:
-      self.add_empty_dir(entry.dest, **attrs)
-    elif entry.entry_type == manifest.ENTRY_IS_TREE:
-      self.add_tree(entry.src, entry.dest, **attrs)
-    elif entry.entry_type == manifest.ENTRY_IS_EMPTY_FILE:
-      self.add_empty_file(entry.dest, **attrs)
+        attrs['names'] = (entry["user"], attrs.get('names')[1])
+    if entry["type"] == "symlink":
```
aiuto, line 296: This may blow out memory usage compared to an int or a reused string.

aiuto: THe more I think about it, the more I think we should not switch to a dict.
named tuples are more efficient and immutable.

nacl: Be that as it may, how much of a difference will using the named tuple really make?  The performance difference will be nice, but I doubt it will make any noticeable difference compared to the actual I/O cost of building the package. Immutability is also a nice property, but there's nothing stopping scripts from duplicating the data structure and modifying it as needed.

The main benefit of using a dict is that all attributes are in the same bucket -- there's no need to distinguish between what's official and what's an extension.  It's also simpler to maintain.

The Bazel memory use concerns are a better reason.  I tried taking measurements of that recently, and I was not able to get any useful information: results were all over the place and were not sufficiently consistent between runs.  That could be due to the fact that I was working in WSL which may have compatibility issues.

aiuto: I forgot that piece of code just unpacks, consumes and then throws it away. So there really is no memory concern. I guess this is mostly fine.

pkg/private/build_tar.py
```diff
@@ -278,34 +277,32 @@ def add_tree(self, tree_top, destpath, mode=None, ids=None, names=None):
...
+      self.add_link(entry["dest"], entry["src"], **attrs)
+    elif entry["type"] == "dir":
```
aiuto, line 298: Can you keep the RHS as manifest.ENTRY_IS_DIR.  That gives us a little more compile time support for not shooting ourselves in the foot.

nacl: > I forgot that piece of code just unpacks, consumes and then throws it away. So there really is no memory concern. I guess this is mostly fine.

I thought the memory usage concerns were within the long-lived Bazel process instead of the python scripts it runs.  Given the above, can we still proceed?  I'm happy to look into cleaning this up.

> Can you keep the RHS as manifest.ENTRY_IS_DIR. That gives us a little more compile time support for not shooting ourselves in the foot.

I'll take a look.

aiuto:
Why the rename to pkg_files_new?

PR #496, opened 2021-12-24
Rename pkg_mklink.src to target

### LINKED ISSUE #480: Adjust attribute names to better match style guide (e.g. src in pkg_mklink)
According to the [.bzl style guide](https://docs.bazel.build/versions/main/skylark/bzl-style.html), attributes named `srcs` should be considered labels pointing to inputs.

`rules_pkg` is kinda odd in this respect -- `srcs` either refers to input files and/or packaging inputs (e.g. something with `PackageFilesInfo`).

One example that is perhaps problematic is `src` in `pkg_mklink`: internally we use `src` as a convention for a rule input that takes a single label/file, but `pkg_mklink` sets that to a string.  This has confused aspects that rely on these conventions.

Consider changing the rule attributes and provide a migration path.  This should be done prior to 1.0 if possible.

aiuto:
Fixed #480


pkg/mappings.bzl
```diff
@@ -494,6 +494,15 @@ pkg_mklink = rule(
...
     attrs = {
+        "target": attr.string(
+            doc = """Link "target", a path on the filesystem.
+
+            This is what the link "points" to, and may point to an arbitrary
+            filesystem path, even relative paths.
+
+            """,
+            mandatory = True,
+        ),
```
nacl, line 505: While I like that the link target should be called "target", I feel there's a potential bit of confusion with "dest" -- it could refer to the link's destination rather than the destination in the package. 

Maybe "dest" should  be "link_name"?  That follows conventions closer to the `ln` (unix) and `mklink` (windows) commands.

pkg/mappings.bzl
```diff
@@ -537,6 +537,30 @@ pkg_mklink = rule(
     provides = [PackageSymlinkInfo],
 )
 
+def pkg_mklink(name, target, dest, attributes=None, src=None, **kwargs):
```
nacl, line 540: So, I wonder how necessary this is.  It may be cleaner to provide a migration path than to introduce new logic with an easily missed deprecation marker.

Something like this might work for a majority of cases:

```sh
buildozer -f - <<EOF
rename src target //:%pkg_mklink
EOF
```

nacl (approved):
LGTM, but I wonder if some more could be done here too.  See comments.

pkg/mappings.bzl
```diff
@@ -537,6 +537,30 @@ pkg_mklink = rule(
...
+
+  Args:
+    name: target name
+    target: target path that the link should point to.
+    dest: the path in the package that should point to the target.
+    attributes: file attributes.
+  """
+  if src:
+    if target:
+      fail("You can not specify both target and source. Use source")
```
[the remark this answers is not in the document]
aiuto: done

pkg/mappings.bzl
```diff
@@ -537,6 +537,30 @@ pkg_mklink = rule(
...
+    print("Warning: pkg_mklink.src is deprecated. Use target.")
```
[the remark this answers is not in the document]
aiuto: Added buildifer. 
There seems to be religion about never printing things except errors. I am an apostate that believes deprecation warnings are of value rather than noise. I gave up fighting the battle for support in bazel.

pkg/mappings.bzl
aiuto: Great. I like link_name better too.

pkg/mappings.bzl
aiuto: I like to leave backwards compatible behavior around for a release or two. It makes day of upgrade less painful.

But this is why we need deprecation warnings in Bazel. The default can be to suppress them, but you have run with a flag to make them errors. Then you can spot them and migrate.  Right now, our philosophy or no warnings forces people to look to other sources to know they need to change something.

Issue #489, opened 2021-12-15
//tests:stamp_test inputs become stale too quickly


nacl:
I often test all of `rules_pkg` using something like `bazelisk test //...`.

When I leave my desk for the day and come back the following day, I sometimes see `//tests:stamp_test` mysteriously fail after making some changes.  This is due to the timestamps no longer being valid:

```
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_apsaltis/67ca5e5a723760c2131afe9d288ecae5/sandbox/darwin-sandbox/9/execroot/rules_pkg/bazel-out/darwin-fastbuild/bin/tests/stamp_test.runfiles/rules_pkg/tests/stamp_test.py", line 104, in test_not_epoch_times_tar
    self.assertTarFilesAreAlmostNew('stamped_tar.tar')
  File "/private/var/tmp/_bazel_apsaltis/67ca5e5a723760c2131afe9d288ecae5/sandbox/darwin-sandbox/9/execroot/rules_pkg/bazel-out/darwin-fastbuild/bin/tests/stamp_test.runfiles/rules_pkg/tests/stamp_test.py", line 82, in assertTarFilesAreAlmostNew
    self.check_mtime(info.mtime, file_path, info.name)
  File "/private/var/tmp/_bazel_apsaltis/67ca5e5a723760c2131afe9d288ecae5/sandbox/darwin-sandbox/9/execroot/rules_pkg/bazel-out/darwin-fastbuild/bin/tests/stamp_test.runfiles/rules_pkg/tests/stamp_test.py", line 63, in check_mtime
    self.fail(
AssertionError: Archive /private/var/tmp/_bazel_apsaltis/67ca5e5a723760c2131afe9d288ecae5/sandbox/darwin-sandbox/9/execroot/rules_pkg/bazel-out/darwin-fastbuild/bin/tests/stamp_test.runfiles/rules_pkg/tests/stamped_tar.tar contains file . with mtime:1639538959, expected:1639586337 +/- 10000
```

It would be nice if the relevant artifacts could be rebuilt when they have expired in the cache.  I don't know how implement this, though.

This problem is never seen in CI, since it builds everything from scratch all the time.

aiuto:
Yes. I see the same. We could up the window to make it happen every 2 or 3 days.
I made my peace with doing a bazel --clean every so often. Even over multiple years, I think that will cost me far less time than trying to solve the problem.

nacl:
[...]

PR #497, opened 2021-12-24
Make stamp_test a little less finicky.

aiuto:
Make //tests:stamp_test a little less finicky.

This addresses most of #489.
- makes the window larger
- improves the warning message

It does not use a fixed stamp value, because that would loose the critical test that we are actually getting a real build date rather than some constant.

### DIFF
--- a/tests/stamp_test.py
+++ b/tests/stamp_test.py
@@ -36,7 +36,7 @@ class StampTest(unittest.TestCase):
   # CI run and have the test run a few seconds later. Ideally, this would
   # be no greater than the expected total CI run time, but the extra margin
   # does not hurt.
-  ALLOWED_DELTA_FROM_NOW = 10000  # seconds
+  ALLOWED_DELTA_FROM_NOW = 30000  # seconds
 
   def check_mtime(self, mtime, file_path, file_name):
     """Checks that a time stamp is reasonable.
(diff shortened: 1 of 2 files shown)


tests/stamp_test.py
```diff
@@ -61,9 +61,12 @@ def check_mtime(self, mtime, file_path, file_name):
     if ((mtime < self.target_mtime - StampTest.ALLOWED_DELTA_FROM_NOW)
         or (mtime > self.target_mtime)):
        self.fail(
-           'Archive %s contains file %s with mtime:%d, expected:%d +/- %d' % (
+           'Archive %s contains file %s with mtime:%d, expected:%d +/- %d.' % (
                file_path, file_name, mtime, self.target_mtime,
-               StampTest.ALLOWED_DELTA_FROM_NOW))
+               StampTest.ALLOWED_DELTA_FROM_NOW) +
+           '  This may be a false positive if your build cache is more than' +
+           ' %s seconds old.  If so, try bazel clean and rerun the test.' %
```
nacl, line 68: Could perhaps phrase this in terms of minutes or hours rather than seconds -- would require less thinking by those unfamiliar.

Not a critical problem, though.

tests/stamp_test.py
```diff
@@ -36,7 +36,7 @@ class StampTest(unittest.TestCase):
   # CI run and have the test run a few seconds later. Ideally, this would
   # be no greater than the expected total CI run time, but the extra margin
   # does not hurt.
-  ALLOWED_DELTA_FROM_NOW = 10000  # seconds
+  ALLOWED_DELTA_FROM_NOW = 30000  # seconds
```
nacl, line 39: It might be nice to make this a day (86400s) to make it a little more concrete.

nacl (approved):
LGTM, but perhaps the duration could be made a little longer and the units changed in the error message?  Not critical, though.

tests/stamp_test.py
aiuto: nano seconds!!!

tests/stamp_test.py
aiuto: Does that make it too wide? I'm really at a loss. 
No matter how wide I make it, I find it still fails when I run it inside Google. Our caching previously built things is good enough to keep it for days. :-(

Issue #498, opened 2021-12-26
Provide a full integration test for pkg_zip


aiuto:
pkg_zip should have an intergration test that
- builts a zip file using an instance of every supported feature
- unzips that file and inspects the file system for the expected results

Doing this with maximum fidelity requires, at a minimum, using the unzip tool which would be the logical default on any given OS. If there are multiple competitors, we would need to test against several versions. So, a robust integration test would require an "unzip" toolchain that finds the right unzip tool for each OS. 

OTOH, doing this completely is a lot of work for little payoff. We currently have tests that use python zipfile to inspect .zip files to see if their content is as expected. The python code base has more testing resources thrown at it than we ever will. I expect it has closer fidelity to the expected behavior on any OS than we could ever achieve. I'm happy to go with the presumption that if the python zipfile module tells us we will get a file with specific attributes, so will the OS local 'unzip'.

nacl:
Another thing to note here is that what python supports in its zipfile framework is not necessarily what is supported by local `unzip` binaries.  In particular, there are alternate compression modes that may not be supported by the standard `unzip` tool and file format.

PR #499, opened 2021-12-26
Add symlink support to pkg_zip

aiuto:
Add symlink support to pkg_zip.  #309

This is, by intent, missing a complete integration test. That is for another quarter #498.


pkg/private/zip/build_zip.py
```diff
@@ -27,6 +27,7 @@
 # Unix dir bit and Windows dir bit. Magic from zip spec
 UNIX_DIR_BIT = 0o40000
 MSDOS_DIR_BIT = 0x10
```
nacl, line 29: [`stat.FILE_ATTRIBUTE_DIRECTORY`](https://docs.python.org/3/library/stat.html?highlight=s_irusr#stat.FILE_ATTRIBUTE_DIRECTORY), which is equivalent?

tests/zip/zip_test.py
```diff
@@ -26,6 +26,7 @@
 # Unix dir bit and Windows dir bit. Magic from zip spec
 UNIX_DIR_BIT = 0o40000
 MSDOS_DIR_BIT = 0x10
+UNIX_RWX_BITS = 0o777
```
nacl, line 29: Use symbolic constants from the `stat` module instead?

tests/zip/BUILD
```diff
@@ -42,6 +42,15 @@ pkg_mkdirs(
     ],
 )
 
+pkg_mklink(
+    name = "link",
+    src = "/usr/local/foo/foo.real",
+    dest = "/usr/bin/foo",
```
nacl, line 48: Does this need to be updated?

nacl (approved):
LGTM, but I think this needs more testing on Windows, or lack of testing should be documented.  I might have some time to look into this next week, but I can't make any guarantees.

See below comments for other ideas, mostly on improvements using the `stat` module.

pkg/private/zip/build_zip.py
```diff
@@ -27,6 +27,7 @@
...
+UNIX_SYMLINK_BIT = 0o120000
```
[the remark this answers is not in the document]
aiuto: I took it from the zip spec rather than stat.S_IFLNK because the API is zip.
Yes, it is identical, but the zip spec does not explicitly say it matches the posix bits.

pkg/private/zip/build_zip.py
```diff
@@ -27,6 +27,7 @@
 # Unix dir bit and Windows dir bit. Magic from zip spec
 UNIX_DIR_BIT = 0o40000
```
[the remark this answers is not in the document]
aiuto: Same reasoning.

tests/zip/BUILD
aiuto: Yes. Done

aiuto:
> LGTM, but I think this needs more testing on Windows, or lack of testing should be documented. I might have some time to look into this next week, but I can't make any guarantees.

symlnks on windows are there own area of fun. Since we are not reading links, what we have should build the same bits on any OS. The question is how they unpack, which is fortunately not our problem.

pkg/private/zip/build_zip.py
aiuto: But that is not available in the macos python implementation. It is unfortunate that they constants vary by OS.

tests/zip/zip_test.py
aiuto: I'm inclined to keep the same style as the other magic bits.

PR #503, opened 2022-01-10
Added support for the built-in tags attribute so that tags can be set on pkg rules

### LINKED ISSUE #504: Setting tags does not impact the execution of rules_pkg rules.
Hi,

I am using `pkg_tar` and I would like to set the bazel tag `no-remote-cache` to avoid caching the tar build result. The `tags` attribute is a built-in from Bazel and it is possible to set `tags = ["no-remote-cache"]` on the `pkg_tar` rule, however setting this value does not appear to have any impact. When building the target, the Bazel status line still reads `remote cache, multiplex worker`

It seems like that the `execution_requirements` attribute on Bazel `ctx.actions.run` needs to be set with the values from tags like 
```
execution_requirements = {
    "no-remote-cache": "1",
}
```

twheys:
Fixes #504 

I'd like to set tags, specifically `no-remote-cache`, on my use of the `pkg_tar` rule. This is a built-in rule from Bazel, but it seems it needs to be connected to the associated action in the custom rule. I've created a small utility to generate execution requirements based on `ctx.attr.tags` if it is present, which allows the attribute to be set on a rule.

I've added this for all `pkg_*` rules, please let me know if you have any feedback.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -32,7 +32,12 @@ load(
     "process_src",
     "write_manifest",
 )
-load("//pkg/private:util.bzl", "setup_output_files", "substitute_package_variables")
+load(
+    "//pkg/private:util.bzl",
+    "setup_output_files",
+    "substitute_package_variables",
+    "compute_execution_requirements",
+)
 load("//pkg/private/deb:deb.bzl", _pkg_deb = "pkg_deb")
 load("//pkg/private/zip:zip.bzl", _pkg_zip = "pkg_zip")
 
@@ -228,6 +233,7 @@ def _pkg_tar_impl(ctx):
     files.append(arg_file)
     ctx.actions.write(arg_file, "\n".join(args))
 
+
     ctx.actions.run(
         mnemonic = "PackageTar",
         progress_message = "Writing: %s" % output_file.path,
@@ -243,6 +249,7 @@ def _pkg_tar_impl(ctx):
             "PYTHONUTF8": "1",
         },
         use_default_shell_env = True,
+        execution_requirements = compute_execution_requirements(ctx),
     )
     return [
         DefaultInfo(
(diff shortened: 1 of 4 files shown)


aiuto:
Can you create an issue about this which describe the exact problem and intent.
I am not understanding what is going on here.  It seems to conflate execution requirrements and tags.
Most tags should just work, because we pass through kwargs.  The bit about computing execution requirements from tags seems strange. It seems like we should just have an ability to set execution requirements.

twheys:
Yeah, I'll create an issue ticket

twheys:
Alright, I created an issue describing my use case and the issue. I agree, it's a bit strange to need to set execution-requirements with the tags. The tags attribute is built-in and inherently available on the pkg_tar rule, but in my testing it didn't seem to have any affect on the execution of the rule.

The Bazel documentation describes this feature [here](https://docs.bazel.build/versions/3.1.0/be/common-definitions.html): 
> Bazel modifies the behavior of its sandboxing code if it finds the following keywords in the tags attribute of any test rule or genrule, or the keys of execution_requirements for any Starlark action.

So it seems the tags only apply if they are passed in to execution_requirements.

aiuto:
This would forward all tags as execution requirements. That is just not right. It subverts the meaning of tags.

aiuto:
To be fair, the bazel behaviour seems inconsistent, w.r.t. tags applying for tests but not rules, but let's get to the bottom of that.  I want to examine what other rule sets are doing in this space and find some unified expectations.

We can't forward every tag as an exec requirement - that has potential to cause clashes with common uses and future executor features. And most of the onese mentioned in the bazel docs are about running the test program rather than executing the action which builds the test, so it is not obvious they apply. Exclusive and manual, for example are meaningless except for a test.

FWIW: For the most part, you want the rule author to set action execution requirements rather than let individual users do it. But that is a discussion for bazel-discuss rather than this rule set.

PR #513, opened 2022-02-01
Add the rpmbuild version into the toolchain

aiuto:
This is a precursor PR to using the version to detect support for source date epoch.

It does not solve the problem of why the source date epoch tests fail on my 4.17.0 version of rpmbuild.


nacl (approved):
LGTM

PR #519, opened 2022-02-03
Fix some lint warnings.

aiuto:
Turn # comments to module docstrings.
Light attribute reorderings.
This is not the entire repository. I want to do only trivial ones in this pass.

The only intent is to shut lint up. Not to improve the descriptions.  Those are better done individually as people work on their slices.


aiuto:
I still don't understand the robots.txt thing. It appears in another PR too.

jylinv0 (approved):

PR #524, opened 2022-02-04
Try to make the docs a little better.

aiuto:
- Fix stardoc adding `<p align=center>` in markdown tables.
- Add rough capability to handle macros that wrap a rule.

The wrapping technique is:
- In the wrapping macro, use the docstring  @wraps(some_rule)
- Then we don't emit the docs for the wrapper
- In the docs for some_rule, replace the string "some_rule" with the name of the wrapper macro.

It is a first attempt to make things better.  It is crude, but it mostly works. If Stardoc ever adds this capability [(feature request)](https://github.com/bazelbuild/stardoc/issues/120) we can move to that.


doc_build/BUILD
```diff
@@ -11,16 +11,39 @@
...
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
 load("@bazel_stardoc//stardoc:stardoc.bzl", "stardoc")
+load("@rules_python//python:defs.bzl", "py_library")
 load("//:version.bzl", "version")
 
-# HOW TO:
-#   bazel build //doc_build:reference
-#   cp bazel-bin/doc_build/reference.md docs/latest.md
-#   git commit -m 'update docs' docs/latest.md
+filegroup(
```
nacl, line 27: We should probably be using `pkg_files`es for these at some point.  It's kinda ironic that we aren't.

doc_build/BUILD
```diff
@@ -11,16 +11,39 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+"""Generate the reference documentation.
+
+How to:
+   bazel build //doc_build:reference
+   cp bazel-bin/doc_build/reference.md docs/latest.md
+   git commit -m 'update docs' docs/latest.md
```
nacl, line 19: It would be nice if we could `bazel run` this as one command, but not a critical matter.

doc_build/merge.py
```diff
@@ -0,0 +1,72 @@
...
+import sys
+import typing
+
+
+ID_RE = re.compile(r'<a id="#(.*)">')
+WRAPS_RE = re.compile(r'@wraps\((.*)\)')
+CENTER_RE = re.compile(r'<p align="center">([^<]*)</p>')
+
+
+def merge_file(file: str, out, wrapper_map:typing.Dict[str, str]) -> None:
```
nacl, line 31: This forces use of python 3.6 or newer.  Not that it's a problem, but this might cause interesting problems on older platforms.  Guess we'll deal with them if/when they appear.

doc_build/merge.py
```diff
@@ -0,0 +1,72 @@
...
+  with open(file, 'r') as inp:
+    content = inp.read()
+    m = ID_RE.search(content)
+    this_pkg = m.group(1) if m else None
```
nacl, line 35: Could we use an explicit comparison to `None` or `False` here and below?

doc_build/merge.py
```diff
@@ -0,0 +1,72 @@
...
+    m = WRAPS_RE.search(content)
+    if m:
+      # I wrap something, so don't emit me.
+      wrapper_map[m.group(1)] = this_pkg
+      return
+    # If something wraps me, rewrite myself with the wrapper name.
+    if this_pkg in wrapper_map:
+      content = content.replace(this_pkg, wrapper_map[this_pkg])
+    merge_text(content, out)
+
```
nacl, line 45: Nit: double whitespace needed?

doc_build/BUILD
```diff
@@ -42,9 +65,8 @@ genrule(
     name = "reference",
     srcs = ["%s.md" % rule for rule, _ in ORDER],
     outs = ["reference.md"],
-    # Compensate for https://github.com/bazelbuild/stardoc/issues/118.
-    # Convert escaped HTML <li> back to raw text
-    cmd = "cat $(SRCS) | sed -e 's/&lt;li&gt;/<li>/g' >$@",
+    cmd = "$(location :merge) $(SRCS) >$@",
+    tools = [":merge"],
```
nacl, line 69: `exec_tools`?

nacl (approved):
Definitely an improvement, but it's a bit hacky.

Might be worth saying that in the source file in question, but otherwise should be fine for our immediate purposes.

doc_build/BUILD
aiuto: > Might be worth saying that in the source file in question, but otherwise should be fine for our immediate purposes.

The PR description does say "crude, but mostly works".
My actual goal is not to make this code good. It is to provide salient examples of what Stardoc should do and get the functionality built on that side. Then everyone benefits.

aiuto: filegroup to pkg_files for this.
Yes. You have a good point. But
- another PR, because we would want to do it everywhere
- the entire 'srcs', recursive idiom is an ugly wart on this. Sometimes, you want rules to just look at your source repo overall, and not at package boundaries
- bazel is getting new native rules to help gather subdirectories exactly for this use case, so let's wait until those roll out.

doc_build/merge.py
aiuto: Isn't 3.6 EOL?

doc_build/merge.py
aiuto: DOne

dannysullivan (approved):

doc_build/BUILD
aiuto: I could. Maybe some day. Perhaps a pkg_install command eventually.

doc_build/merge.py
aiuto: Why?  This is typical Python usage.

doc_build/BUILD
aiuto: done

PR #525, opened 2022-02-04
Convert pkg_tar to using ctx.args to build the arg string.

aiuto:
This enables us to only spill to params files when we have to, saving an action most of the time.  It also opens up the possibility to always spill to params files on Windows, so that we will never run into issues with arg quoting issues.

### DIFF
--- a/pkg/pkg.bzl
+++ b/pkg/pkg.bzl
@@ -80,27 +80,26 @@ def _pkg_tar_impl(ctx):
     # Find a list of path remappings to apply.
     remap_paths = ctx.attr.remap_paths
 
+    # Start building the arguments.
+    args = ctx.actions.args()
+    args.add("--root_directory", ctx.attr.package_base)
+    args.add("--output", output_file.path)
+    args.add("--mode", ctx.attr.mode)
+    args.add("--owner", ctx.attr.owner)
+    args.add("--owner_name", ctx.attr.ownername)
+
     # Package dir can be specified by a file or inlined.
     if ctx.attr.package_dir_file:
         if ctx.attr.package_dir:
             fail("Both package_dir and package_dir_file attributes were specified")
-        package_dir_arg = "--directory=@" + ctx.file.package_dir_file.path
+        args.add("--directory", "@" + ctx.file.package_dir_file.path)
         files.append(ctx.file.package_dir_file)
     else:
         package_dir_expanded = substitute_package_variables(ctx, ctx.attr.package_dir)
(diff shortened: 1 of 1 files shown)


jylinv0 (approved):

pkg/pkg.bzl
```diff
@@ -187,25 +186,23 @@ def _pkg_tar_impl(ctx):
         )
 
     if ctx.attr.modes:
-        args += [
-            "--modes=%s=%s" % (_quote(key), ctx.attr.modes[key])
-            for key in ctx.attr.modes
-        ]
+        for key in ctx.attr.modes:
+            args.add("--modes", "%s=%s" % (_quote(key), ctx.attr.modes[key]))
```
nacl, line 190: Could this loop and the ones below be replaced with `args.add_all`?  Might be something we could nicely do with a lambda in Bazel 5, if/when we want to break compatibility.

pkg/pkg.bzl
```diff
@@ -80,27 +80,26 @@ def _pkg_tar_impl(ctx):
...
+    args.add("--mode", ctx.attr.mode)
+    args.add("--owner", ctx.attr.owner)
+    args.add("--owner_name", ctx.attr.ownername)
+
     # Package dir can be specified by a file or inlined.
     if ctx.attr.package_dir_file:
         if ctx.attr.package_dir:
             fail("Both package_dir and package_dir_file attributes were specified")
-        package_dir_arg = "--directory=@" + ctx.file.package_dir_file.path
+        args.add("--directory", "@" + ctx.file.package_dir_file.path)
```
nacl, line 95: Do we really need to get the `path` here?  Would something like this work?

```starlark
args.add("--directory", ctx.file.package_dir_file, format = "@%s")
```

nacl (approved):
Some small nitty things, but otherwise LGTM.

pkg/pkg.bzl
aiuto: `add_all` does not work for this case. It does something like 
`--modes a=b,c=d,....` but we need '--modes=a=b --modes=c=d ...`
The same for all the others.

pkg/pkg.bzl
aiuto: That will give us the same thing. 
The rest of the code is explicit using `<f>.path` at the boundary between rule and the action. I think that is clearer than relying on the implicit conversion of file to file.path done by `%s`. 
Then there is the question of which is more efficient. The format solution might be slightly better, but this is only one arg for the action. Overall there is only the copy of this one path to save.

PR #530, opened 2022-02-08
improve the docs table of contents

aiuto:
Make the docs table of contents more useful by including the .bzl files where the rules are defined. E.g.  //pkg:zip.bzl%pkg_zip.   Without that, you just have to know.

It would be nice if stardoc had support for adding the bzl path out of the box.
https://github.com/bazelbuild/stardoc/issues/120


nacl (approved):
LGTM

PR #532, opened 2022-02-09
tests for zip tree artifact

nickbreen:
Added tests for zip with tree artifacts (currently failing due to not being implemented).


nickbreen:
I need to change the email for my commit to be covered by my employer's CLA.

nickbreen:
OK, need to use a different github account.

PR #537, opened 2022-02-12
Add tree artifact support to pkg_zip

### LINKED ISSUE #309: pkg_zip supports pkg_filegroup framework

### LINKED ISSUE #311: pkg_deb supports pkg_filegroup framework
The idea is that we could take a `srcs` list of inputs instead of a tar file as `data`.
Instead of having two actions, one to build the .tar and the second to make the .deb from it, we would have a single on which builds a synthetic tar directly from the inputs.  This would save a copy.

aiuto:
Fixes #309

It is a largish change only because I refactored biuld_zip from free methods to a class.
That made it easier to pass around the default mode and timestamp.


pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,145 @@ def parse_date(ts):
...
+    # paths should not have a leading ./
+    dest = '' if dest == '.' else dest + '/'
+
+    to_write = {}
+    for root, dirs, files in os.walk(tree_top):
+      # While `tree_top` uses '/' as a path separator, results returned by
+      # `os.walk` and `os.path.join` on Windows may not.
+      root = os.path.normpath(root).replace(os.path.sep, '/')
+
+      dirs = sorted(dirs)
```
jylinv0, line 187: Don't we need to create dir entries?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,145 @@ def parse_date(ts):
...
+        f_mode = 0o755 if os.access(content_path, os.X_OK) else 0o644
+      else:
+        f_mode = mode
+      entry_info = self.make_zipinfo(path=path, mode=f_mode)
+      entry_info.compress_type = zipfile.ZIP_DEFLATED
+      if not content_path:
+        self.zip_file.writestr(entry_info, '')
+      else:
+        with open(content_path, 'rb') as src:
+          self.zip_file.writestr(entry_info, src.read())
```
jylinv0, line 209: Would `add_tree` takes care the cases when content_path points to an empty file or a link?

pkg/private/zip/build_zip.py
aiuto: That is a good question.

We don't have to. unzip will create them as needed.
And, since zip can not specify owners, and the directories should all  755, I don't see the point.

In the future, we *might* want the ability to let the mode specified on the tree artifact apply to intermediate directories, but I would rather do that when someone comes up with a good use case, and treat that as a separate feature requreest.

aiuto: Line removed.

pkg/private/zip/build_zip.py
```diff
@@ -125,12 +217,14 @@ def main(args):
...
     default_mode = int(args.mode, 8)
 
-  with zipfile.ZipFile(args.output, mode='w') as zip_file:
-    if args.manifest:
-      with open(args.manifest, 'r') as manifest_fp:
-        manifest = json.load(manifest_fp)
-        for entry in manifest:
-          _add_manifest_entry(args, zip_file, entry, default_mode, ts)
+  if not args.manifest:
+    raise Exception('Missing --manifest')
```
nacl, line 221: Maybe add `required=True` to the declaration of this argument in `_create_argument_parser`?

tests/zip/BUILD
```diff
@@ -206,8 +211,9 @@ py_test(
         # it with this name.
         "test_zip_out.foo",
         "test_zip_package_dir0.zip",
-        "test_zip_timestamp.zip",
         "test_zip_permissions.zip",
+        "test_zip_timestamp.zip",
+        "test_zip_tree",
```
nacl, line 216: Any particular reason why this one doesn't end with `.zip` and the others do?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
+    tree_top = os.path.normpath(tree_top).replace(os.path.sep, '/')
+
+    # Again, we expect /-style paths.
+    dest = destpath.strip('/')  # redundant, dests should never have / here
+    dest = os.path.normpath(dest).replace(os.path.sep, '/')
+    # paths should not have a leading ./
+    dest = '' if dest == '.' else dest + '/'
+
+    to_write = {}
+    for root, dirs, files in os.walk(tree_top):
```
nacl, line 182: Add `onerror` here for improved robustness?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
+      # While `tree_top` uses '/' as a path separator, results returned by
+      # `os.walk` and `os.path.join` on Windows may not.
+      root = os.path.normpath(root).replace(os.path.sep, '/')
+
+      rel_path_from_top = root[len(tree_top):].lstrip('/')
+      if rel_path_from_top:
```
nacl, line 188: Please compare to something here.  Empty string?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
+      destpath: the path under which to place the files
+      mode: if not None, file mode to apply to all files
+    """
+
+    # We expect /-style paths.
+    tree_top = os.path.normpath(tree_top).replace(os.path.sep, '/')
+
+    # Again, we expect /-style paths.
+    dest = destpath.strip('/')  # redundant, dests should never have / here
+    dest = os.path.normpath(dest).replace(os.path.sep, '/')
```
nacl, line 177: Can paths end with `\` on Windows platforms here?  Maybe swap the order of these functions?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
+    # paths should not have a leading ./
+    dest = '' if dest == '.' else dest + '/'
```
nacl, line 179: The `/` may be removed on the end on line 176 only to potentially be readded here.  Maybe just `lstrip()` there?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
-    zip_file.writestr(entry_info, '')
-  elif entry_type == manifest.ENTRY_IS_LINK:
-    entry_info.compress_type = zipfile.ZIP_STORED
-    # Set directory bits
-    entry_info.external_attr |= (UNIX_SYMLINK_BIT << 16)
-    zip_file.writestr(entry_info, src)
-  # TODO(#309): All the rest
+
+class ZipWriter(object):
+
```
nacl, line 74: Does this class need documentation?

pkg/private/zip/build_zip.py
```diff
@@ -68,53 +69,144 @@ def parse_date(ts):
...
+  def add_manifest_entry(self, options, entry):
+    """Add an entry to the zip file.
+
+    Args:
+      options: parsed options
+      zip_file: ZipFile to write to
+      entry: manifest entry
+    """
+
```
nacl, line 131: Is there supposed to be an empty line after the docstrings like this?  This seems inconsistent with the preexisting functions in this file.

nacl (approved):
LGTM modulo comments.

This change was a little hard to read: the creation of the `ZipWriter` class probably should have been done separately from the TreeArtifact integration so that it was more obvious what changed in `add_manifest_entry`.

pkg/private/zip/build_zip.py
aiuto: Removed

pkg/private/zip/build_zip.py
aiuto: destpath would be created by our tools, so it will have a '/'.
Then normpath may turn / into \, so we have to turn it back. 
I thinks this works right as is.

tests/zip/BUILD
aiuto: No, just an oversight. Fixed.

pkg/private/zip/build_zip.py
aiuto: Yes. That makes sense. Done

aiuto:
> LGTM modulo comments.
> 
> This change was a little hard to read: the creation of the `ZipWriter` class probably should have been done separately from the TreeArtifact integration so that it was more obvious what changed in `add_manifest_entry`.

Yes. This ended up as to PRs. I should have made it 3.

pkg/private/zip/build_zip.py
aiuto: Yes. 176 intentionally makes the invariant that there are no '/' on either end.  So something like 
/./' becomes '.'.   Then we normalize to a good path for the OS, which might turn '' to '.'. Then we make sure we can not end up with './'. 
I don't care if I do extra work in any given line, as long as the invariants are easy to understand.

pkg/private/zip/build_zip.py
aiuto: I'm not going to lose sleep over that now. It is private to this program. The only point of it is to hold things without having globals. It never was documented before.

pkg/private/zip/build_zip.py
aiuto: Idiomatic python would not compare to an empty string.
Are we seeing different corporate style guides?

pkg/private/zip/build_zip.py
aiuto: That should be its own PR. If I add it, I then need a reliable test to trigger it.
That might be more effort than this entire PR.

Issue #531, opened 2022-02-09
pkg_tar should not put  ./ at the beginning of all paths


aiuto:
It does not match the behavior or regular tar. It is also unneeded. Time to take it out.

Consider a directory named only ".", or "./".  On untar, should it change the permissions of the current directory?  That is probably never what the user would want. So that file simply should not exist.

For any other file it is redundant, since tar extracts  "./a" identically to "a".

I had some conversations with people building docker containers with pkg_tar and pkg_deb.  They never understood what the leading "./" was for in bazel's rules. Most containers us straight, relative file paths.

PR #540, opened 2022-02-13
Stop archive writer from always adding './' as a root path.

aiuto:
- clean up tests appropriately
- remove achive.add_dir because we do not use it

This is a precursor change to removing the unwanted leading ./ from tar writers.
See #50, #531 

A side effect of this change is that most tar output will no longer have the first entry be a directory name '.'.
This was always useless and potentially dangerous behavior. You can not create '.', but if you had a tarball
with strange permissions on the '.' entry, would that presume to set the permissions on a previously created directory?
No good can come of that.  If someone expectst that in their tests, it is a breaking change for them.
We will have to bump the minor version number in a followup PR.

### DIFF
--- a/pkg/private/archive.py
+++ b/pkg/private/archive.py
@@ -122,7 +122,7 @@ class TarFileWriter(object):
                name,
                compression='',
                compressor='',
-               root_directory='.',
+               root_directory='',
                default_mtime=None,
                preserve_tar_mtimes=True):
     """TarFileWriter wraps tarfile.open().
@@ -174,11 +174,14 @@ class TarFileWriter(object):
                                               stdout=open(name, 'wb'))
       self.fileobj = self.compressor_proc.stdin
     self.name = name
-    self.root_directory = root_directory.rstrip('/').rstrip('\\')
-    self.root_directory = self.root_directory.replace('\\', '/')
+    # tarfile uses / instead of os.path.sep
+    self.root_directory = root_directory.replace(os.path.sep, '/').rstrip('/')
+    if self.root_directory:
+      self.root_directory = self.root_directory + '/'
 
     self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj)
     self.members = set([])
+    # The directories we have created so far
     self.directories = set([])
 
   def __enter__(self):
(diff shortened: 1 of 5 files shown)


aiuto:
@beasleyr-vmw

pkg/private/archive.py
```diff
@@ -346,9 +294,9 @@ def add_file(self,
         tarinfo.size = os.fstat(f.fileno()).st_size
         self._addfile(tarinfo, f)
     else:
-      if kind == tarfile.DIRTYPE:
-        self.directories.add(name)
       self._addfile(tarinfo)
+    if kind == tarfile.DIRTYPE:
+      self.directories.add(name.rstrip('/'))
```
jylinv0, line 299: nit: `name` is not going to end with `/` before reaching this line, so the rstripping seems unnecessary?

jylinv0 (approved):

pkg/private/archive.py
```diff
@@ -174,11 +174,14 @@ def __init__(self,
...
-    self.root_directory = self.root_directory.replace('\\', '/')
+    # tarfile uses / instead of os.path.sep
+    self.root_directory = root_directory.replace(os.path.sep, '/').rstrip('/')
+    if self.root_directory:
+      self.root_directory = self.root_directory + '/'
 
     self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj)
     self.members = set([])
+    # The directories we have created so far
     self.directories = set([])
```
nacl, line 185: Nit: just saying `set` should be vaild and still produce the same result as this.

tests/tar/pkg_tar_test.py
```diff
@@ -113,23 +108,20 @@ def test_strip_prefix_dot(self):
 
   def test_strip_files_dict(self):
     content = [
-        {'name': '.'},
         {'name': './not-etc'},
         {'name': './not-etc/mapped-filename.conf'},
```
nacl, line 112: Help me understand this -- is this change only supposed to be removing the `.` entry from the tarfile?

Reading the commit message, I wasn't sure if this change would remove the `.` from the tarfile altogether or just remove that one entry.  Might be worth clarifying that point

nacl (approved):
Some nitty things, but seems fine otherwise.

I'd like to see `archive.py` replaced with something else eventually, but that'll probably be post-1.0.

pkg/private/archive.py
aiuto: Yes. add_root promises that name does not end in /.
Changed to an assert.

pkg/private/archive.py
aiuto: Yes. Changed.

pkg/private/archive.py
```diff
@@ -174,11 +174,14 @@ def __init__(self,
                                               stdout=open(name, 'wb'))
       self.fileobj = self.compressor_proc.stdin
     self.name = name
-    self.root_directory = root_directory.rstrip('/').rstrip('\\')
-    self.root_directory = self.root_directory.replace('\\', '/')
+    # tarfile uses / instead of os.path.sep
+    self.root_directory = root_directory.replace(os.path.sep, '/').rstrip('/')
+    if self.root_directory:
+      self.root_directory = self.root_directory + '/'
```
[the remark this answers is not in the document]
aiuto: I don't want to rely on path manipulations because I want it explicit that we are
normalizing from the os sep to /.   We may be processing  paths of generated output files which can have \ in them on windows.

tests/tar/pkg_tar_test.py
aiuto: This change only removes the '.' entry from the tar and zip files. 
That never should have been been in an archive.

But the side effect of the change is also that a followup in pkg_tar can allow the leading './' be removed, without archive sticking it right back in again. That will be a big enough PR in itself, because it adds a layer of new tests.

pkg/private/archive.py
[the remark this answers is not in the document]
nacl: IMO this shouldn't matter and would make the codebase a little more consistent.

For example, our UNIXy archive writer module could be written to assume that they receive UNIX-style paths, with the normalization to UNIX-style paths done earlier prior to invoking the archive writer.

PR #541, opened 2022-02-14
Add `artifact_name` to `print_relnotes`

### LINKED ISSUE #435: `print_rel_notes` should support customizing `tarball_name`

aiuto:
This allows you to use artifact names that do not follow the standard convention.

Fixes #435

### DIFF
--- a/pkg/releasing/defs.bzl
+++ b/pkg/releasing/defs.bzl
@@ -4,6 +4,7 @@ def print_rel_notes(
         name,
         repo,
         version,
+        artifact_name = None,
         outs = None,
         setup_file = "",
         deps_method = "",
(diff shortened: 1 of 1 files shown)


aiuto:
@guw

guw:
@aiuto just got to testing this with main branch but it doesn't seem to work. I still don't get the proper download url generated into `http_archive`

aiuto:
Can you include a BUILD file snippet so I can make sure that you and I are
doing the same thing.

guw:
@aiuto BUILD file is here: https://github.com/salesforce/bazel-jdt-java-toolchain/blob/07571cc819ecd76fff2908aaad2fb9a07034cef7/dist/BUILD#L30

My goal is to build `rules_jdt.tar.gz` and publish this.
`bazel build //dist:rules_jdt` produces `bazel-bin/dist/rules_jdt.tar.gz`

When I run `cat bazel-bin/dist/relnotes.txt` is still uses `https://github.com/salesforce/bazel-jdt-java-toolchain/releases/download/latest/bazel-jdt-java-toolchain-latest.tar.gz` as the tar.

guw:
@aiuto https://github.com/bazelbuild/rules_pkg/pull/561

aiuto:
Thanks for the PR.   I edited the title and description a bit to be more
specific about why the fix was needed.  I believe PR descriptions should be
mostly self contained and not require one to know context from a bug
conversation.

PR #542, opened 2022-02-15
Move make_rpm_test to tests/rpm

aiuto:
This makes it easier to suppress all the rpm test on platforms where we do not have rpmbuild. While I could argue that make_rpm_test tests the portable part (make_rpm) I don't think there is much value in having it split out
- we actually do test it in CI on centos, so we will fail if not there
- the main developers all have rpm capable machines, so we test before sending PRs
- the test mocks up so much I wonder if it has value. The last time we updated it was 2019


nacl (approved):
LGTM

PR #547, opened 2022-02-15
remove the obsolete deb_packages code. 

aiuto:
Having web search find this is a disservice to users


jylinv0:
There're a few mentions of deb_packages, e.g. OWNERS. Would you remove them as well?

aiuto:
Good catch. I removed from developers.md, but I kept codeowners for now because I want to make sure we don't remove the forwarding note too soon. I want to give it another year.

jylinv0 (approved):

PR #551, opened 2022-02-19
Refactor archive.py to make adding parent directories easier to understand

aiuto:
This is a precursor PR to cleaning up the use of './' in pkg_tar and
pkg_deb. The followups may require more changes to the tar writer,
and doing this refactor first will make anything downstream easier to
reason about.

The core of this change is to change adding the parent tree from
a recursive call of add_file to an iterative one building up a
path from root to tail. The recursion had difficult to understand
termination conditions, with checks sprinkled in a few places.
The refactoring makes the concept clearer.

#531

### DIFF
--- a/pkg/private/archive.py
+++ b/pkg/private/archive.py
@@ -316,8 +316,6 @@ class TarFileWriter(object):
     if name == '.':
       return
     name = self.add_root_prefix(name)
-    if kind == tarfile.DIRTYPE and name in self.directories:
-      return
     if name in self.members:
       return
 
--- a/pkg/private/archive.py
+++ b/pkg/private/archive.py
@@ -35,6 +35,7 @@ COMPRESSIONS = ('', 'gz', 'bz2', 'xz') if HAS_LZMA else ('', 'gz', 'bz2')
 # See: https://github.com/bazelbuild/bazel/issues/1299
 PORTABLE_MTIME = 946684800  # 2000-01-01 00:00:00.000 UTC
 
+_DEBUG_VERBOSITY = 0
 
 class SimpleArFile(object):
   """A simple AR file reader.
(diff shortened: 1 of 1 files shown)


pkg/private/archive.py
```diff
@@ -252,25 +315,17 @@ def add_file(self,
       return
     if name == '.':
       return
-    name = self.add_root(name)
-    # Do not add a directory that is already in the tar file.
+    name = self.add_root_prefix(name)
     if kind == tarfile.DIRTYPE and name in self.directories:
       return
+    if name in self.members:
```
jylinv0, line 321: Do we need to consider a directory and a file of same name may co-exist? If not, we can delete L319~L320, where we check duplicate directories because we'll check if it exist in self.members. If so, we'll need to check `kind` in this if.

pkg/private/archive.py
```diff
@@ -317,15 +369,16 @@ def add_tar(self,
       name_filter: filter out file by names. If not none, this method will be
           called for each file to add, given the name and should return true if
           the file is to be added to the final tar and false otherwise.
-      root: place all non-absolute content under given root directory, if not
-          None.
+      root: place all content under given root directory, if not None.
```
jylinv0, line 372: Could I put `root=.`, `root=../` or `root=./foo`?

aiuto: That will become unsupported in the followup CL that fixes pkg_tar to not have it.

pkg/private/archive.py
aiuto: A directory and file of the same name can never coexist. 

Those 2 lines can be deleted. FWIW, I wanted to eliminate self.directories, but there are tests which peer into this class and look at self.members. In order to keep a promise that this PR only changed the implementation, and not the external behavior, I did not do that here.  I may in a future cleanup.

jylinv0 (approved):

Issue #284, opened 2021-01-28
Deprecation: archive_name to be removed in all pkg_* rules


aiuto:
`package_file_name` is a superset of the capabilities of `archive_name`. 
Estimated date: 2021-08-01 or version 1.0, whichever comes first

PR #552, opened 2022-03-01
Get rid of the long deprecated archive_name and extension from pkg_zip.

### LINKED ISSUE #414: Cannot use `out` to set file name in `pkg_zip`
The docs of `pkg_zip` say
> Use out or package_file_name to specify the output file name

However, using out results in

```
ERROR: Traceback (most recent call last):
        File "/workspaces/semmle-code/extractor-python/BUILD.bazel", line 73, column 5, in <toplevel>
                dist(
        File "/workspaces/semmle-code/common.bzl", line 12, column 12, in dist
                pkg_zip(name = name + ".zip", out = name, srcs = [name])
        File "/home/codespace/.cache/bazel/_bazel_codespace/3c41b00dfcd8c4fb714360e042371c98/external/rules_pkg/pkg.bzl", line 723, column 17, in pkg_zip
                pkg_zip_impl(
Error in pkg_zip_impl: rule(...) got multiple values for parameter 'out'
```

Presumably this is because `out` is unconditionally set. Instead, it should do something ala
```
if not kwargs.get("out"):
  kwargs["out"] = archive_name + "." + extension
```
instead.

aiuto:
- Make 'out' work in a reasonable way. Fixes #414
- Partial fix for #284

### DIFF
--- a/pkg/private/zip/zip.bzl
+++ b/pkg/private/zip/zip.bzl
@@ -138,7 +138,10 @@ limited to a granularity of 2 seconds.""",
         ),
 
         # Common attributes
-        "out": attr.output(mandatory = True),
+        "out": attr.output(
+            doc = """output file name. Default: name + ".zip".""",
+            mandatory = True,
+        ),
         "package_file_name": attr.string(doc = "See Common Attributes"),
         "package_variables": attr.label(
             doc = "See Common Attributes",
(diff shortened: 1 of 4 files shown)


pkg/private/zip/zip.bzl
```diff
@@ -138,7 +138,10 @@ limited to a granularity of 2 seconds.""",
         ),
 
         # Common attributes
-        "out": attr.output(mandatory = True),
+        "out": attr.output(
+            doc = """output file name. Default: name + ".zip".""",
+            mandatory = True,
```
jylinv0, line 143: A `default` declaration seems conflict with the mandatory flag?

aiuto: It is not an actual default.  The default is done in the macro wrapper.
We do this kind of thing all the time because stardoc does not have the ability to merge the declaration from the rule with the wrapper you use around it.
Lots of discussion about that in https://github.com/bazelbuild/stardoc/issues/27

jylinv0: I meant if it's a mandatory field, I assume it will throw error when `out` is not given, and default value is meaningless. It's confusing when doc says there'll be a default value to a mandatory field.

jylinv0 (approved):

pkg/private/zip/zip.bzl
aiuto: Yes. It is confusing, but I don't have a good way around it.
The field *is* mandatory at the rule level, but is optional (with a default) at the wrapper macro.  This is a common idiom in starlark rules.

PR #554, opened 2022-03-03
Fix pkg_tar to not add the ./ to the prefix of every member.

### LINKED ISSUE #549: pkg_tar has both package_dir and package_base.
This is needlessly confusing. They both almost do the same thing, and the reason for the semantic difference is lost to history.

aiuto:
Fixes: #531

- Remove leading './' from merged tar files. Also applies to pkg_deb.
- Remove package_base from pkg_tar. Closes #549
- Use build_tar --directory as the prefix to add to everything.

Note that the code is intentionaly not testing use cases that try to add "./"
as an explicit root. This might still work, but it is a bad idea, for the
reasons described in #531 . If someone comes up with a real need for
using a leading "./" AND it does not work, they can file a feature
request to support it.

### DIFF
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -25,6 +25,17 @@ from pkg.private import build_info
 from pkg.private import manifest
 
 
+def normpath(path):
+  """Normalize a path to the format we need it.
+  
+  os.path.normpath changes / to \ on windows, but tarfile needs / style paths.
+
+  Args:
+    path: (str) path to normalize.
+  """
+  return os.path.normpath(path).replace(os.path.sep, '/')
+
+
 class TarFile(object):
   """A class to generates a TAR file."""
 
@@ -53,8 +64,7 @@ class TarFile(object):
     self.tarfile.close()
 
   def normalize_path(self, path: str) -> str:
-    # Note: normpath changes / to \ on windows, but tarfile needs / style paths.
-    dest = os.path.normpath(path).replace(os.path.sep, '/')
+    dest = normpath(path)
     # paths should not have a leading ./
     if dest.startswith('./'):
       dest = dest[2:]
@@ -118,7 +128,7 @@ class TarFile(object):
       ids = (0, 0)
     if names is None:
       names = ('', '')
-    dest = os.path.normpath(dest).replace(os.path.sep, '/')
+    dest = normpath(dest)
     self.tarfile.add_file(
         dest,
         content='' if kind == tarfile.REGTYPE else None,
@@ -169,7 +179,7 @@ class TarFile(object):
       names: (username, groupname) for the file to set ownership.  An empty
         file will be created as `destfile` in the layer.
     """
-    symlink = os.path.normpath(symlink).replace(os.path.sep, '/')
+    symlink = normpath(symlink)
     self.tarfile.add_file(
         symlink,
         tarfile.SYMTYPE,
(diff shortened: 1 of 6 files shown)


pkg/private/tar/build_tar.py
```diff
@@ -31,27 +31,40 @@ class TarFile(object):
...
     return self
 
   def __exit__(self, t, v, traceback):
     self.tarfile.close()
 
+  def normalize_path(self, path: str) -> str:
+    # Note: normpath changes / to \ on windows, but tarfile needs / style paths.
+    dest = os.path.normpath(path).replace(os.path.sep, '/')
+    # paths should not have a leading ./
+    if dest.startswith('./'):
```
jylinv0, line 59: Should we use `for` instead of `if`?

pkg/private/tar/build_tar.py
```diff
@@ -64,18 +77,14 @@ def add_file(self, f, destfile, mode=None, ids=None, names=None):
...
+    dest = self.normalize_path(destfile)
     # If mode is unspecified, derive the mode from the file's mode.
     if mode is None:
       mode = 0o755 if os.access(f, os.X_OK) else 0o644
     if ids is None:
       ids = (0, 0)
     if names is None:
       names = ('', '')
-    # Note: normpath changes / to \ on windows, but we expect / style paths.
-    dest = os.path.normpath(dest).replace(os.path.sep, '/')
```
jylinv0, line 78: Should we apply the same refactoring to other methods?

pkg/private/tar/build_tar.py
```diff
@@ -64,18 +77,14 @@ def add_file(self, f, destfile, mode=None, ids=None, names=None):
        names: (username, groupname) for the file to set ownership. `f` will be
          copied to `self.directory/destfile` in the layer.
     """
-    dest = destfile.lstrip('/')  # Remove leading slashes
-    if self.directory and self.directory != '/':
-      dest = self.directory.lstrip('/') + '/' + dest
+    dest = self.normalize_path(destfile)
```
jylinv0, line 80: Do we remain the same behavior if we change `dest` from related path to possibly absolute path?

pkg/private/tar/build_tar.py
aiuto: I don't understand.  How would a 'for' help here?

pkg/private/tar/build_tar.py
aiuto: The already do the same thing. But I extracted this idiom into a helper.

pkg/private/tar/build_tar.py
aiuto: Yes, this is fine.   'dest' is just the path we put in the tarfile. It should be relative all the time, but someone
might want to set the package prefix to something absolute like '/etc'. That will still give a valid tar file.

Of course, doing that is almost always a bad idea. All reasonable implementations will strip the leading / on untar.

pkg/private/tar/build_tar.py
jylinv0: Just being curious why we use 'for' to strip prepending `./` in the util functions but not here.

jylinv0 (approved):

pkg/private/tar/build_tar.py
aiuto: It's just a question of do we have multiple things or not.
Here we have one path, so there is no need to iterate over a set.

PR #558, opened 2022-03-09
Move artifact tree handling to process_label_list.

aiuto:
- Use it from pkg_zip
- Move processing of files from DefaultInfo to process src
  - eventually we could move tree handling from process_label_list to
    process_src
- Add more docstrings.

This is a precursor to adding executable and runfiles handling to all rules.

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -64,6 +64,7 @@ _DestFile = provider(
 
 def _check_dest(content_map, dest, src, origin):
     old_entry = content_map.get(dest)
+
     # TODO(#385): This is insufficient but good enough for now. We should
     # compare over all the attributes too. That will detect problems where
     # people specify the owner in one place, but another overly broad glob
@@ -87,7 +88,7 @@ def _merge_attributes(info, mode, user, group):
 def _process_pkg_dirs(content_map, pkg_dirs_info, origin, default_mode, default_user, default_group):
     attrs = _merge_attributes(pkg_dirs_info, default_mode, default_user, default_group)
     for dir in pkg_dirs_info.dirs:
-        dest = dir.strip('/')
+        dest = dir.strip("/")
         _check_dest(content_map, dest, None, origin)
         content_map[dest] = _DestFile(
             src = None,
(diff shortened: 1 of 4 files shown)


pkg/private/pkg_files.bzl
```diff
@@ -133,11 +134,14 @@ def _process_pkg_filegroup(content_map, pkg_filegroup_info, origin, default_mode
...
         _process_pkg_symlink(content_map, psl[0], psl[1], default_mode, default_user, default_group)
 
-def process_src(content_map, src, origin, default_mode, default_user, default_group):
+def process_src(ctx, content_map, files, src, origin, default_mode,
+                default_user, default_group):
     """Add an entry to the content map.
 
     Args:
-      content_map: The content map
+      ctx: rule context
```
nacl, line 142: `ctx` appears unused.  Consider removing it, or documenting this fact.

pkg/private/zip/zip.bzl
```diff
@@ -39,6 +37,7 @@ def _pkg_zip_impl(ctx):
     args.add("-t", ctx.attr.timestamp)
     args.add("-m", ctx.attr.mode)
     inputs = []
+    file_deps = []
```
nacl, line 40: Perhaps declare this adjacent to where it is used, perhaps on line 49?

nacl (commented):
Makes sense.  See below for nits.

Please mention that `buildifier` made changes to files in your commit message.

nacl (approved):
Meant to approve initially.  Please do take a look at my comments.

It might also be worth adding test cases for this functionality if they aren't covered elsewhere.

pkg/private/pkg_files.bzl
aiuto: That was left over from another idea.  Removed.

pkg/private/zip/zip.bzl
aiuto: Moved.

aiuto:
> Meant to approve initially. Please do take a look at my comments.

Addressed them.

> It might also be worth adding test cases for this functionality if they aren't covered elsewhere.

Most are covered, but indirectly through integration tests.  If I can find a free day, I can write some comprehensive tests that test the resulting manifest. OTOH, I'm not sure if we want to do that today, or after we change the manifest format. Doing it later could make the fail output from the test much easier to understand.

PR #566, opened 2022-04-03
Make bzl_library public for Stardoc

### LINKED ISSUE #565: Increase visibility of all .bzl files to public.
It's unfortunate but the concept of `private` breaks other tools (eg. Stardoc).
https://github.com/bazelbuild/stardoc/issues/93

guw:
Workaround for https://github.com/bazelbuild/stardoc/issues/93.
Fixes #565

### DIFF
--- a/doc_build/BUILD
+++ b/doc_build/BUILD
@@ -118,6 +118,7 @@ bzl_library(
         "//pkg:mappings.bzl",
         "//pkg:package_variables.bzl",
         "//pkg:path.bzl",
+        "//pkg:pkg.bzl",
         "//pkg:providers.bzl",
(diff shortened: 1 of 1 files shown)


aiuto:
I think this will be subsumed by #567.  That does what I think you want in a very different way.  It will also work from the releases, by adding  //pkg:bzl_srcs as a usable bzl_library input.

Can you take a look at that one?

aiuto:
Hi Gunner. Could you take a lot at #567? I think it solves your need, but in a way that also works for people using the release build instead of snapshot.

aiuto:
I'm going to presume this is solved by #567.

PR #567, opened 2022-04-04
Make the .bzl files available as input to a bzl_library

### LINKED ISSUE #565: Increase visibility of all .bzl files to public.

aiuto:
Make the .bzl files available as input to a bzl_library, without having to resort to the invasivness of a shadow tree of just the .bzl files without the other distributable files.

Fixed #565

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -14,6 +14,7 @@
 
 # -*- coding: utf-8 -*-
 load("@rules_python//python:defs.bzl", "py_binary", "py_library")
+load("@rules_pkg//pkg/private:make_starlark_library.bzl", "starlark_library")
 
 package(default_applicable_licenses = ["//:license"])
 
@@ -42,7 +43,23 @@ filegroup(
         "//pkg/legacy:standard_package",
         "//pkg/rpm:standard_package",
     ],
-    visibility = ["//distro:__pkg__"],
+    visibility = [
+        "//distro:__pkg__",
+        "//pkg:__pkg__",
+    ],
+)
+
+starlark_library(
+    name = "bzl_srcs",
+    srcs = [
+        ":standard_package",
+        "//pkg/private:standard_package",
+        "//pkg/private/deb:standard_package",
+        "//pkg/private/tar:standard_package",
+        "//pkg/private/zip:standard_package",
+        "//pkg/releasing:standard_package",
+    ],
+    visibility = ["//visibility:public"],
 )
 
 # Used by pkg_rpm in rpm.bzl.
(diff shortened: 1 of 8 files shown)


achew22:
Out of curiosity, why is a shadow tree so onerous? It is how every other language that I'm aware of in the Bazel ecosystem works. Can you help me understand why this is an improvement over the "normal" model of having an `x_library` with `deps` on other `x_library`s?

pkg/BUILD
```diff
@@ -42,7 +43,23 @@ filegroup(
...
+
+starlark_library(
+    name = "bzl_srcs",
+    srcs = [
+        ":standard_package",
+        "//pkg/private:standard_package",
+        "//pkg/private/deb:standard_package",
+        "//pkg/private/tar:standard_package",
+        "//pkg/private/zip:standard_package",
+        "//pkg/releasing:standard_package",
```
guw, line 60: This needs `"//pkg:pkg.bzl"`

pkg/private/make_starlark_library.bzl
```diff
@@ -0,0 +1,35 @@
...
+def _make_starlark_library(ctx):
+    direct = []
+    transitive = []
+    for src in ctx.attr.srcs:
+        if StarlarkLibraryInfo in src:
+            transitive.append(src[StarlarkLibraryInfo])
+        else:
+            for file in src[DefaultInfo].files.to_list():
+                if file.path.endswith(".bzl"):
+                    print(file.path)
```
duarten, line 17: Leftover debug statement?

pkg/BUILD
aiuto: Added.

pkg/private/make_starlark_library.bzl
aiuto: Thanks for catching that.

aiuto:
> Out of curiosity, why is a shadow tree so onerous? It is how every other language that I'm aware of in the Bazel ecosystem works. 

There are a few things I don't like about it.
- We're not actually happy with the pattern within the Bazel code itself. We're trying to get away from that a bit with the recently added `subpackages` method. 
- I have to do it twice, once for the sources (including the python) to go in the distribution, and then again for just the .bzl files.  The point of this CL is to have one human written mechanism to gather the distribution tree and then a mechanical way to unsure the .bzl files are available.
- the need for the .bzl files is because of the current implementation of stardoc, not for a conceptual reason. The design we would like to move to is one where bazel directly emits the doc information rather than our standalone parsing. That would change the flavor of the stardoc rule to require only the file you want emitted.

PR #570, opened 2022-04-05
Allow pkg_tar to have a package_dir == './'

### LINKED ISSUE #50: make pkg_tar's "./" mandatory prefix optional

aiuto:
We still never create a lone '.' as a directory in the archive. That is
intentional

There are a few parts to this:

- Remove the root_directory logic from tar_writer. It was a needless feature
  inherited from an ancient implementation. It was not used by anything except
  the tests.
- Stop re-normalizing paths in tar_writer. It should do what it is asked.  The
  implication is that all the test which accounted for test data which which
  contained tar files with paths like './a', will now keep the './'.
- Add a test to show prefix_dir == './' does work.

RELNOTES: pkg_tar no longer prefixes paths with './'.

Fixes #50

### DIFF
--- a/distro/packaging_test.py
+++ b/distro/packaging_test.py
@@ -81,7 +81,7 @@ class PackagingTest(unittest.TestCase):
     # TODO(aiuto): Find tar in a disciplined way
     content = subprocess.check_output(
         ['tar', 'tzf', 'bazel-bin/dummy_tar.tar.gz'])
-    self.assertEqual(b'BUILD\n', content)
+    self.assertEqual(b'etc/\netc/BUILD\n', content)
 
 
 if __name__ == '__main__':
--- a/distro/testdata/BUILD.tpl
+++ b/distro/testdata/BUILD.tpl
@@ -7,7 +7,7 @@ pkg_tar(
     ],
     extension = "tar.gz",
     owner = "0.0",
-    package_dir = ".",
+    package_dir = "etc",
     tags = [
         "manual",
     ],
--- a/pkg/private/tar/tar_writer.py
+++ b/pkg/private/tar/tar_writer.py
@@ -46,7 +46,6 @@ class TarFileWriter(object):
                name,
                compression='',
                compressor='',
-               root_directory='',
                default_mtime=None,
                preserve_tar_mtimes=True):
     """TarFileWriter wraps tarfile.open().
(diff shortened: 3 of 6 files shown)


jylinv0 (approved):
nit: I'm sure we have a test to prove we reject/ignore input file `.`. But I can't find it, could you point it out?

PR #576, opened 2022-04-12
Support for setting executable bits.

### LINKED ISSUE #96: pkg_zip drops execute attribute when zipping executable file

aiuto:
- mostly fix #96 by finding executables and setting mode==755 on them

This new feature can not detect all executables. Bazel does not have the right capability to make that easy. This seems to get most binaries, except shell. You will have to wrap those in pkg_files to set the mode bits.

This also does a little refactoring of the code for readability.

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -277,8 +277,6 @@ def add_label_list(
             default_group = default_group,
         ):
             # Add in the files of srcs which are not pkg_* types
-            # We can 
-
             add_from_default_info(
                 content_map,
                 file_deps,
@@ -318,7 +316,6 @@ def add_from_default_info(
 
     # Auto-detect the executable so we can set its mode.
     the_executable = get_my_executable(src)
-    file_deps.append(src[DefaultInfo].files)
     all_files = src[DefaultInfo].files.to_list()
     for f in all_files:
         d_path = dest_path(f, data_path, data_path_without_prefix)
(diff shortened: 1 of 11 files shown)


pkg/private/pkg_files.bzl
```diff
@@ -277,30 +277,103 @@ def add_label_list(
...
-                    add_single_file(
-                        content_map,
-                        d_path,
-                        f,
-                        origin = src.label,
-                        mode = default_mode,
-                        user = default_user,
-                        group = default_group,
-                    )
+            # We can
```
jylinv0, line 280: Unfinished comment?

pkg/private/pkg_files.bzl
```diff
@@ -277,30 +277,103 @@ def add_label_list(
...
+      default_mode: fallback mode to use for Package*Info elements without mode
+      default_user: fallback user to use for Package*Info elements without user
+      default_group: fallback mode to use for Package*Info elements without group
+    """
+    if not DefaultInfo in src:
+        return
+
+    # Auto-detect the executable so we can set its mode.
+    the_executable = get_my_executable(src)
+    file_deps.append(src[DefaultInfo].files)
```
jylinv0, line 321: It's duplicating L156.

pkg/private/pkg_files.bzl
```diff
@@ -277,30 +277,103 @@ def add_label_list(
...
+    Inputs:
+      src: A label.
+    Returns:
+      File or None.
+    """
+    the_executable = None  # The file which is the actual executable.
+    if DefaultInfo in src:
+        di = src[DefaultInfo]
+        # XXX print(ctx.label, dir(di))
+        if hasattr(di, "files_to_run"):
```
jylinv0, line 364: Maybe we can concatenate all nested if with `and`?

jylinv0 (commented):
Does this mean if I want to explicitly set the mode for a file, which was seems as executable, then I won't be able to?

pkg/private/pkg_files.bzl
aiuto: It would be come an unwieldly line. It would end up looking like
```
if DefaultInfo in src and hasattr(src[DefaultInfo], "files_to_run") and hasattr(src[DefaultInfo].files_to_run) .....
````

**Comment by aiuto** [comment]
> Does this mean if I want to explicitly set the mode for a file, which was seems as executable, then I won't be able to?
No.  This change only takes effect for files which are added because of DefaultInfo.  Files which are added from pkg_files take a path where the mode is set in that rule.

File: pkg/private/pkg_files.bzl [hunk 4]
```diff
@@ -277,30 +277,103 @@ def add_label_list(
...
+
+    Inputs:
+      src: A label.
+    Returns:
+      File or None.
+    """
+    the_executable = None  # The file which is the actual executable.
+    if DefaultInfo in src:
+        di = src[DefaultInfo]
+        # XXX print(ctx.label, dir(di))
```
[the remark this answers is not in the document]
aiuto: Done

pkg/private/pkg_files.bzl
aiuto: Good catch. Yes.

pkg/private/pkg_files.bzl
aiuto: Removed

aiuto:
Review comments addressed. Fixed the tests for windows.
Bazel emits the executables on Windows as foo.exe instead of foo. This means any tests that look for expected file names have to take different data sets.

jylinv0 (approved):

PR #577, opened 2022-04-21
Substitute package variables in `pkg_zip#package_dir`.

dlj-NaN:
This change adds logic and tests for substituting values from `pkg_zip#package_variables` into `pkg_zip#package_dir`. This behavior largely matches `pkg_tar` (`pkg_zip` does not have an equivalent to `pkg_tar#package_dir_file`, but they are otherwise the same).


dlj-NaN:
> I would have liked to see an issue for this first, but it seems on the right track. I can look next week.

Sure, filed #578.

aiuto (approved):

PR #581, opened 2022-05-12
try to skin the docs

aiuto:


aiuto:
Force merge in docs three to see if we can to private CSS.

PR #591, opened 2022-05-21
Use direct label for stamp detect

aiuto:


jylinv0 (approved):
It's undocumented, but select seems accepts string and label as keys.

aiuto:
> It's undocumented, but select seems accepts string and label as keys.

In this case, we need the label to resolve once to @rules_pkg//pkg/private:private_stamp_detect,
and then be the same in macro expansion later.
With the str() around it, when pkg_tar macro was expanding it was seeing "//pkg_private" and then reinterpreting that w.r.t. their local repository rather than @rules_pkg.

Issue #522, opened 2022-02-03
pkg_deb mishandles the description field


aiuto:
- It re-flows it, so any formatting you had is lost.
- It does not handled the Description in the changes file correctly

That is just broken. It must be a holdover choice from the original google implementation.

aiuto:
According to https://www.debian.org/doc/debian-policy/ch-controlfields.html#syntax-of-control-files, first line in a multiline field could be empty, e.g. https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-changes and https://www.debian.org/doc/debian-policy/ch-controlfields.html#files. There could be blank lines between paragraphs, represented as a whitespce followed by a period.

And more than one proceeding whitespaces have special meaning: https://www.debian.org/doc/debian-policy/ch-controlfields.html#description, so you may want to preserve the leading whitespaces from the input.

PR #595, opened 2022-06-30
Keep original line break for multiline debian fields

bozaro:
I want to keep original line breakings in multiline description.
Currently all paragraphs are joined by space and rewriped as single paragraph :(

### DIFF
--- a/pkg/private/deb/make_deb.py
+++ b/pkg/private/deb/make_deb.py
@@ -125,10 +125,14 @@ def MakeDebianControlField(name, value, wrap=False):
   if isinstance(value, list):
     value = u', '.join(value)
   if wrap:
-    result += u' '.join(value.split('\n'))
-    result = textwrap.fill(result,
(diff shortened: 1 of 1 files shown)


jylinv0:
> There is already an issue related to this. Can you verify that and this PR are not in conflict
> […](#)

I didn't find the issue, but I remember talking about one of desired feature to support multiline control field. There's one issue that's related to description field: https://github.com/bazelbuild/rules_pkg/issues/522, and probably partially related to this.

aiuto:
I spent this afternoon looking at this and came to the conclusion that both the original code and this PR are both doing it wrong.  The spec is subtle about it, but it it is up to the displaying tool to do the line wrapping. Not the tool which creates the  control field. https://www.debian.org/doc/debian-policy/ch-controlfields.html#description

So, I'm going to close this and create a new PR which does no description wrapping.

PR #601, opened 2022-07-19
Explicitly set the tar format to GNU.

### LINKED ISSUE #216: Python 3.8 changed default tar format from GNU to PAX. Respond appropriately.

aiuto:
This means that:
1. you'll get the same format no matter what version of Python you have.
   At Python 3.8 the default changed from GNU to PAX.
2. The default will be suitable for building Debian packages containing
   long file names.

A followup PR may add the capability to allow PAX tar writing, but I
do not know the urgency of that at this time, so that is a future
feature request.

Fixes #216

### DIFF
--- a/pkg/private/tar/tar_writer.py
+++ b/pkg/private/tar/tar_writer.py
@@ -97,7 +97,8 @@ class TarFileWriter(object):
       self.fileobj = self.compressor_proc.stdin
     self.name = name
 
-    self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj)
+    self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj,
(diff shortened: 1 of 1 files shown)


nacl (approved):
It probably be worth adding in some testers that use the newer version of python so we can confirm that everything is happy there.  Ubuntu 20.04 may be enough here.

aiuto:
I started a set of tests but it became meaningless quickly because it was not clear what my goals were. What I think I need is the kind of long file name that would be different in each of the tar formats, and then to see that they make valid .deb files.

Issue #153, opened 2015-12-02
Fix and document pkg_tar() runfiles support.


chin33z:
When build with --dynamic_mode=fully, pkg_tar() only tars the "files" into the tarball. It does not include the shared library(located in the runfile tree) it needs. It would be nice if pkg_tar packing all the resources it needs. Thanks.

aiuto:
I have been doing a sprint on supporting pkg_files* in pkg_tar.  I think I have a design I like. I wanted to stick it out here for comment.  

We start with some requirements
- `*_binary `rules must not change in any way
- Distributions might want to place a binary in a different relative path from the root than the package path in your workspace.  For example.   `//my/project/src:footool` might be installed to /usr/bin/footool
- Distributions might want data and/or binary runfiles in specific places
  - data files might be best in /usr/share/footool/...
  - binary runfiles (.so or .py) might be in /usr/bin/footool.runfiles/... or maybe /usr/lib/footool/runfiles/...
- The policy of where files might be mapped belongs to the packager, so we want to specify it somewhere else than in the `*_binary` rule
- A single `pkg_tar` (or `pkg_*`) target might bundle multiple binaries each with different choices for relocating runfiles. So, we want to declare data and runfile remapping outside the `pkg_tar` rule to avoid clutter.

So, the idea is you would do this
```
cc_binary(name="foo", ..., data=[":baz"])

pkg_files(
   name='foo_bin',
   srcs = [':foo'],
   strip_prefix=strip_prefix.from_root(),   # removes package path from ultimate path of foo
   prefix = '/usr/bin'                                     # makes it appear as /usr/bin/foo
   runfiles_prefix = '/usr/share/foo',           # makes runfiles appear as a tree under /usr/share/foo
)

pkg_zip(name = 'foozip', srcs=['foo_bin'])
```
What this is missing though, is the way to know that some runfiles are noarch and some are not. That is, we don't have a good way to automatically know what should be in /usr/share/foo vs. /usr/lib/foo.  I started a separate discussion thread on bazel-discuss to talk about that.

Extended example of the progress so far: https://github.com/aiuto/rules_pkg/tree/rich/examples/rich_structure

limdor:
> I have been doing a sprint on supporting pkg_files* in pkg_tar. I think I have a design I like. I wanted to stick it out here for comment.

[...]

PR #605, opened 2022-07-26
First cut at runfiles support in pkg_* rules

aiuto:
Add include_runfiles support to the manifest builder used by pkg_tar and pkg_zip.  This pulls in runfiles in the same layout as bazel does. This is not a great layout, but it is understandable.  For example, an executable with a `data` dep might look like
```
foo/an_exectuable
foo/an_executable.runfiles/foo/an_excutable
foo/an_executable.runfiles/foo/data.txt
```
Note the duplication of an_executable.  That could be expensive, so we will eventually want a way to suppress that.

Next steps:
- Add e2e tests for `pkg_tar` and `pkg_zip`.
- Add `include_runfiles` to `pkg_files`. 

See #153  for overall design and features to be implemented.


depthwise:
So I've tried this with the most recent rules_python to package a Python binary runfiles, and here's the problem I'm encountering: there's `DefaultInfo.default_runfiles.root_symlinks` that rules_pkg doesn't seem take into account. This breaks `py_binary` in 2 ways:
  1. `bazel_tools/tools/python/py3wrapper.sh` is missing (although it is present under external, just not directly at root)
  2. Python bootstrap expects the various external PIP modules to be in runfiles root (they are symlinked when building in bazel), but because these are, well, `root_symlinks`, they're missing too.

So here's a suggestion: perhaps this should take `root_symlinks` into account? I know this property is undocumented, but it is very much a thing.

aiuto:
That's going to have to be a second pass.
When we start doing symlinks instead of files, it's not obvious what is
correct:
- create symlink and hope the file is there
- copy the file to the expected place and create the link
- copy the target to the location of the symlink and make it a real file.

The first choice is probably wrong.
If the second choice, what if the expected place is outside the workspace?
The third one seems right, and yet still wrong.

It's going to take some poking about to cover all the issues.

pkg/private/pkg_files.bzl
```diff
@@ -259,6 +259,10 @@ def add_label_list(
       default_group: fallback mode to use for Package*Info elements without group
     """
 
+    if hasattr(ctx.attr, "include_runfiles"):
+        include_runfiles = ctx.attr.include_runfiles
+    else:
+        include_runfiles = False
```
nacl, line 265: This could be shortened to

```starlark
include_runfiles = getattr(ctx, "include_runfiles", False)
```

nacl (approved):
LGTM all things considered, but I think the commit message heading could be made more specific and call out exactly what happened.

BTW, it looks like `pkg_zip` doesn't even support `include_runfiles` as an attribute, so that may need to be added in and tested.

pkg/private/pkg_files.bzl
```diff
@@ -332,14 +339,29 @@ def add_from_default_info(
...
-                f,
+                dest_path = d_path,
+                src = f,
                 origin = src.label,
                 mode = fmode,
                 user = default_user,
                 group = default_group,
             )
-
+    if include_runfiles:
```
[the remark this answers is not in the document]
aiuto: I want to deal with factoring it out later.
I expect that I don't know the exact needs yet.
I am wondering if pkg_files should use add_label_list and pass in a lambda which can do the remapping.

pkg/private/pkg_files.bzl
aiuto: I changed my mind.   That change actually gives a different result. The getattr with keeps yielding False, even when set.

PR #607, opened 2022-08-18
Support Bzlmod

tetsuok:
This PR adds a support [Bzlmod](https://bazel.build/build/bzlmod) to upstream MODULE.bazel file to this repo as per bazelbuild/bazel-central-registry#124. It seems there is an existing similar PR #485, but it seems there is no update since February. I'm totally fine to abandon this PR if #485 is better as long as rules_pkg supports Bzlmod.


aiuto (changes requested):
This should not be an example.

The PR that adds bzlmod support should 
- add the module
- add a new CI run that invovkes bazel with bzlmod enabled

So, pretty much 2 files.

tetsuok:
@aiuto Thanks for the review. Sure. I'll update this PR to add only two files.

tetsuok:
It seems like CI is failing with the [almost same error message](https://buildkite.com/bazel/rules-pkg/builds/1925#0182b51f-3a3c-484c-8476-666192060ab7) as #485.
Looking at https://github.com/bazelbuild/rules_pkg/pull/485#issuecomment-1039207335, no need to keep this PR, #485 is enough.

tetsuok:
Closing as commented above.

Issue #115, opened 2019-11-10
No option to preserve relative symlinks in pkg_tar


colatkinson:
The current behavior of `pkg_tar` is that if I have a file and a symlink to that file, and both are given as sources, the symlink is "expanded" and two copies of the original file are added to the repo.

For some context, my use case is that I'm trying to package an external toolchain's libc into a tar for use in Docker containers (cross-compiling is... fun). As you can imagine, this is massively ballooning the size of my images.

While I could certainly attempt to do something like filtering all the links out and then generating a `symlinks` dict to pass in, I was wondering if there would be interest in adding a flag to allow this behavior in the main rule. I think it could be useful in a variety of contexts, especially in generating library debs for use by others.

The behavior as I imagine it would be along the lines of passing a `preserve_symlinks` flag, which would then include symlinks *if* they only point to files in the archive. This validation should be simple enough, and would remove risks of non-hermeticity. Alternatively, symlinks pointing elsewhere could follow the current behavior, while hermetic symlinks could be preserved.

Please let me know if there are any concerns I'm missing, or if there's an obvious solution to my problem that I missed somewhere in the docs.

aiuto:
Clearly a missing feature. I would like to see this in pkgfilegroup and reused across all rules.

pudelkoM:
In case someone needs this and is fine with a workaround until this is properly fixed here, we wrote a Starlark wrapper that invokes `pkt_tar` with an appropriate flagfile to create the symlinks:

https://github.com/stratum/stratum/blob/3e8f0cdf1abdbdf5458b802058c8413084884420/bazel/rules/package_rule.bzl#L12-L47

aiuto:
This should be doable now with pkg_mklink, but it won't do the auto-magic expand.
I think that feature is not that useful.  In practice, what you usually want in your package is something like
- a binary that is the result of a rule, and you want to place that somewhere like /usr/share/MyApp/bin/myapp
- a symlink from /usr/bin/myapp to /usr/share/MyApp/bin/myapp

Bazel doesn't really deal with symlinks as a differentiable first-class file type that well, so it would be excessive engineering here to do that unless that was added.

aiuto:
[...]

PR #603, opened 2022-07-20
Create an example of the kind of link tree Node users must create

aiuto:
This just creates the tool to make the kind of test data that simulates an NPM treey.

See #115 for context.

PR #609, opened 2022-08-23
test: demonstrate symlink issue

thesayyn:
#115


aiuto:
Does this add new cases beyond what #603 added?

thesayyn:
I belive so. i couldn't see the use of `ctx.actions.symlink` `ctx.actions.declare_file()` in that PR. symlink info is given beforehand which is not something this PR tries to repro.

PR #613, opened 2022-09-02
Fix #612

### LINKED ISSUE #612: Directories in tree artifacts are written to tar files with mode 644 instead of 755
Intermediare directories in tree artifacts get written with the wrong permissions in pkg_tar.
They end up as 644 rather than 755, rendering them unreadable.

The fix might be as simple as: 
```
pkg/private/tar/build_tar.py
268c268
<             mode=f_mode,
---
>             mode=0o755,
```

aiuto:


nacl (approved):
This is the best default; LGTM.

nacl:
Also, would a unit test be appropriate here?

aiuto:
I do want to add a unit test. I was up far too late last night doing this, so I just made a quick fix and PR so that I could run the CI tests. I'll add something before merging.

PR #615, opened 2022-09-12
Do not warn if the origin paths are the same.

faximan:
The current conditions for printing build warnings are overly restrictive. While still not perfect, I believe this is a step in the right direction.


aiuto (approved):
As discussed in other mail, the detection is deep.
This is fine for now, because it is certainly less noisy than before.

PR #619, opened 2022-09-24
Bump dependencies

hanneskaeufler:
This bumps `rules_license`, `bazel_skylib`
as well as `rules_python` to their newest
versions as of today.

There is no particular reason for the bump other
than playing around with bzlmod and seeing them
being quite outdated.

### DIFF
--- a/pkg/deps.bzl
+++ b/pkg/deps.bzl
@@ -25,26 +25,27 @@ def rules_pkg_dependencies():
         http_archive,
         name = "bazel_skylib",
         urls = [
-            "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
-            "https://github.com/bazelbuild/bazel-skylib/releases/download/1.1.1/bazel-skylib-1.1.1.tar.gz",
+            "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.3.0/bazel-skylib-1.3.0.tar.gz",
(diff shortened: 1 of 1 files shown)


aiuto:
It would be more useful to try to come up with a MODULE.bazel file and add the right tests to .bazelci/tests.yaml

In general, I'm not really inclined to update to the newest unless I need a feature from it.  If my dependencies are following semantic versioning correctly, then we should be able to use the same major versions of deps for years. That is the ideal case for LTS users.  If someone wants a newer version of rules_pkg, that should not change any of the other dependencies they have used, unless rules_pkg really needs a newer one.  Of course, if they load rules_pkg before rule_foo, and they update rules_foo, but that needs a newer rules_python, then they still get the old dep that rules_pkg wanted.  Which basically is a proof by example that the deps.bzl scheme doesn't work.

We're trying to fix that with bzlmod, but that's not going to help the people who will use 4.x and 5.x for the next few years.  Let me think about this more

aiuto (commented):
I'm going to bring this up in the next eng meeting.
I want to codify a policy on NOT updating until we have to.
That works better for LTS users.  So, on hold for now

hanneskaeufler:
Thanks for the updates. I can of course understand such a policy, its just that I didnt expect it😄 Feel free to close if you come to any such conclusion.Am 20.10.2022 um 17:57 schrieb aiuto ***@***.***>:﻿
@aiuto commented on this pull request.

I'm going to bring this up in the next eng meeting.
I want to codify a policy on NOT updating until we have to.
That works better for LTS users.  So, on hold for now

aiuto:
I'm going to close this out for the reasons I outlined earlier. 
The last few days has reminded me that one person updating a low level dep can ripple along through many repositories.  This causes churn in other projects. I'll update the contrib guidelines.

PR #620, opened 2022-10-07
Allow $(var) substitution in filenames

### LINKED ISSUE #520: package variables should support $(var) expansion for backwards compatibility to MakeVars
As ugly as the bazel make variables hacks are, it turns out that converting a large code base to a different style is a huge time waster.  We should support $(var) expansion in package_file_name in addition to {var}.

Also, while doing this, we should think about being able to get ctx.var expansion without having to make a package variables just to forward it.

aiuto:
Allow $(var) substitution in filenames and include everything in ctx.var in the substitution dictionary.

Fixes #520

### DIFF
--- a/pkg/private/util.bzl
+++ b/pkg/private/util.bzl
@@ -78,12 +78,12 @@ def substitute_package_variables(ctx, attribute_value):
 
     if type(attribute_value) != "string":
         fail("attempt to substitute package_variables in the attribute value %s which is not a string" % attribute_value)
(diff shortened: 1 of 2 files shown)


aiuto:
This seems like an unimportant thing to do right now, but it unblocked moving Google over to pkg_deb from a legacy tool. That will increase the likelihood that I can get other people to work on the tools.

buildbreaker2021 (approved):

PR #621, opened 2022-10-10
Update docs to show $(var) usage and using ctx.var directly.

aiuto:
I should have done this in pr #620


aiuto:
Friendly ping.

buildbreaker2021 (approved):

PR #637, opened 2022-11-22
Apply package_dir to pkg_mklink in pkg_tar

### LINKED ISSUE #610: package_dir attribute is not applied to pkg_mklink
I'm upgrading from rules_pkg 0.5.2 to rules_pkg 0.7.0 and trying to use the new "package_dir" parameter to add the "./" prefix that was removed as part of (https://github.com/bazelbuild/rules_pkg/issues/531).
But it seems that "package_dir" does not work with pkg_mklink.

The following code illustrates the problem:
```
load("@rules_pkg//:pkg.bzl", "pkg_tar")
load("@rules_pkg//:mappings.bzl", "pkg_mklink", "pkg_files")

pkg_files(
    name = "myfile_pkg",
    srcs = ["myfile"],
)

pkg_mklink(
    name = "mylink_lnk",
    link_name = "mylink",
    target = "myfile",
)

pkg_tar(
    name = "myarchive",
    srcs = ["myfile_pkg", "mylink_lnk"],
    package_dir = "."
)
```
Invoking
```
bazel build myarhive.tar
tar tf bazel-bin/myarchive.tar
```
Gives:
```
./myfile
mylink
```
I would have expected:
```
./myfile
./mylink
```

aiuto:
- Fix brokenness in pkg_tar
- Add tests (it was working) to pkg_zip

Fixes #610

### DIFF
--- a/tests/tar/tar_writer_test.py
+++ b/tests/tar/tar_writer_test.py
@@ -217,6 +217,7 @@ class TarFileWriterTest(unittest.TestCase):
     expected_content = [
         {"name": "my"},
         {"name": "my/package"},
+        {"name": "my/package/mylink"},
         {"name": "my/package/nsswitch.conf"},
     ]
     self.assertTarFileContent(package_dir, expected_content)
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -176,9 +176,9 @@ class TarFile(object):
       names: (username, groupname) for the file to set ownership.  An empty
         file will be created as `destfile` in the layer.
     """
-    symlink = normpath(symlink)
+    dest = self.normalize_path(symlink)
     self.tarfile.add_file(
-        symlink,
+        dest,
         tarfile.SYMTYPE,
         link=destination,
         mode = mode,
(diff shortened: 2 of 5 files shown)


tests/tar/tar_writer_test.py
```diff
@@ -217,6 +217,7 @@ def testPackageDirAttribute(self):
     expected_content = [
         {"name": "my"},
         {"name": "my/package"},
+        {"name": "my/package/mylink"},
```
nacl, line 220: It seems that this behavior in some cases might necessitate a tight binding between what's in `package_dir` and the contents of `pkg_mklink` if a symlink needs to reference something relative to the base of the package.

Thoughts?  This could perhaps be fixed via a template.

nacl (approved):
LGTM, but I wonder how well this scales.  It probably works well enough in 99% of the use cases in any cases; see the comment below.

Also, on submit, please call out `pkg_tar` in the commit summary (first line).

tests/tar/tar_writer_test.py
aiuto: I think the case you have is  
- ` mk_link(foo -> bar)`
- `packgae_dir = baz`

Gives: ` baz/foo -> bar`, but we want `baz/foo => baz/bar`.

I'm going to call that a distinct feature request.  I can see use cases for both leaving the link relative (or absolute) and for re-rooting it.

PR #639, opened 2022-11-23
Remove unnecessary `to_list()` calls

fmeum:
Avoids flattening depsets during the analysis phase by passing depsets into `ctx.actions.run`'s `input` parameter.


aiuto:
Nice performance improvement. Thanks.

aiuto (approved):

PR #640, opened 2022-11-23
Explicitly store implicit parent directories in zip files

### LINKED ISSUE #622: Directory entries not added to zip contents
When creating a zip using `pkg_zip` with files nested in a directory, the zip only contains the files, and not the entries for the directories. This causes unpacking to fail with some "unzip" implementations.

`pkg_zip` should create all the directory entries required for a file before adding that file to the zip.

### LINKED ISSUE #638: ZIP files contain no directory entries
Zip files produced by `rules_pkg` do not contain directory entries. For example, they look like this:
```
META-INF/MANIFEST.MF
xds/one.xsd
xds/two.xst
```

Although technically not required by the spec, a convention is to add directory entries to zip files (0 bytes, file name ending with `/`). A lot of tools seem to follow/implement this convention. 

Expected:
```
META-INF/
META-INF/MANIFEST.MF
xds/
xds/one.xsd
xds/two.xst
```

Missing directory entries may cause issues with the Java ecosystem, which seems to heavily rely on the directory entries. For example, see this Spring Framework issue:
https://github.com/spring-projects/spring-framework/issues/28230#issuecomment-1324702762

fmeum:
Tooling in the Java world (and likely elsewhere) has come to depend on all directories implicilty present as parent directories of files to be listed explicitly as a member of a ZIP file.

Fixes #638
Fixes #622

### DIFF
--- a/pkg/private/zip/build_zip.py
+++ b/pkg/private/zip/build_zip.py
@@ -20,7 +20,6 @@ import os
 import zipfile
 
 from pkg.private import build_info
-from pkg.private import helpers
 from pkg.private import manifest
 
 ZIP_EPOCH = 315532800
@@ -98,12 +97,12 @@ class ZipWriter(object):
     self.zip_file.close()
     self.zip_file = None
 
-  def make_zipinfo(self, path: str, mode: int):
+  def make_zipinfo(self, path: str, mode):
     """Create a Zipinfo.
 
     Args:
       path: file path
-      mode: file mode
+      mode: file mode (int or str)
     """
     entry_info = zipfile.ZipInfo(filename=path, date_time=self.time_stamp)
     # See http://www.pkware.com/documents/casestudies/APPNOTE.TXT
@@ -115,25 +114,26 @@ class ZipWriter(object):
     # permission and file type bits, while the low order two contain MS-DOS FAT
     # file attributes.
     if mode:
-      f_mode = int(mode, 8)
+      if isinstance(mode, str):
+        f_mode = int(mode, 8)
+      else:
+        f_mode = mode
     else:
       f_mode = self.default_mode
     entry_info.external_attr = f_mode << 16
     return entry_info
 
-  def add_manifest_entry(self, options, entry):
+  def add_manifest_entry(self, entry):
     """Add an entry to the zip file.
 
     Args:
-      options: parsed options
       zip_file: ZipFile to write to
       entry: manifest entry
     """
     entry_type, dest, src, mode, user, group = entry
 
     # Use the pkg_tar mode/owner remaping as a fallback
-    non_abs_path = dest.strip('/')
-    dst_path = _combine_paths(options.directory, non_abs_path)
+    dst_path = dest.strip('/')
     if entry_type == manifest.ENTRY_IS_DIR and not dst_path.endswith('/'):
       dst_path += '/'
     entry_info = self.make_zipinfo(path=dst_path, mode=mode)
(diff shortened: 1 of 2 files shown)


fmeum:
FYI @guw

pkg/private/zip/build_zip.py
```diff
@@ -189,24 +189,47 @@ def add_tree(self, tree_top: str, destpath: str, mode: int):
...
-        f_mode = mode
-      entry_info = self.make_zipinfo(path=path, mode=f_mode)
-      entry_info.compress_type = zipfile.ZIP_DEFLATED
-      if not content_path:
-        self.zip_file.writestr(entry_info, '')
-      else:
+      if content_path:
+        # If mode is unspecified, derive the mode from the file's mode.
+        if mode is None:
+          f_mode = 0o755 if os.access(content_path, os.X_OK) else 0o644
```
fmeum, line 201: This code path isn't covered by tests, so it wasn't noticed that `make_zipinfo` claimed to take `mode` as an `int`, when in fact it only supported `str`. I am now supporting both.

pkg/private/zip/build_zip.py
```diff
@@ -189,24 +189,47 @@ def add_tree(self, tree_top: str, destpath: str, mode: int):
...
+        entry_info = self.make_zipinfo(path=dir_path, mode=0o755)
+        entry_info.compress_type = zipfile.ZIP_STORED
+        # Set directory bits
+        entry_info.external_attr |= (UNIX_DIR_BIT << 16) | MSDOS_DIR_BIT
+        self.zip_file.writestr(entry_info, '')
 
+def _load_manifest(args, manifest_fp):
+  manifest_map = {}
+  for entry in json.load(manifest_fp):
+    entry[1] = _combine_paths(args.directory, entry[1])
```
aiuto, line 222: You could make this a little tighter by passing directory alone, rather than the entire args block

pkg/private/zip/build_zip.py
```diff
@@ -189,24 +189,47 @@ def add_tree(self, tree_top: str, destpath: str, mode: int):
...
+    manifest_map[entry[1]] = entry
+  manifest_keys = list(manifest_map.keys())
+  # Add all parent directories of entries that have not been added explicitly.
+  for dest in manifest_keys:
+      parent = dest
+      for _ in range(dest.count("/")):
+        parent, _, _ = parent.rpartition("/")
+        if parent and parent not in manifest_map:
+            manifest_map[parent] = [manifest.ENTRY_IS_DIR, parent, None, "0o755", None, None]
```
aiuto, line 231: I think this fails if we had a manifest like
```
foo/bar/baz/bin
abc/def
foo/bar    # an explicit pkg_mkdirs and they wanted something like 711 on it.
```
Perhaps this loop should not add directly to manifest_map. Instead, add to a temp map
and then merge those into manifest_map if they are not in there.

aiuto (commented):
Thanks!    Just a few nits.

pkg/private/zip/build_zip.py
fmeum: I am only adding the directory if it isn't already contained in the updated map (see line 230) and thus only if isn't contained in the original manifest. How is that different from a temp map and merging that?

pkg/private/zip/build_zip.py
fmeum: Changed.

fmeum:
I am now using your approach to fix the `mode` argument type.

aiuto (approved):

Issue #302, opened 2021-02-24
Try to derive default file mode from input types


nacl:
The current mechanism for setting file modes in `pkg_files` is ad-hoc and requires human intervention to set permissions, even when it's relatively obvious what they're supposed to be: runnable binaries should be runnable, libraries shouldn't.

Investigate mechanisms for determining how we might derive this information.  The most likely source of this information is the provider interfaces.

---

That's a good point, however I don't think we're quite there.  We'd have to do some introspection magic in `pkg_files` to identify the nature of the dependent rules.  Providers?  Rule names?  This will need to be investigated.  At first brush, this feels like it'll be hard to get right.

I feel that this can be postponed to a later task.  I'll make a ticket for it.

_Originally posted by @nacl in https://github.com/bazelbuild/rules_pkg/pull/290#discussion_r582332072_

aiuto:
Providers sounds like the right thing. One simple thing might be to look at srcs for `DefaultInfo`.
If there is an `executable` element in `DefaultInfo` we know that file should be 0755.

nacl:
That looks to be `DefaultInfo.files_to_run.executable`: https://docs.bazel.build/versions/master/skylark/lib/FilesToRunProvider.html

PR #641, opened 2022-11-24
pkg_zip: Some unicode file handling fixes and basic tests

aiuto:
- add a very basic test for reading files with non-ASCII names
- fix processing of manifest in pkg_zip to account for utf-8 encoding with backwards compatibilty to python 3.6
- skips on windows because that is broken in ways I can't test easily right now.


pkg/private/manifest.py
```diff
@@ -49,9 +49,16 @@ def __repr__(self):
 
 def read_entries_from(fh):
     """Return a list of ManifestEntry's from `fh`"""
-    raw_entries = json.load(fh)
+    # Subtle: decode the content with read() rather than in json.load() because
+    # the load in older python releases (< 3.7?) does not know how to decode.
+    raw_entries = json.loads(fh.read())
```
nacl, line 54: Hah, fun fun.

pkg/private/manifest.py
```diff
@@ -49,9 +49,16 @@ def __repr__(self):
...
     return [ManifestEntry(**entry) for entry in raw_entries]
 
+def read_entries_from_file(manifest_path):
+    """Return a list of ManifestEntry's from the manifset file at `path`"""
```
nacl, line 58: Nit: Typo: `manifset` -> `manifest`

tests/BUILD
```diff
@@ -53,6 +52,22 @@ filegroup(
...
+# Data source for Unicode handling tests
+#
+pkg_files(
+    name = "utf8_files",
+    srcs = [
+        "//tests/testdata/utf8:files",
+    ],
+    # Note: This is temporary. We need to fix a latent bug where
+    # we are using 555 as the default for many things. That was the
+    # Google internal behavior.
```
nacl, line 65: Is there a task for this?

tests/zip/unicode_test.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2022 The Bazel Authors. All rights reserved.
```
nacl, line 1: 2023 now?

pkg/private/zip/build_zip.py
```diff
@@ -153,6 +155,7 @@ def add_manifest_entry(self, entry):
       entry_info.compress_type = zipfile.ZIP_STORED
       # Set directory bits
       entry_info.external_attr |= (UNIX_SYMLINK_BIT << 16)
+      src = src.encode('utf-8')
```
nacl, line 158: Is this needed?  This seems to be done implicitly as of python 3.6 (https://docs.python.org/3.6/library/zipfile.html#zipfile.ZipFile.writestr).

I think we need to document somewhere what our minimum supported version of python is.  CentOS 7 and Ubuntu 18.04 LTS both have 3.6, so perhaps we should standardize on that?

pkg/private/zip/build_zip.py
```diff
@@ -142,6 +142,8 @@ def add_manifest_entry(self, entry):
 
     if entry_type == manifest.ENTRY_IS_FILE:
       entry_info.compress_type = zipfile.ZIP_DEFLATED
+      # Using utf-8 for the file names is for python <3.7 compatibility.
+      src = src.encode('utf-8')
```
nacl, line 146: It might be more concise to do this directly in the context manager below instead, but this is fine.

pkg/private/zip/build_zip.py
```diff
@@ -142,6 +142,8 @@ def add_manifest_entry(self, entry):
...
       with open(src, 'rb') as src:
```
nacl, line 147: Avoid aliasing and rename the second `src`?

tests/zip/zip_test_lib.py
```diff
@@ -91,4 +91,5 @@ def assertZipFileContent(self, zip_file, content):
             # legacy rule implementation.
             attr = 0o555
           self.assertEqual(oct((info.external_attr >> 16) & UNIX_RWX_BITS),
-                           oct(attr))
+                           oct(attr),
+                           info.filename)
```
nacl, line 95: Nit: Maybe call out the keyword arg `msg` here?

nacl (approved):
LGTM other than some nitty things below

pkg/private/manifest.py
aiuto: This took a lot of head scratching to discover.

pkg/private/manifest.py
aiuto: fixed

tests/BUILD
aiuto: Pointed to https://github.com/bazelbuild/rules_pkg/issues/302

pkg/private/zip/build_zip.py
aiuto: Added 3.6 as a requirement to developers.md.

pkg/private/zip/build_zip.py
aiuto: Done

tests/zip/zip_test_lib.py
aiuto: Done

pkg/private/zip/build_zip.py
aiuto: Done.  It avoids 3 different meanings for `src`

PR #643, opened 2022-11-29
feat: expose tar manifest as an output

alexeagle:
This allows targets to run something like jq over the manifest contents

### DIFF
--- a/pkg/private/tar/tar.bzl
+++ b/pkg/private/tar/tar.bzl
@@ -228,6 +228,10 @@ def _pkg_tar_impl(ctx):
             files = depset([output_file]),
             runfiles = ctx.runfiles(files = outputs),
         ),
+        # NB: this is not a committed public API.
+        # The format of this file is subject to change without notice,
+        # or this OutputGroup might be totally removed.
+        # Depend on it at your own risk!
         OutputGroupInfo(
             manifest = [manifest_file], 
         ),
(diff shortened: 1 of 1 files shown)


aiuto:
This will only lead to disappointment.  @nacl is about to change the format to JSON.
In a few months, we may change it to textproto.
It's an internal thing.

alexeagle:
Well in the last two weeks I've already started depending on it. If and when you change the format, I'll adapt.

Would you accept the PR if it was documented as Internal Only so you're not obligated to support it?

pkg/private/tar/tar.bzl
```diff
@@ -225,6 +225,9 @@ def _pkg_tar_impl(ctx):
             files = depset([output_file]),
             runfiles = ctx.runfiles(files = outputs),
         ),
+        OutputGroupInfo(
```
aiuto, line 228: As you suggested, can you leave a comment that the specification is subject to change in any minor release.

aiuto:
Sure. By 1.0 it will be fairly stable, but I expect some churn in the next
few months.

Of course, people wanting to get at the intermediate form leads me to a
question.
Should there be a new generic packaging type here, or are their various
language
specific packaging types that could benefit from a declared interface from
pkg_* rules
manifests.

As I write that, this brings up another question.  Do you need to get the
manifest
from pkg_tar specifically, or would a rule to simply produce the manifest
be the
thing you really want?

Gah.... now the feature creep is hitting me.  Should we be able to
interject a
manifest builder, so we can create a MANIFEST file that is a member of a
.zip or .jar file as it is created.

pkg/private/tar/tar.bzl
aiuto: Friendly ping.

alexeagle: Thanks for the ping, done.

aiuto (approved):

PR #647, opened 2022-12-07
Fix pkg_tar directory entries

jameshilliard:
Directories appear to work for zip files but not tar files.

This should ensure we don't try to process directories as normal files.

Fixes issues like:
```python
rules_pkg/pkg/private/tar/tar_writer.py", line 242, in add_file
    with open(file_content, 'rb') as f:
IsADirectoryError: [Errno 21] Is a directory: 'bazel-out/darwin-fastbuild/bin/py/selenium/webdriver/common/devtools/v106'
```


AutomatedTester:
This PR works well, I've got it in the Selenium tree for now.

nacl:
Could you provide test cases for this?

If this is supporting what I think it is (passing in directory labels that are not [TreeArtifacts](https://bazel.build/reference/glossary#tree-artifact)), I am hesitant to support this any further, as is [not a best practice](https://bazel.build/concepts/dependencies#using-labels-reference-directories).  @aiuto may have a differing opinion.

Please correct me if I'm wrong, but the relevant [code in the Selenium tree](https://github.com/SeleniumHQ/selenium/blob/d8b45c8c66eca37e8773769218f2319ae7dff995/py/BUILD.bazel#L260-L271) is a `genrule` that looks to be emitting a directory, which is unfortunately [unsound](https://bazel.build/reference/be/general#general-advice) (the documentation's words, not mine) and IIRC Bazel emits a warning saying something to this point.  This IMO should instead be a first-class rule that emits a TreeArtifact (via `ctx.actions.declare_directory`, if you didn't already know).

tests/tar/BUILD
```diff
@@ -60,6 +60,20 @@ genrule(
     cmd = "for i in $(OUTS); do echo 1 >$$i; done",
 )
 
+genrule(
+    name = "generate_dir_file",
+    outs = ["lib"],
+    cmd = "mkdir -p $@; echo 1 >$@/nsswitch.conf",
```
aiuto, line 66: This is not a case we should actively support. It doesn't work in remote build situations.

aiuto (commented):
This PR is trying to add a feature that will not work all the time.  If you are building a tree of files you can't explicitly name or enumerate, the only bazel supported way is to use ctx.actions.declare_directory to make a tree artifact.

tests/tar/BUILD
jameshilliard: Well the behavior is at least currently inconsistent with zip files so not sure what best option here is.

jameshilliard: Does this [ctx.actions.declare_directory](https://github.com/SeleniumHQ/selenium/pull/11392) based rule look correct?

aiuto:
Maybe, but that case should be working already.

tests/tar/BUILD
aiuto: That looks mostly right. It does a declare_directory, then passes the .path of that to some tool to fill in.

pkg/private/tar/build_tar.py
```diff
@@ -309,7 +309,10 @@ def add_manifest_entry(self, entry_list, file_attributes):
     elif entry.entry_type == manifest.ENTRY_IS_EMPTY_FILE:
       self.add_empty_file(entry.dest, **attrs)
     else:
-      self.add_file(entry.src, entry.dest, **attrs)
+      if os.path.isdir(entry.src):
```
aiuto, line 312: This is not the place we should fix it.
For correctness, it needs to be in pkg.bzl, where we test that the input file has `is_directory` set.  But we are already doing that.

PR #656, opened 2023-01-05
Make the default behavior of pkg_files#strip_prefix more intuitive; rename it to srcs_strip_prefix

### LINKED ISSUE #354: pkg_files(strip_prefix) behavior is confusing w.r.t. flattening a tree of files

nacl:
`pkg_files#strip_prefix` provides the capability to remove path segments from files prior to them entering the output package.  The default for this attribute is to flatten the outputs (e.g. remove all directory structure), which is potentially confusing to users and can cause odd problems.  What is generally more expected (and error-prone) is to preserve the package-relative path, as specified when you pass `strip_prefix.from_pkg()` to `pkg_files`,

This change implements this update to the default, and implements the following additional changes:

- `pkg_files#strip_prefix` is now `pkg_files#srcs_strip_prefix` -- making it more clear what `strip_prefix` is actually applying to
-  `strip_prefix#files_only` (which flattens the directory structure) is now `strip_prefix#flatten`.

Given that this is a breaking change, we have provided a migration script that users can run to implement this change in their trees.  I have partially tested this in an internal tree, and the outputs are identical.  This script uses [libcst](https://github.com/Instagram/LibCST) to process the BUILD files and implement transformations.  Updates to .bzl files were not attempted due to complexity requirements; users will have to apply them manually.

All BUILD changes were implemented by applying the script like so:

```
git ls-files '**/BUILD' | xargs bazel run //migration:strip_prefix --
```

`bazel run` the above target for more details.  It can be run as `@rules_pkg//migration:strip_prefix` if the appropriate WORKSPACE stuffs are added.

Fixes #354.  Supplants #492.

TODO: provide instructions for migrating without downloading rules_pkg -- copying and tweaking some of the stuff added to the WORKSPACE file should be enough.


pkg/mappings.bzl
```diff
@@ -337,19 +337,18 @@ pkg_files = rule(
             """,
             default = "",
         ),
-        "strip_prefix": attr.string(
-            doc = """What prefix of a file's path to discard prior to installation.
+        "srcs_strip_prefix": attr.string(
```
aiuto, line 340: Do we really have to do this rename?  It will force incompatibility between versions.
That will basically mean I'll have to fork this for Google to keep the old naming for a very long time.

aiuto:
This seems to be different from what I thought we agreed on, which was that strip prefix would strip from root or package relative, and flatten was a separate attribute to flatten paths after the stripping.

Also, quick side question.  does strip_prefix='.' still do the right thing, which is to strip the package and not flatten the rest.

tests/mappings/BUILD
```diff
@@ -82,6 +83,7 @@ pkg_files(
     srcs = [
         "mappings_test.bzl",
     ],
+srcs_strip_prefix = strip_prefix.flatten(),
```
nacl, line 86: Hm.  Need to fix indentation here.

nacl:
> This seems to be different from what I thought we agreed on, which was that strip prefix would strip from root or package relative, and flatten was a separate attribute to flatten paths after the stripping.

I think we were thinking different things when we came to an agreement on this in December.  Could you provide an example of what you intend the behavior to be with the separate `flatten` attribute ?  I tried reasoning about it a bit, and there doesn't seem to be much that couldn't be otherwise be accomplished by either creating a new BUILD file in a subdirectory, or by manually specifying `prefix`

- Stripping from `a`, `flatten=False`: `b/c 
 
> Also, quick side question. does strip_prefix='.' still do the right thing, which is to strip the package and not flatten the rest.

I believe so; that code hasn't changed.

PR #660, opened 2023-01-13
Genericize package manifest system and interface

nacl:
The current way that rules_pkg communicates with must packagers is using a manifest file, which is currently a JSON data structure based on a an array of arrays.  While generally readable, it looks strange, as it was to reduce Bazel resource usage (JSON strings in memory).  Further, our Python code is directly bound to this data structure format.

However, if we want to add more or change this, it becomes cumbersome on both the Starlark and Python sides. This change alleviates concerns generally by:

- Converting all manifests to a JSON "object" style, improving readability. Numerous "golden" tests were updated to support this.
- Replace the `collections.namedtuple`-based `ManifestEntry` object with one that is a little more flexible and type-safe.
- Providing a function (`read_entries_from`) that converts a file-like object into a list of `ManifestEntry`s, and replacing all JSON reading in packagers (`tar`, `zip`, `install`) and their tests with this function.  While this is theoretically usable by consumers of rules_pkg, it is not intended to be supported as such.

Other convenience factors or things addressed:

- `ManifestEntry.entry_type` is now just `ManifestEntry.type`
- Bazel 6 now stringifies repository-local labels with a preceding `@`, unlike prior versions.  Adapt to this in the manifest writer.

Future changes will extend this interface to allow for custom attributes to be passed from `pkg_files` and friends, which, among other things, will be necessary to more generically support `pkg_rpm`.

Provides some groundwork for (but doesn't resolve) #385.

### DIFF
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -11,9 +11,6 @@ lts: &lts
 rolling: &rolling
   bazel: rolling
 
-bazel5: &bazel5
-  bazel: 5.2.0
-
 #
 # Groups of tests, by platform
 #
@@ -112,7 +109,3 @@ tasks:
     name: lts_windows
     <<: *windows
     <<: *lts
-  bzl5_windows:
-    name: bzl5_windows
-    <<: *windows
-    <<: *bazel5
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -11,6 +11,9 @@ lts: &lts
 rolling: &rolling
   bazel: rolling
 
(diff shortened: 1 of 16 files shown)


aiuto (approved):
Sorry this took so long. It landed just before a crunch time.
I love PRs like this.  All sorts of plumbling is different, but the tests remain the same.

nacl:
> Sorry this took so long. It landed just before a crunch time. I love PRs like this. All sorts of plumbling is different, but the tests remain the same.

No worries, thanks for taking a look!

PR #668, opened 2023-02-10
Fix a potential TypeException caused by None type

sqybi:
`os.access(content_path, os.X_OK)` will raise a TypeException when `content_path` is None.
`f_mode` is only used when `content_path` is neither None nor empty, so I decided to move the assignment into this else-branch.

Thanks @mutongx for reporting this issue.

### DIFF
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -27,7 +27,7 @@ from pkg.private.tar import tar_writer
 
 def normpath(path):
   """Normalize a path to the format we need it.
-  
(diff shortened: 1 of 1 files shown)


aiuto:
Sigh.  The presence of this PR means that we don't have an adequate test case for this situation.
Do you have time to add one?

aiuto (approved):
I'll take the lack of response about adding a test as a "no".

sqybi:
Really busy these days. I will try to add some tests on the weekend.

PR #669, opened 2023-02-10
Add verify_archive rule to do e2e tests on built archives.

### LINKED ISSUE #644: Provide a tool for testing content of zip/tar files
The intent is that consumers of the rules should have a tool that facilitates writing non-brittle tests to validate they set things up the right way.  I have seen many cases of people writing test that do things like bit compares to golden files. These breaks too often.  Off the cuff concept model.

```
validate_archive_test(
    name = "my_test",
    target = ":foo.zip",   # or tar.  The rule should take either. Maybe .deb, rpm too.
    # must contain everything here, in any order, and can contain more
    contains = [
         "usr/local/sbin/abld",
         "usr/local/sbin/web_server",
         ".*/foo/bar.*.exe",     # yes, these are regexes.
    ],
    does_not_contain_regex = [   # none of these regexes must match any file
          ".*/MANIFEST",
     ],
     has_entries = ">=3",
     require_entries_sorted = True,   # But what does sorted mean?  What locale?
)
```

This may be a matter of improving our current test utilities and externalizing them.

aiuto:
We could do more, but this should be good enough to start.

Let's review this as is for the match, and then I will add zip support in a followup.

Fixes #644

### DIFF
--- a/pkg/BUILD
+++ b/pkg/BUILD
@@ -100,3 +100,13 @@ py_binary(
     python_version = "PY3",
     visibility = ["//visibility:public"],
 )
+
+# This might be public, but use at your own risk
+py_library(
+    name = "verify_archive_test_lib",
+    srcs = ["verify_archive_test_lib.py"],
+    srcs_version = "PY3",
+    visibility = ["//visibility:public"],
+)
+
+exports_files(["verify_archive_test_main.py.tpl"])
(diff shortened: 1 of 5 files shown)


sdtwigg (approved):
docstrings look fine and sufficiently self-explanatory

aiuto:
Thanks.

aiuto:
Actually, I need an approve review.

Issue #670, opened 2023-02-10
pkg_files and pkg_tar produce different tarballs.


AustinSchuh:
```
pkg_files(
    name = "test-pkg-files-with-attributes",
    srcs = [ 
        "//tests:testdata/loremipsum.txt",
    ],  
    attributes = pkg_attributes(
        group = "1000",
        mode = "0440",
        user = "0",
    ),
    prefix = "/foo/bar",
)

pkg_tar(
    name = "test-pkg-tar-from-pkg-files-with-attributes",
    srcs = [ 
        ":test-pkg-files-with-attributes",
    ],  
)

pkg_tar(
    name = "test-pkg-tar-with-attributes",
    srcs = [ 
        "//tests:testdata/loremipsum.txt",
    ],  
    owner = "0.1000",
    package_dir = "/foo/bar",
)

```

I would expect both of these to create the same .tar

```
austin[169160] aschuh-3950x (main) ~/local/rules_pkg/tests/tar
$ tar tvf ../../bazel-bin/tests/tar/test_pkg_tar_with_attributes.tar; tar tvf ../../bazel-bin/tests/tar/test_pkg_tar_from_pkg_files_with_attributes.tar
drwxr-xr-x 0/1000            0 1999-12-31 16:00 foo/
drwxr-xr-x 0/1000            0 1999-12-31 16:00 foo/bar/
-r--r----- 0/1000          543 1999-12-31 16:00 foo/bar/loremipsum.txt
drwxr-xr-x 0/1000            0 1999-12-31 16:00 foo/
drwxr-xr-x 0/1000            0 1999-12-31 16:00 foo/bar/
-r--r----- 0/1000          543 1999-12-31 16:00 foo/bar/loremipsum.txt
```

But, the sha256 doesn't match.

```
austin[169202] aschuh-3950x (main) ~/local/rules_pkg/tests/tar
$ sha256sum ../../bazel-bin/tests/tar/test-pkg-tar-with-attributes.tar ../../bazel-bin/tests/tar/test-pkg-tar-from-pkg-files-with-attributes.tar
3567e6a8e5286da1052853fa4b754d57638625a1574aae48ad493152d3e2ef95  ../../bazel-bin/tests/tar/test-pkg-tar-with-attributes.tar
[...]

PR #671, opened 2023-02-11
Add support for setting uid/gid from pkg_attributes

AustinSchuh:
This enables the user to produce identical tarballs to fix https://github.com/bazelbuild/rules_pkg/issues/670 by adding a uid/gid option to pkg_attributes.


pkg/mappings.bzl
```diff
@@ -72,15 +72,21 @@ strip_prefix = struct(
...
 )
 
-def pkg_attributes(mode = None, user = None, group = None, **kwargs):
+def pkg_attributes(
+        mode = None,
+        user = None,
+        group = None,
+        uid = None,
+        gid = None,
+        **kwargs):
```
nacl, line 81: This needs some more consistency checks.  It seems that at least `user`/`uid` and `group`/`gid` should be mutually exclusive, and perhaps `uid` and `gid` require each other.

pkg/mappings.bzl
```diff
@@ -92,6 +98,8 @@ def pkg_attributes(mode = None, user = None, group = None, **kwargs):
       mode: string: UNIXy octal permissions, as a string.
       user: string: Filesystem owning user.
       group: string: Filesystem owning group.
```
nacl, line 100: Clarify: consider saying "user name"/"group name"

pkg/mappings.bzl
```diff
@@ -92,6 +98,8 @@ def pkg_attributes(mode = None, user = None, group = None, **kwargs):
...
+      uid: string: Filesystem owning uid.
+      gid: string: Filesystem owning gid.
```
nacl, line 102: Perhaps expand the acronyms here?  e.g. uid = "user id".

pkg/private/pkg_files.bzl
```diff
@@ -105,10 +109,12 @@ def _process_pkg_dirs(content_map, pkg_dirs_info, origin, default_mode, default_
             user = attrs[1],
             group = attrs[2],
             origin = origin,
+            uid = attrs[3],
+            gid = attrs[4],
```
nacl, line 113: Consider sorting these differently here and below.  For consistency, it seems that `uid`/`gid` should be below `group` rather than `origin`.

pkg/private/pkg_files.bzl
```diff
@@ -163,6 +175,8 @@ def process_src(
       default_mode: fallback mode to use for Package*Info elements without mode
       default_user: fallback user to use for Package*Info elements without user
       default_group: fallback mode to use for Package*Info elements without group
+      default_uid: fallback uid to use for Package*Info elements without uid
+      default_gid: fallback gid to use for Package*Info elements without guid
```
nacl, line 179: Typo: `guid` -> `gid`?

nacl (changes requested):
Thanks for the patch!  Below are things I noticed, but unfortunately the manifest format changed recently, so there's a number of merge conflicts.

Some things I wonder about outside of the motivating use case (discussion in #670):
- What about `pkg_rpm`?  It's fine not to include it, as it might be a little complicated and/or not possible.
- Perhaps `uid`/`gid` should not be named in `pkg_attributes`?  An alternative interface design would be to detect if `user`/`group` are ints, and then convert that to `uid`/`gid` on the backend.  WDYT?

pkg/mappings.bzl
aiuto: Are they really exclusive?  IIRC, the tar format allows both.  This, of course, it prone to strangeness, like allowing   user=root, uid=42. I don't know if we should allow the footgun or not.

pkg/mappings.bzl
AustinSchuh: Done.

pkg/mappings.bzl
AustinSchuh: Done.

pkg/private/pkg_files.bzl
AustinSchuh: Done.

pkg/private/pkg_files.bzl
AustinSchuh: done

pkg/mappings.bzl
AustinSchuh: At this point, I think we document it, and let folks point the foot gun.  Tar allows both, but yes, it is a footgun.

aiuto (approved):

aiuto:
@nacl  I think it is blocked on you resolving a comment.

AustinSchuh:
Ping, @nacl , any additional feedback?

nacl (commented):
Sorry for the delays on this.  I got a bit stuck on some internal stuff and only recently got a chance to refocus.

This seems fine, but I think the docs could use a slight adjustment.  See below WRT the RPM tests.

> Ok, I think this now works again. I don't know how to run the RPM tests. They appear to be disabled locally. If you can give me a hint, I can see what I can do.

The RPM tests require external RPM tooling (`rpm` and `rpmbuild`).  On macOS, you can get a recent copy of these from homebrew.  On Linux, you can use your local package manager.  Either should work.

Unfortunately, RPM packaging is tricky and a portable toolchain setup is difficult to make.

LMK if you have any further issues.  If you're having issues with this, the documentation can be updated and this investigation can be added to the backlog.  Intuitively, it may not make sense for RPMs (and Debs) to support numeric ids.

aiuto:
Friendly ping.  This should get merged, can we come to closure on the last few nits?

pkg/mappings.bzl
```diff
@@ -72,15 +72,21 @@ strip_prefix = struct(
...
     """Format attributes for use in package mapping rules.
 
     If "mode" is not provided, it will default to the mapping rule's default
     mode.  These vary per mapping rule; consult the respective documentation for
     more details.
 
-    Not providing any of "user", or "group" will result in the package builder
-    choosing one for you.  The chosen value should not be relied upon.
+    Not providing any of "user", "group", "uid", or "gid" will result in the package
+    builder choosing one for you.  The chosen value should not be relied upon.
```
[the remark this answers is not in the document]
aiuto: Let's leave this for another pass.

aiuto (approved):

docs/latest.md
```diff
@@ -63,19 +63,22 @@ attributes = pkg_attributes(
...
+permissions. `uid` and `gid` correspond to the numeric uid/gid.  Attributes
+should always be specified using the `pkg_attributes` helper macro.
 
 Each mapping rule has some default mapping attributes.  At this time, the only
 default is "mode", which will be set if it is not otherwise overridden by the user.
 
 If `user` and `group` are not specified, then defaults for them will be chosen
-by the underlying package builder.  Any specific behavior from package builders
+by the underlying package builder.  If `uid` and `gid` are not specified,
+they will default to 0.  Any specific behavior from package builders
```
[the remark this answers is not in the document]
nacl: Deferring to follow-on task.  My understanding is that this file is generated and that this change will need to be made elsewhere eventually.

nacl (approved):

AustinSchuh:
Sorry for the long delay...  I went from having time to not having time really fast.

PR #672, opened 2023-02-13
pkg_tar, pkg_zip: improve support for long paths on Windows

beasleyr-vmw:
This commit updates `pkg_zip` to, when running on Windows, convert zip files' input paths to extended-length paths by (1) making them absolute and then (2) prepending with `\\?\`.

### DIFF
--- a/pkg/private/zip/build_zip.py
+++ b/pkg/private/zip/build_zip.py
@@ -193,14 +193,15 @@ class ZipWriter(object):
         dest_dir = dest
       to_write[dest_dir] = None
       for file in files:
+        content_path = os.path.abspath(os.path.join(root, file))
         if os.name == "nt":
           # "To specify an extended-length path, use the `\\?\` prefix. For
           # example, `\\?\D:\very long path`."[1]
           #
           # [1]: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
-          to_write[dest_dir + file] = "\\\\?\\" + os.path.abspath(os.path.join(root, file))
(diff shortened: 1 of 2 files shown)


pkg/private/zip/build_zip.py
```diff
@@ -193,7 +193,14 @@ def add_tree(self, tree_top: str, destpath: str, mode: int):
...
       to_write[dest_dir] = None
       for file in files:
-        to_write[dest_dir + file] = root + '/' + file
+        if os.name == "nt":
+          # "To specify an extended-length path, use the `\\?\` prefix. For
+          # example, `\\?\D:\very long path`."[1]
+          #
+          # [1]: https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+          to_write[dest_dir + file] = "\\\\?\\" + os.path.abspath(os.path.join(root, file))
+        else:
```
aiuto, line 202: Now that you are handling windows better.  I think it will be clearer if all paths go through abspath....
```
content_path = os.path.abspath(os.path.join(root, file))
if os.name == 'nt':
   to_write[...] = "\\\\?\\" + content_path
else
  to_write[...] = content_path
```
or similar

beasleyr-vmw: Sure thing. I found a similar snippet in build_tar.py. I'm going to include the same fix for that in this PR, but lmk if you'd prefer it to be dealt with separately.

aiuto (approved):

aiuto:
Thanks for doing this.

PR #680, opened 2023-03-13
exec_tools -> tools

aiuto:


katre (approved):

nacl (approved):
Maybe link to why this is needed?  Does this create any potential back compat concerns?

aiuto:
It might cause problems for people using Bazel 5.x to create a distribution, but it won't cause problems for users of the distribution.

PR #681, opened 2023-03-15
pkg_tar should not prefix tree artifacts with ./

### LINKED ISSUE #676: prefix-stripped directory artifacts end up with two versions, with and without leading .
For this BUILD.bazel file making a tar of a directory artifact:
```starlark
load("//:rules.bzl", "make_directory")

# Makes a single file in nested directories: dir-dir/subdir/file
make_directory(
    name = "dir",
)

# Should contain `./subdir/file`
pkg_tar(
    name = "tar",
    srcs = [":dir"],
    strip_prefix = "dir-dir",
)
```

With this simple directory-creating rule:
```starlark
def _impl(ctx):
    d = ctx.actions.declare_directory("%s-dir" % ctx.attr.name, sibling = None)
    ctx.actions.run_shell(
        outputs = [d],
        command = "mkdir -p $1/subdir && touch $1/subdir/file",
        arguments = [d.path],
    )
    return [DefaultInfo(files = depset([d]))]

make_directory = rule(
    implementation = _impl,
)
```

I expect to see just `./subdir/` and `./subdir/file` in the tar, but instead have both `subdir/` and `./subdir` as well as `./subdir/file`:

```console
% bazel build :tar            
INFO: Invocation ID: 8045b9ca-4244-4714-a1c2-6520e6d6fdf6
INFO: Analyzed target //:tar (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //:tar up-to-date:
  bazel-bin/tar.tar
INFO: Elapsed time: 0.135s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
% tar tvf bazel-bin/tar.tar
drwxr-xr-x  0 0      0           0  1 Jan  2000 subdir/
drwxr-xr-x  0 0      0           0  1 Jan  2000 ./subdir/
-r-xr-xr-x  0 0      0           0  1 Jan  2000 ./subdir/file
```

clint-stripe:
Fixes #676

It was enlightening to note that `build_zip` doesn't have the same problem! `build_zip`'s `add_tree` has a key special case that previously we didn't have in `build_tar`:
https://github.com/bazelbuild/rules_pkg/blob/main/pkg/private/zip/build_zip.py#L178-L194

### DIFF
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -232,6 +232,9 @@ class TarFile(object):
 
     # Again, we expect /-style paths.
     dest = normpath(dest)
+    # normpath may be ".", and dest paths should not start with "./"
+    dest = '' if dest == '.' else dest + '/'
+
     if ids is None:
       ids = (0, 0)
     if names is None:
@@ -246,9 +249,9 @@ class TarFile(object):
       dirs = sorted(dirs)
       rel_path_from_top = root[len(tree_top):].lstrip('/')
       if rel_path_from_top:
-        dest_dir = dest + '/' + rel_path_from_top + '/'
+        dest_dir = dest + rel_path_from_top + '/'
       else:
-        dest_dir = dest + '/'
+        dest_dir = dest
       for dir in dirs:
         to_write[dest_dir + dir] = None
       for file in sorted(files):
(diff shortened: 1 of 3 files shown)


aiuto:
Thanks for an easy to review PR.
- It fixes an issue that has been marked as a bug
- It comes with a test

aiuto (approved):

PR #684, opened 2023-03-23
Add OutputGroupInfo for pkg_rpm rule

k0walik:
The `pkg_deb` rule returns OutputGroupInfo that contains `deb = [output_file]` output group. We could add similar return value for `pkg_rpm` rule.

Here is the implementation for `pkg_deb`: https://github.com/bazelbuild/rules_pkg/blob/main/pkg/private/deb/deb.bzl#L166

The open question is whether we should match other groups? `changes` could contain `ctx.attr.changelog` value, but I'm not quite sure what would fit for `out` group.


aiuto:
But why?. Is there a bug that describes the thing you are trying to
accomplish.

k0walik:
Actually, I think the problem is not the OutputGroupInfo part, but DefaultInfo mismatch between `pkg_deb` and `pkg_rpm`:
- `pkg_deb` returns `files = depset([output_file])`, which is a single file: https://github.com/bazelbuild/rules_pkg/blob/main/pkg/private/deb/deb.bzl#L168
- `pkg_rpm` returns `files = depset([outputs])`, which contains two files: https://github.com/bazelbuild/rules_pkg/blob/main/pkg/rpm_pfg.bzl#L664

tests/deb_rpm_tar/BUILD
```diff
@@ -0,0 +1,65 @@
...
+    name = "test_deb",
+    data = ":deb_input",
+    description = "Some package",
+    maintainer = "john.doe@example.com",
+    package = "debpackage",
+    package_file_name = "debpackage-0.0.0.deb",
+    version = "0.0.0",
+)
+
+pkg_tar(
```
aiuto, line 27: Why would someone do this?

k0walik: What if I have to ship multiple files, i.e. instructions or deployment scripts, along with deb/rpm that cannot be part of the package to the air gapped environment? The tarball is the easiest way to distribute.

tests/deb_rpm_tar/test.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
aiuto, line 1: Just drop this test. If we want to test for a single file, then we should be doing an analysis test.

tests/deb_rpm_tar/BUILD
aiuto: OK. 
But we don't need any of this. There are existing test targets that exhibit the right characteristics. Or, if not, they should be in the respective test/rpm or test/deb trees.

pkg/rpm_pfg.bzl
```diff
@@ -659,7 +659,17 @@ def _pkg_rpm_impl(ctx):
...
+
+    output_groups = {
+        "out": [default_file],
+        "rpm": [output_file],
+        "changes": changes
+    }
     return [
+        OutputGroupInfo(**output_groups),
         DefaultInfo(
             files = depset(outputs),
```
aiuto, line 674: The right change here is to align with the other tools.  It should be
```
DefaultInfo(files = depset([output_file]), runfiles = ctx.runfiles(files = outputs))
```

BUT... that breaks existing tests.  So, what don't we reduce this CL to just the changes to this file, which are reasonable, and come back to the problem of pkg_rpm giving two output files later.

tests/deb_rpm_tar/test.sh
k0walik: Done

PR #688, opened 2023-03-28
Remove top level .bzl files. They are all in //pkg.

aiuto:
These are left over from an early stage where the rules were not under //pkg.

RELNOTES: pkg_* rules now must be loaded from @rules_pkg//pkg:x.bzl. The legacy redirections at the top level are removed.


nacl (changes requested):
Looks like there's a few more .bzl files in the root that don't need to be there, e.g. [install.bzl](https://github.com/aiuto/rules_pkg/blob/nits/install.bzl), [mappings.bzl](https://github.com/aiuto/rules_pkg/blob/nits/mappings.bzl), [package_variables.bzl](https://github.com/aiuto/rules_pkg/blob/nits/package_variables.bzl).  Should they be removed too?

aiuto:
> Looks like there's a few more .bzl files in the root that don't need to be there, e.g Should they be removed too?
Done

nacl (approved):

Issue #659, opened 2023-01-11
rules_deb produces invalid Files- and Checksum-sections


msiebuhr:
Per https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-files, `Files` (and Checksum-XX)-sections must  have the first line empty, ex:

```
Files:
 c6f698f19f2a2aa07dbb9bbda90a2754 571925 example_1.2.orig.tar.gz
```

Yet `rules_deb` produces:

```
Files: c6f698f19f2a2aa07dbb9bbda90a2754 571925 example_1.2.orig.tar.gz
```

The relevant code _does_ try to do the right thing when generating the fields

https://github.com/bazelbuild/rules_pkg/blob/547077f0b1d3f60a325429a9b4c6de965ab2f3b4/pkg/private/deb/make_deb.py#L299-L307

But `MakeDebianControlField()` promptly strips it away

 https://github.com/bazelbuild/rules_pkg/blob/547077f0b1d3f60a325429a9b4c6de965ab2f3b4/pkg/private/deb/make_deb.py#L134-L138

A trivial fix could be to write out `Files:\n  " + " ".join([...])` and skip calling `MakeDebianControlField()`?

msiebuhr:
For now, I have hacked around it by piping my .changes-file through `sed -E 's/(Files|Checksums-.*): /\1:\n  /'`.

aiuto:
Might be the same as #632

phlax:
looks same to me

Issue #632, opened 2022-10-28
`pkg_deb` file spec seems to be corrupted in `.changes` file


phlax:
Attempting to upgrade to 0.7.1 our deb publishing pipeline fails unable to use `dscverify` to check the signatures on the debs/changes

On further investigation - this is what changed:

```dsc

Files: 
 104017c7c884bc833a3c46d044295f51 99165210 contrib/devel optional envoy_1.25.0-dev_amd64.deb

```

->

```dsc
Files: ad524b04821992a3037b2ef26c55f8f2 99165210 contrib/devel optional envoy_1.25.0-dev_amd64.deb
```

Checking the spec - it seems like this is not allowed - https://www.debian.org/doc/debian-policy/ch-controlfields.html#files

certainly it causes `dscverify` to fail with:

```console

# dscverify --keyring /usr/share/debsign/keyring.gpg /tmp/install/deb/envoy_1.25.0-dev_amd64.bullseye.changes 
/tmp/install/deb/envoy_1.25.0-dev_amd64.bullseye.changes:
      Good signature found
dscverify: no file spec lines in /tmp/install/deb/envoy_1.25.0-dev_amd64.bullseye.changes
Validation FAILED!!

```

aiuto:
That was probably introduced by the cleanup to handling multiline fields like Description.
TBH, I'm not going to look at anything in rules_pkg until December at the earliest.
Happy to take a PR. My hunch is that we want a different variant or option to MakeDebianControlField
that simply takes a list of values and emits them joined with '\n '.

It would also be great to have deb verification tests that work on CI.

PR #691, opened 2023-03-29
[pkg_deb] Fix multiline fields in changes file

Homulvas:
Fix broken `.changes` file reported in https://github.com/bazelbuild/rules_pkg/issues/632 
As per documentation https://www.debian.org/doc/debian-policy/ch-controlfields.html#debian-changes-files-changes
Files and checksum fields always start with an empty line.
Fix this by marking the fields as `is_multiline=True`. Also remove the trailing whitespace after a single line includes just the field key. Added a test that checks for an exact match to generated `.changes` file.

fixes https://github.com/bazelbuild/rules_pkg/issues/632 
fixes https://github.com/bazelbuild/rules_pkg/issues/659

### DIFF
--- a/pkg/private/deb/make_deb.py
+++ b/pkg/private/deb/make_deb.py
@@ -139,7 +139,7 @@ def MakeDebianControlField(name: str, value: str, is_multiline:bool=False) -> st
           '\\n is not allowed in simple control fields (%s)' % value)
 
   lines = value.split('\n')
-  result = name + ': ' +lines[0].strip() + '\n'
+  result = (name + ': ' +lines[0].strip()).strip() + '\n'
   for line in lines[1:]:
     if not line.startswith(' '):
       result += ' '
@@ -298,13 +298,16 @@ def CreateChanges(output,
           is_multiline=True),
       MakeDebianControlField(
           'Files', '\n ' + ' '.join(
-              [checksums['md5'], debsize, section, priority, deb_basename])),
+              [checksums['md5'], debsize, section, priority, deb_basename]),
+              is_multiline=True),
       MakeDebianControlField(
           'Checksums-Sha1',
-          '\n ' + ' '.join([checksums['sha1'], debsize, deb_basename])),
+          '\n ' + ' '.join([checksums['sha1'], debsize, deb_basename]),
+          is_multiline=True),
       MakeDebianControlField(
           'Checksums-Sha256',
-          '\n ' + ' '.join([checksums['sha256'], debsize, deb_basename]))
+          '\n ' + ' '.join([checksums['sha256'], debsize, deb_basename]),
+          is_multiline=True)
   ])
   with open(output, 'wb') as changes_fh:
     changes_fh.write(changesdata.encode('utf-8'))
--- a/pkg/private/deb/make_deb.py
+++ b/pkg/private/deb/make_deb.py
@@ -14,6 +14,7 @@
 """A simple cross-platform helper to create a debian package."""
 
 import argparse
+from enum import Enum
 import gzip
 import hashlib
 import io
@@ -29,6 +30,8 @@ else:
   OrderedDict = dict
 
 from pkg.private import helpers
+
+Multiline = Enum('Multiline', ['NO', 'YES', 'YES_ADD_NEWLINE'])
 
 # list of debian fields : (name, mandatory, is_multiline[, default])
 # see http://www.debian.org/doc/debian-policy/ch-controlfields.html
@@ -118,7 +121,7 @@ def AddArFileEntry(fileobj, filename,
     fileobj.write(b'\n')  # 2-byte alignment padding
 
 
-def MakeDebianControlField(name: str, value: str, is_multiline:bool=False) -> str:
+def MakeDebianControlField(name: str, value: str, multiline:Multiline=Multiline.NO) -> str:
   """Add a field to a debian control file.
 
   https://www.debian.org/doc/debian-policy/ch-controlfields.html#syntax-of-control-files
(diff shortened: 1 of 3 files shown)


Homulvas:
need to abstract checksums and date from the test scenario

pkg/private/deb/make_deb.py
```diff
@@ -139,7 +139,7 @@ def MakeDebianControlField(name: str, value: str, is_multiline:bool=False) -> st
           '\\n is not allowed in simple control fields (%s)' % value)
 
   lines = value.split('\n')
-  result = name + ': ' +lines[0].strip() + '\n'
+  result = (name + ': ' +lines[0].strip()).strip() + '\n'
```
aiuto, line 142: This seems very subtle.  I asked  how can` ("text" + x.strip()).strip()` be any different from `('text' + x.strip()`?
Is this to protect against the case where  x is '\n' because `value` above is `"\nsome text\n..."`
It seems yes, because we add the leading newline in the calls.
Perhaps this could all be more readable if we did not pass in the leading newline and changed above to be
`lines = [x.strip() for x in values.strip().split('\n')]`.

Homulvas: name + ‘: ‘ results in a trailing space of the first line is empty (as it should be for these fields).
On the other hand description in the control file has both the summary line that goes straight after the key and a multiline text after. Maybe we should indeed treat the control description field as a special case as right now it’s the only multiline control field.

Homulvas: made explicit handling for two separate multiline scenarios with an enum

pkg/private/deb/make_deb.py
```diff
@@ -30,6 +31,8 @@
 
 from pkg.private import helpers
 
+Multiline = Enum('Multiline', ['NO', 'YES', 'YES_ADD_NEWLINE'])
```
aiuto, line 34: Why are you making a tri-state when there are only 2. They are either 1 line, or they are multiline and all in the same file. Having the YES state just forces us to have tests for something that doesn't get used by the code.

tests/deb/control_field_test.py
```diff
@@ -52,23 +52,23 @@ def test_multiline(self):
     self.assertEqual(
         'Description: fizzbuzz\n',
         make_deb.MakeDebianControlField(
-            'Description', 'fizzbuzz', is_multiline=True))
+            'Description', 'fizzbuzz', multiline=make_deb.Multiline.YES))
```
aiuto, line 55: The tests  should all be for the format we actually use, YES_ADD_NEWLINE.

aiuto (changes requested):

pkg/private/deb/make_deb.py
Homulvas: Description in the control file is the third state (see line 166).
We can treat both multiline scenarios as the same thing but that was exactly what my previous attempt was. In that case we end up with strings that start with spaces and logic that needs to trim extra spaces in certain scenarios.

Please state which approach do you prefer and what changes would you like to see to get that landed.

tests/deb/control_field_test.py
Homulvas: We use both. Admittedly this should also include a case for `YES_ADD_NEWLINE` but that is already tested by the changes file test.

aiuto: I don't see any calls to MakeDebianControlField that use YES instead of YES_AND_ADD_NEWLINE.

Homulvas: In `CreateDebControl`
`multiline = Multiline.YES if values[2] else Multiline.NO` will be `Multiline.YES` for `Description` and will call `controlfile += MakeDebianControlField(fieldname, kwargs[key], multiline)`

aiuto:
So then the question is if that is correct or not. IRRC, there were another set of bugs about the control file fields being wrong. I don't recall if they were on github or internal.
Do you have pointers to the definitive specs. I'm going to be tied up for a while and won't have time to look deeply into that before end of April.

Homulvas:
> i would suggest using `dscverify` to validate the package files


before
```
$ dscverify fizzbuzz_4.5.6_all.changes --no-sig-check
fizzbuzz_4.5.6_all.changes:
dscverify: no file spec lines in fizzbuzz_4.5.6_all.changes
Validation FAILED!!
```
after
```
$ dscverify fizzbuzz_4.5.6_all.changes --no-sig-check
fizzbuzz_4.5.6_all.changes:
   validating fizzbuzz_4.5.6_all.deb
All files validated successfully.
```

before:
```
$ changestool fizzbuzz_4.5.6_all.changes verify
Checking Source packages...
Checking Binary consistency...
ERROR: 'fizzbuzz_4.5.6_all.changes' has binary 'fizzbuzz' in 'Binary:' and 'Description:' header, but no files for it found!
Checking checksums...
```
after:
```
$ changestool fizzbuzz_4.5.6_all.changes verify
Checking Source packages...
Checking Binary consistency...
Checking checksums...
```

pkg/private/deb/make_deb.py
aiuto: The core problem is that you're trying to improve on a quick hack which was badly thought through 10+ years ago.  There are a few distinct cases  which were shoehorned into one method.  One is the "description + list of elements".  Another is single line   Name: value.    The final was the "First line is on the field line and the rest is reflowed multiline."    Since the concepts were never clear, people (including myself) have been implementing this wrong since the beginning.

Making this more maintainable is low priority for me right now, so let's just take yours as is. If it ends up breaking users who have relied on the previous behavior, I'll figure something out at that time.

aiuto (approved):

Homulvas:
@aiuto can you land this as I don't have access to that?

PR #701, opened 2023-05-18
fix working dir on integration tests

aiuto:
The old spec was just wrong. bazelci must have updated the framework, and now errors out on it.


aiuto:
Force merging to fix CI.

PR #705, opened 2023-05-26
Use Label constructor for verify_archive_test macro

### LINKED ISSUE #704: `verify_archive_test` refers to non-existent package `pkg`
In experimenting with the new `verify_archive_test` macro I noticed using it outside of `rules_pkg` itself would lead to a loading error.

Example `rules_pkg_test/BUILD`:
```starlark
load("@rules_pkg//pkg:verify_archive.bzl", "verify_archive_test")

pkg_tar(
    name = "external_archive",
    srcs = ["file.txt"],
)

verify_archive_test(
    name = "external_archive_test",
    max_size = 1,
    min_size = 1,
    must_contain = ["file.txt"],
    target = ":external_archive",
)
```

When building/testing the `external_archive_test` target Bazel produces the following output (line numbers omitted):

```
rules_pkg_test/BUILD: no such package 'pkg': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - ~/rules_pkg_test/pkg and referenced by '//:external_archive_test'
```

It appears the `verify_archive_test` macro is evaluating `"//pkg:verify_archive_test_lib"` inside the caller's BUILD context instead of within `rules_pkg`.

sschwebach:
Fixes #704 

One of the `deps` within the `verify_archive_test` macro depends on a label, which would be incorrectly evaluated in a caller's `BUILD` file to be a non-external package.

This change wraps the label in a `Label()` call to ensure it's referenced properly in the `rules_pkg` context.

I added a new external repository to make a test since it didn't quite seem appropriate to put it in the mapping repository, though I can certainly do so if desired.

### DIFF
--- a/pkg/verify_archive.bzl
+++ b/pkg/verify_archive.bzl
@@ -122,7 +122,7 @@ def verify_archive_test(name, target,
         data = [target],
         python_version = "PY3",
         deps = [
-            "//pkg:verify_archive_test_lib",
+            Label("//pkg:verify_archive_test_lib"),
             "@bazel_tools//tools/python/runfiles",
         ],
     )
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -32,6 +32,12 @@ local_repository(
     path = "tests/mappings/external_repo",
 )
 
+# Used to test invoking tar rules (such as verify_archive_test) in an external repo
+local_repository(
+    name = "tar_tests_external_repo",
+    path = "tests/tar/external_repo",
+)
+
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
--- a/tests/tar/BUILD
+++ b/tests/tar/BUILD
@@ -283,7 +283,13 @@ verify_archive_test(
     must_not_contain_regex = [
         "^five.is.right.out",
     ],
-    max_size = 2,
+
+test_suite(
+    name = "verify_archive_test_suite",
+    tests = [
+        ":repackaging_long_filename_test",
+        "@tar_tests_external_repo//pkg:external_archive_test",
+    ],
 )
 
 pkg_tar(
--- a/tests/tar/external_repo/WORKSPACE
+++ b/tests/tar/external_repo/WORKSPACE
@@ -1,0 +1,15 @@
+# Copyright 2023 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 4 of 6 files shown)


pkg/verify_archive.bzl
```diff
@@ -122,7 +122,7 @@ def verify_archive_test(name, target,
         data = [target],
         python_version = "PY3",
         deps = [
-            "//pkg:verify_archive_test_lib",
+            Label("//pkg:verify_archive_test_lib"),
```
aiuto, line 125: This doesn't seem right. We shouldn't have to use Label() around deps like this.
What was your setup when you noticed the problem in the first place?
Were you using the distribution or trying to vendor in the code?

aiuto (changes requested):
I would like to avoid adding more local workspaces for tests.
If we end up needing this, it should be folded into the other one we have.

pkg/verify_archive.bzl
sschwebach: I initially noticed the issue after bringing in `rules_pkg` as a dependency in my own project and creating a `verify_archive_test` target there. The error message made it look like it was searching for `//pkg:verify_archive_test_lib` from my own workspace instead of `rules_pkg`.

aiuto: Oh... I see it now. Because it's not a `py_test`, it's a *generated* `py_test`.

aiuto:
I'll right. I'm stumped. I tried with both Label(//pkg:verify_archive_test_lib) and as @rules_pkg//pkg:varify_archive_test_lib.  The integration tests fail.  
I think your change is right, but I don't understand why the test fails.

aiuto:
I am now thinking this is needlessly efficient.
We could just fold the test lib into the .tpl file and eliminate the library. Then the problem you had would go away.

sschwebach:
> I'll right. I'm stumped. I tried with both Label(//pkg:verify_archive_test_lib) and as @rules_pkg//pkg:varify_archive_test_lib. The integration tests fail. I think your change is right, but I don't understand why the test fails.

I'm equally stumped on this one. From what I can tell there shouldn't be any difference using this from one external workspace compared to another. I'll see if I can find what might be causing the issue before looking at merging the library into the template.

aiuto:
I'm even more convinced that merging into the template is the best solution. It means that if you import rules_pkg into your own third_party, you don't have to rebase the import of the library.
In fact, my comment in the template "Hey reviewer!!! What if we just added the source to the test lib here, so we would not have to make the library for that public?" hints that I considered that in the first implementation.

sschwebach:
I've pushed a change to merge the two together, along with an additional change to remove the `json` and `runfiles` imports that don't seem to be used. I also edited the title/description since it's a bit different than the original change.

pkg/verify_archive_test_main.py.tpl
```diff
@@ -15,9 +15,95 @@
 
 import unittest
```
aiuto, line 16: This is redundant now

aiuto (changes requested):
Nice. Just one extra import.

And now the tests pass.

pkg/verify_archive_test_main.py.tpl
sschwebach: Fixed.

aiuto (approved):

PR #707, opened 2023-06-05
Add tar `xattr` support

bozaro:
# Motivation
I need tool to set "max locked memory" ulimit for `mongos` inside Kubernetes environment.
This tool need `etcap cap_sys_resource+ep` xattr on file.

This PR allows set `xattr` on files inside .tar artchives.

# Example
Allow set `xattr` inside tar archives, for example:
```
pkg_tar(
    name = "tool-layer",
    extension = "tar",
    files = {
        ":ulimit-wrapper": "/usr/local/bin/ulimit-wrapper",
    },
    # setcap cap_sys_resource+ep ulimit-wrapper && getfattr -d -e hex -m - ulimit-wrapper
    xattr = [
        "security.capability=0x0100000200000001000000000000000000000000",
    ],
)
```


aiuto:
Can you start a bug about this so we can get the requirements right first.
It looks like you are setting xattr on each file.  If that is the need, then we should raise this to the level of other attributes, like owner and perms, so we can apply it to each file distinctly.

bozaro:
I want to make something like `Dockerfile`:
```Dockerfile
FROM ubuntu:22.04
...
COPY app /usr/local/bin/app
RUN setcap cap_sys_resource+ep /usr/local/bin/app
```

In `BUILD.bazel` I got rules like:
```starlark
load("@rules_pkg//:pkg.bzl", "pkg_tar")
load("@io_bazel_rules_docker//container:container.bzl", "container_image")

...

pkg_tar(
    name = "app-tar",
    extension = "tar",
    files = {
        ":app": "/usr/local/bin/app",
    },
    xattr = [
        "security.capability=0x0100000200000001000000000000000000000000",
    ],
)

container_image(
    name = "app-image",
    ...
    tars = [
        ":app-tar",
    ],
)
```

In this example `xattr` I set `xattr` to all one file, but in this PR also supported `xattrs` attribute like:
```starlark
    xattrs = {
    	"usr/local/bin/app": [
    	    "security.capability=0x0100000200000001000000000000000000000000",
        ]
    },
```

aiuto:
Can you think about how this should be in pkg_files.  Adding it only to pkg_tar is sort of a dead end.
We want tools that describe what you want independently of the final package type.

pkg/private/tar/build_tar.py
```diff
@@ -36,27 +37,47 @@ def normpath(path):
...
+  xattrs = {}
+  if xattr_list:
+    for item in xattr_list:
+      idx = item.index("=")
+      if idx < 0:
+        raise ValueError("Unexpected xattr item format {} (want key=value)".format(item))
+      key = item[:idx]
+      raw = item[idx+1:]
+      if raw.startswith("0x"):
+        xattrs[key] = bytes.fromhex(raw[2:]).decode('latin-1')
```
fruechel-canva, line 50: You probably want to use `.decode('utf-8', "surrogateescape")` here as certain `xattrs` will be re-encoded incorrectly otherwise. On re-encoding the `tarfile` module uses `surrogateescape` to restore any values that don't cleanly decode to UTF-8.

PR #708, opened 2023-06-05
Add stamping to .deb package version and description

bozaro:
Allow use stamping variables inside deb description.

### DIFF
--- a/pkg/private/deb/deb.bzl
+++ b/pkg/private/deb/deb.bzl
@@ -17,6 +17,7 @@ load("//pkg:providers.bzl", "PackageArtifactInfo", "PackageVariablesInfo")
 load("//pkg/private:util.bzl", "setup_output_files")
 
 _tar_filetype = [".tar", ".tar.gz", ".tgz", ".tar.bz2", "tar.xz"]
+_stamp_condition = Label("//pkg/private:private_stamp_detect")
 
 def _pkg_deb_impl(ctx):
     """The implementation for the pkg_deb rule."""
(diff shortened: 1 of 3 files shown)


pkg/private/deb/make_deb.py
```diff
@@ -348,6 +348,10 @@ def main():
   parser.add_argument(
       '--triggers',
       help='The triggers file (prefix with @ to provide a path).')
+  parser.add_argument(
+      '--stamp',
+      action='append',
```
aiuto, line 353: Why is this not a bool?  Can you update the description do describe what feature is being added?

pkg/private/deb/deb.bzl
```diff
@@ -107,6 +108,12 @@ def _pkg_deb_impl(ctx):
     else:
         fail("Neither description_file nor description attribute was specified")
 
+    # Files for stamping variables
+    if ctx.attr.private_stamp_detect:
```
aiuto, line 112: This should follow the logic of what the others do. Where there is a stamp parameter which can be 1, 0, or -1, and we only use the stamp dectect if -1.

pkg/private/deb/deb.bzl
```diff
@@ -107,6 +108,12 @@ def _pkg_deb_impl(ctx):
...
+        args += ["--stamp=@" + ctx.version_file.path]
+        args += ["--stamp=@" + ctx.info_file.path]
+        files += [ctx.version_file, ctx.info_file]
```
aiuto, line 115: Why the info file?

PR #710, opened 2023-06-07
Install ossf scorecard

aiuto:


meteorcloudy (approved):

meteorcloudy:
Yeah, ideally, we probably should, but there are 81 repos under bazelbuild, so it'll probably take some work..

PR #712, opened 2023-06-27
Write debian changes Date field in UTC rather than local time

aiuto:
This fixes some subtle problems caused by having CI and developer machines in different time zones.
It's also the right thing.

### DIFF
--- a/pkg/private/deb/make_deb.py
+++ b/pkg/private/deb/make_deb.py
@@ -288,7 +288,7 @@ def CreateChanges(output,
 
   changesdata = u''.join([
       MakeDebianControlField('Format', '1.8'),
-      MakeDebianControlField('Date', time.ctime(timestamp)),
+      MakeDebianControlField('Date', time.asctime(time.gmtime(timestamp))),
       MakeDebianControlField('Source', package),
       MakeDebianControlField('Binary', package),
       MakeDebianControlField('Architecture', architecture),


aiuto:
Very strange.  Now the tests are failing on mac because of some strange unicode handling of the Description field.
Do you know if we changed the python versions on buildkite recently?

tetromino:
I can reproduce this in local testing on my Macbook with Bazel HEAD; everything worked fine with 6.2.1

rules_pkg python code is correct AFAICT.

Something in Bazel at HEAD broke argument encoding in spawned processes on macOS  :/

tetromino:
https://github.com/bazelbuild/bazel/issues/18792 - update to JDK 20 broke non-ascii argument encoding on macOS.

sdtwigg:
I think I should insist you squash these commits into one commit.

aiuto:
> I think I should insist you squash these commits into one commit.

That rebase is tricky. It fails on each squashed change and I have to hand edit.  The merge squashes it.

aiuto:
Oh... the tests are failing for orthogonal reasons..

aiuto:
I rebased it down. You can't tell in the github UI though.

tetromino:
For me in manual testing, after https://github.com/bazelbuild/bazel/commit/f00439d8b7dd5b10ac357c413928418e0e1059c7, //tests/deb:pkg_deb_test passes on macOS.

But it still fails on the CI machine at the exact same git commit :/

sdtwigg (approved):

PR #714, opened 2023-06-28
Update pkg_deb to use ctx.action.args() for building the command line.

aiuto:
Update pkg_deb to use ctx.action.args() for building the command line.

- This cleans up a legacy style, and gets us better arg quoting.
- But if your attributes  are not-ASCII, this is insufficient. You need a version of bazel past https://github.com/bazelbuild/bazel/pull/18972
- See https://github.com/bazelbuild/bazel/issues/18792 for a fun discussion of both problem.

### DIFF
--- a/pkg/private/deb/deb.bzl
+++ b/pkg/private/deb/deb.bzl
@@ -75,6 +75,9 @@ def _pkg_deb_impl(ctx):
     if ctx.attr.triggers:
         args.add("--triggers", "@" + ctx.file.triggers.path)
         files.append(ctx.file.triggers)
+    if ctx.attr.md5sums:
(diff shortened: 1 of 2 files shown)


pkg/private/deb/deb.bzl
```diff
@@ -34,79 +34,82 @@ def _pkg_deb_impl(ctx):
         package_file_name = package_file_name,
     )
 
-    changes_file = ctx.actions.declare_file(output_name.rsplit(".", 1)[0] + ".changes")
+    out_file_name_base = output_name.rsplit(".", 1)[0]
```
cgrindel: Have you considered using skylib's `paths` module to handle path-related manipulations and extraction?

cgrindel (approved):

PR #723, opened 2023-08-08
Require python3 for executable scripts

tetromino:
Otherwise, on old Linux systems where python == python2, we try to use python2 and fail: https://buildkite.com/bazel/bazel-skylib/builds/2768#0189d6bc-463d-4c32-ad02-46ed3c01da5d


tetromino:
I misdiagnosed the problem: the issue is that python 3.5 does not support the "foo: type" syntax for variables.

To fix, I think we'd need to depend on a recent rules_python.

Issue #696, opened 2023-05-02
Add changelog attribute to pkg_deb rule


k0walik:
I noticed the `pkg_deb` rule does not provide `changelog` input attribute, while `pkg_rpm` does. Is it intended behavior? My use case assumes creating DEB and RPM packages, where the changelog is automatically generated and forwarded to package generation rules.

While looking at the rule implementation it seems to be trivial task.

aiuto:
Not intended. It's just that no one ever bothered to add it.

PR #725, opened 2023-08-11
Add changelog attribute to pkg_deb

k0walik:
Allow passing changelog file to `pkg_deb` rule.
This addresses #696 issue.


aiuto:
It should not automatically propagate to the .changes file.  That choice should be up to each target. Personally, I would never put the raw changelog into a binary distribution anyway.

aiuto (approved):

PR #726, opened 2023-08-11
Append changelog content to RPM spec file

k0walik:
Consume input changelog file and append the content to generated spec file


pkg/make_rpm.py
```diff
@@ -270,8 +272,13 @@ def SetupWorkdir(self,
       'POST_SCRIPTLET': "%post\n" + self.post_scriptlet,
       'PREUN_SCRIPTLET': "%preun\n" + self.preun_scriptlet,
       'POSTUN_SCRIPTLET': "%postun\n" + self.postun_scriptlet,
+      'CHANGELOG': ""
     }
 
+    if changelog_file:
+      self.changelog = SlurpFile(os.path.join(original_dir, changelog_file))
+      tpl_replacements["CHANGELOG"] = "%changelog\n" + self.changelog
```
aiuto, line 280: Is this right? It just dumps in unquoted?  What if the changelog file contains
`%pre` as on a line by itself.

k0walik: The rpm packaging guideline doesn't say anything about quoting the content of changelog file, only the format itself (https://rpm-packaging-guide.github.io/):

> The last section, `%changelog`, is a list of datestamped entries for each Version-Release of the package. They log packaging changes, not software changes. Examples of packaging changes: adding a patch, changing the build procedure in `%build`.
>
>Follow this format for the first line:
>
>* Day-of-Week Month Day Year Name Surname <email> - Version-Release
>
>Follow this format for the actual change entry:
>
>Each change entry can contain multiple items - one for each change
>
>Each item starts on a new line.
>
>Each item begins with a - character.
>
>An example datestamped entry:

```
%changelog
* Tue May 31 2016 Adam Miller <maxamillion@fedoraproject.org> - 0.1-1
- First bello package
- Example second item in the changelog for version-release 0.1-1
```

In our case the changelog attribute should only contain relevant entries, similar to pkg_deb rule, and all we need is to place the file content in the right location of spec file. If the changelog contains incorrect content, such as `%pre` you mentioned, then it's up to rpmbuild to detect this and fail building package. Unless you meant something different?

aiuto (approved):

Issue #727, opened 2023-08-13
Pkg_rpm fails to build packages targeted to platforms other than the host machine.


ogalbxela:
The current version of the "pkg_rpm" rule exposes an "architecture"-argument, which is meant to indicate the package's intended platform. It is being used to form the package name (with regard to "arch"-suffix) as well as to to define the "BuildArch" option in the generated RPM spec file.
However, as per RPM packaging guide (https://rpm-packaging-guide.github.io/):
"'BuildArch' should be used if the package is not architecture dependent. For example, if written entirely in an interpreted programming language, set this to BuildArch: noarch. If not set, the package automatically inherits the Architecture of the machine on which it is built".
    
It appears it actually has nothing to do with target architecture. Instead it represents either architecture of host machine or "noarch".

Possible solution to address this concern is use to pass the value "architecture"-argument to the "rpmbuild"-utility as the value for the "--target" option. It may break compatibility as "architecture"-argument it might already be in use by someone who relies on current behavior. So introducing brand new "target_architecture"-argument may also be an option here. 

Please consider possible fix implementation https://github.com/Mellanox/rules_pkg/pull/1. Note: proposed patch attempts to back compatibility concern. It also currently lack of tests to cover newly added option. Also, whoever will review the issue/patch, please advise proper way to test it.

Below is a "synthetic" example aimed to demonstrate concern. Note: in order to reduce sample size, I removed cross compilation part here just providing binary pre-compiled for aarch64

Versions are: Bazel: 6.0.0; Rules_pkg: 0.9.1; host: ubuntu 20.04.2 x86_64; target: linux-aarch64

```shell
$ uname -a
Linux alexbl-dev-home 5.15.0-46-generic #49~20.04.1-Ubuntu SMP Thu Aug 4 19:15:44 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
$ 
$ 
$ tree
.
├── src
│   ├── binary-aarch64
│   └── BUILD
└── WORKSPACE
$ 
$ 
$ file src/binary-aarch64 
src/binary-aarch64: ELF 64-bit LSB executable, ARM aarch64, version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux-aarch64.so.1, for GNU/Linux 3.7.0, not stripped
$ 
$ 
$ rpmbuild --version
RPM version 4.14.2.1
$ 
$ 
$ cat WORKSPACE
[...]

PR #729, opened 2023-08-13
Support for cross-platform RPM package generation

ogalbxela:
Please consider a patch aimed to add proper support for constructing RPM packages aimed for the CPU architecture other than the host machine. This addresses https://github.com/bazelbuild/rules_pkg/issues/727 issue.

This change introduces proper support for generating RPM packages for the designated target platform.

The current version of the "pkg_rpm"-rule exposes an "architecture"-option, which is meant to indicate the package's intended platform. Unfortunately, this option is being incorrectly utilized. It is being used to define the "BuildArch"-option in the RPM spec file. However, as per RPM packaging guide (https://rpm-packaging-guide.github.io/):
  "'BuildArch' should be used if the package is not architecture dependent. For
  example, if written entirely in an interpreted programming language, set this
  to BuildArch: noarch. If not set, the package automatically inherits the
  Architecture of the machine on which it is built".

So, it sounds like it actually has nothing to do with target architecture. It is supposed to represent either architecture of host machine or "noarch".

The purpose of this patch is to maintain backward compatibility while provide the solution for the described issue.
It preserves the existing "architecture"-attribute of the "pkg_rpm" rule, as it might already be in use by someone.
It introduces the "target_architecture"-attribute to the "pkg_rpm" rule, aiming to utilize it as a "arch"-suffix for the package name and to provide it to the "rpmbuild"-utility as the value for the "--target" option.

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -178,13 +178,14 @@ class RpmBuilder(object):
   RPMS_DIR = 'RPMS'
   DIRS = [SOURCE_DIR, BUILD_DIR, RPMS_DIR, TEMP_DIR]
 
-  def __init__(self, name, version, release, arch, rpmbuild_path,
-               source_date_epoch=None,
+  def __init__(self, name, version, release, arch, target_arch,
+               rpmbuild_path, source_date_epoch=None,
                debug=False):
     self.name = name
     self.version = helpers.GetFlagValue(version)
     self.release = helpers.GetFlagValue(release)
     self.arch = arch
+    self.target_arch = target_arch
     self.files = []
     self.rpmbuild_path = FindRpmbuild(rpmbuild_path)
     self.rpm_path = None
@@ -354,6 +355,10 @@ class RpmBuilder(object):
         '--buildroot=%s' % buildroot,
     ]  # yapf: disable
 
+    # Target platform
+    if self.target_arch:
+      args += ['--target=%s' % self.target_arch]
+
     # Macro-based RPM parameter substitution, if necessary inputs provided.
     if self.preamble_file:
       args += ['--define', 'build_rpm_options %s' % self.preamble_file]
@@ -462,7 +467,10 @@ def main(argv):
                       help='The release of the software being packaged.')
   parser.add_argument(
       '--arch',
-      help='The CPU architecture of the software being packaged.')
+      help='The CPU architecture of the machine on which it is built. '
+           'If the package is not architecture dependent, set this to "noarch".')
+  parser.add_argument('--target_arch',
+      help='The CPU architecture of the target platform the software being packaged for.')
   parser.add_argument('--spec_file', required=True,
                       help='The file containing the RPM specification.')
   parser.add_argument('--out_file', required=True,
@@ -501,7 +509,7 @@ def main(argv):
   try:
     builder = RpmBuilder(options.name,
                          options.version, options.release,
-                         options.arch, options.rpmbuild,
+                         options.arch, options.target_arch, options.rpmbuild,
                          source_date_epoch=options.source_date_epoch,
                          debug=options.debug)
     builder.AddFiles(options.files)
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -142,7 +142,7 @@ def _make_absolute_if_not_already_or_is_macro(path):
     # this can be inlined easily.
     return path if path.startswith(("/", "%")) else "/" + path
 
-#### Input processing helper functons.
+#### Input processing helper functions.
 
 # TODO(nacl, #459): These are redundant with functions and structures in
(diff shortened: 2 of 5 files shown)


aiuto:
Can we figure the right behavior out in the issue before jumping to code.
I'll comment there.

ogalbxela:
> Can we figure the right behavior out in the issue before jumping to code. I'll comment there.

Thank you, Tony, I answered in the issue.

pkg/make_rpm.py
```diff
@@ -462,7 +467,10 @@ def main(argv):
                       help='The release of the software being packaged.')
   parser.add_argument(
       '--arch',
-      help='The CPU architecture of the software being packaged.')
+      help='The CPU architecture of the machine on which it is built. '
+           'If the package is not architecture dependent, set this to "noarch".')
```
aiuto, line 471: Doesn't this comment apply to targert_arch?

pkg/rpm_pfg.bzl
```diff
@@ -791,21 +794,30 @@ pkg_rpm = rule(
         # funny if it's not provided.  The contents of the RPM are believed to
         # be set as expected, though.
         "architecture": attr.string(
-            doc = """Package architecture.
+            doc = """Host architecture.
 
             This currently sets the `BuildArch` tag, which influences the output
```
aiuto, line 799: Let's drop "currently" in these

pkg/rpm_pfg.bzl
```diff
@@ -791,21 +794,30 @@ pkg_rpm = rule(
...
             architecture of the package.
 
             Typically, `BuildArch` only needs to be set when the package is
```
aiuto, line 802: You might as well make the exact  behavior and use clear and tie it back to Bazel definitions when the terms overlap. The points to make are

- If the package is really architecture independent, say "noarch"
- Otherwise specify a CPU architecture of the target platform your are building for.
- If you say nothing, it will default to the CPU arch of your execution platform.

Note that when cross compiling the target and execution platforms may not match.

pkg/rpm_pfg.bzl
```diff
@@ -791,21 +794,30 @@ pkg_rpm = rule(
...
-            only valid for specific architectures.
+            not architecture dependent (e.g. written in an interpreted
+            language).
 
             When no attribute is provided, this will default to your host's
             architecture.  This is usually what you want.
 
             """,
         ),
+        "target_architecture": attr.string(
```
aiuto, line 811: Do we need both?  I have not figured out the value of setting both, since only one gets used.

aiuto:
Friendly ping

jiceatscion:
Any plan to ever merge this? It is pity that this very convenient rule cannot be used for general rpm builds, while a solution is just sitting here in limbo.

aiuto:
> Any plan to ever merge this? ...

Good question.  I keep seeing it in the overview as waiting on the original author, so I have been ignoring it.
It does seem OK though. I can look over the weekend.

jiceatscion:
Any hope off seeing this merged?

aiuto:
> Any hope off seeing this merged?

Yes. If you can work with me on this.  Have you patched it in and tried it?  Does it have the right behaviors?
The original author went cold on answering my questions.

jiceatscion:
> > Any hope off seeing this merged?
> 
> Yes. If you can work with me on this. Have you patched it in and tried it? Does it have the right behaviors? The original author went cold on answering my questions.

Awsome, I will gladly try this and get back to you!

pkg/rpm_pfg.bzl
chickenandpork: @aiuto @ogalbxela  when cross-compiling, target/exec may differ, but the logical default value would be the target.

When not cross-compiling, the target and exec match, so the logical default value would be the target.

Is there interest in setting the default to the target rather than exec platform?  Or a risk of breaking change in the default ?

ogalbxela:
Hi @aiuto ,

It sounds like there's eventually some demand for this functionality.

Let me summarize your points before I touch the code again. From what I understand, your suggestions are as follows:
  - Use the existing “arch” argument instead of adding a new “target_arch” argument. (The only reason for considering a new argument was backward compatibility for users who may already be using the “arch” argument since behavour will be changed)
  - If "arch" is set to "noarch", do not pass the "--target" flag to rpmbuild. Instead, add the line “BuildArch=noarch” to the RPM spec.
  - If  "arch" is set to something other than "noarch", pass the “--target=<arch>” flag to rpmbuild using that value.
  - If the “arch” is not set, derive target architecture from the toolchain ("target_cpu"-parameter of create_cc_toolchain_config_info) and pass the “--target=<arch>” flag to rpmbuild using that derived value.

Please let me know if I missed/missinterpreted anything.

vinsburg:
Can anything be done to promote this PR? :crossed_fingers:

PR #730, opened 2023-08-15
chore: update versions of dependencies

### LINKED ISSUE #698: WORKSPACE setup gives ancient rules_python
Users following the install instructions may end up running rules_python 0.1.0 due to 
https://github.com/bazelbuild/rules_pkg/blob/main/pkg/deps.bzl#L36
It depends whether they call `rules_pkg_dependencies()` earlier during their WORKSPACE setup than some other macro that installs a newer rules_python.

(note, bzlmod users don't have this problem thanks to Minimal Version Selection)

alexeagle:
In particular, our dependency on rules_python 0.1.0 is a risk for users who run rules_pkg_dependencies too early in their WORKSPACE. Note, under bzlmod this isn't a concern at all.

Somewhat breaking change, as users who were relying on the versions we provide will now get the updates, but rules_pkg isn't 1.0 yet.

Fixes #698

### DIFF
--- a/pkg/deps.bzl
+++ b/pkg/deps.bzl
@@ -14,40 +14,37 @@
 
 """WORKSPACE dependencies for rules_pkg/pkg."""
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", _http_archive = "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
-# @federation: BEGIN @rules_pkg
+def http_archive(**kwargs):
+    maybe(_http_archive, **kwargs)
 
 def rules_pkg_dependencies():
-    maybe(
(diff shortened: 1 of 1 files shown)


pkg/deps.bzl
```diff
@@ -14,40 +14,37 @@
 
 """WORKSPACE dependencies for rules_pkg/pkg."""
 
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:http.bzl", _http_archive = "http_archive")
 load("@bazel_tools//tools/build_defs/repo:utils.bzl", "maybe")
 
-# @federation: BEGIN @rules_pkg
+def http_archive(**kwargs):
```
aiuto, line 20: Really? I hate redefinitions of common things.
Of course, if I want to complain, http_archive should have had a "maybe" option from the beginning.

aiuto:
I think this is failing because the python update requires including platforms.

aiuto (approved):
I added platforms so visionos is there. But we should not really need this in the future.

PR #742, opened 2023-08-23
Start removing licenses(notice) and replacing with license()

aiuto:
`licenses()` does not do anything.


aiuto:
@teeler - can we do this?

aiuto:
Lowered in urgency. It's going to make my imports harder if I do this today.

aiuto:
friendly ping from reviewers.
license() is a no-op in Bazel. It is not used in Blaze when applicable_licenses is used.

Wyverald (approved):

cgrindel (approved):

aiuto:
The failing test is unrelated to this. It is that --noenable_bzlmod does not work with bazel at head, and bazelci is ignoreing the explicit bazel version.

PR #747, opened 2023-08-31
Document that package_dir also uses package_variables

flode:
The pkg_zip and pkg_tar rule already support variable substitution in their package_dir parameter, it was only documented for package_variables so far.


pkg/private/tar/tar.bzl
```diff
@@ -253,7 +253,8 @@ pkg_tar_impl = rule(
             doc = """(note: Use strip_prefix = "." to strip path to the package but preserve relative paths of sub directories beneath the package.)"""
         ),
         "package_dir": attr.string(
-            doc = """Prefix to be prepend to all paths written."""
+            doc = """Prefix to be prepend to all paths written.
+The name may contain variables in the forms {var} and $(var). The values for substitution are specified through `package_variables` or taken from [ctx.var](https://bazel.build/rules/lib/ctx#var).""",
```
aiuto, line 257: I would rather this points to the explanation under Common Attributes. 
Can you add an anchor tag (#package_file_name) to that file and point to that.

aiuto (changes requested):
Thanks for helping to make the docs better.

pkg/private/tar/tar.bzl
flode: Good idea, I also made the package_file_name and package_variables links

aiuto (approved):
Thanks!

PR #749, opened 2023-09-01
When pkg_tar.prefix_dir == base of symlink path, don't double-dip.

aiuto:
Correct for case where tar has prefix_dir and files/symlinks, but the user has already mapped the prefix_dir into those paths. Under the old behavior, prefix_dir was (incorrectly) not added to symlinks, so some users made the desired prefix part of the link.  Protecting them against the double inclusion of the prefix is not strictly pedantic, but is the least surprising behavior, as their intent was most likely to get what the old behavior would have achived.

Add tests for this, which required improving verify_archive_test.

Along the way, it seems that tree artifacts with internal symlinks are a problem. This might be a bazel bug.
https://github.com/bazelbuild/rules_pkg/issues/750 is a reminder to look in to this some day.


tests/tar/BUILD
```diff
@@ -341,6 +341,9 @@ py_test(
         "pkg_tar_test.py",
     ],
     data = [
+        ":test-pkg-tar-from-pkg-files-with-attributes",
```
aiuto, line 344: This resorting is buildifier

rbielak (approved):
lgtm

rbielak:
lgtm

sdtwigg (approved):

aiuto:
It's obsolete because we have the equivalent in the pkg_files rule. Rather
than maintain features
in tar only, but putting it in pkg_files we can reuse that in zip, deb, and
rpm.

The files and symlinks features in pkg_tar were never quite designed. They
were hacked in over time
at Google to solve some team's immediate needs.

coffinmatician:
That makes sense, thank you for helping me out. I'll try converting our internal usage in the near future and see how it goes.

Issue #579, opened 2022-05-06
Packaging binaries with runfiles tree


mvukov:
I have a simple use-case (I believe) where I want to package a binary with runfiles tree correctly set up inside the archive -- pretty much exactly what rules_docker does to package a binary in a Docker layer. pkg_zip flattens everything and that doesn't work for me (at least that was the case when I tried to use an older version of this repo).

So, I took what I need from app_layer from rules_docker and created rule [binary_pkg_tar](https://github.com/mvukov/rules_ros/blob/main/third_party/packaging.bzl#L306). This is interesting when one wants to cross-compile a binary, package it (with all deps, data), transfer to a target device, unpack and run the binary.

I just wanted to share this here in hope folks find it useful.

aiuto:
Can you describe this in more detail? Looking at the code it is hard to
understand the intent
of what you start with, what you end with, and if this belongs in rules_pkg
or rules_docker.
My current opinion is that rules docker should have all of the rules for
building docker images.
The job of rules_pkg is to make it easy for rules_docker to consume
pkg_files (or maybe pkg_tar)
directly. I don't know enough about building docker images to determine
requirements for that,
so you'll have to educate.

mvukov:
The goal of this little project was to create a rule that creates a valid runfiles folder for a packaged binary; simplest case: one wants to package a cc_binary/py_binary with a bit more complex deps, e.g. with data fields. At the time I created this (~1 year ago), pkg_tar would flatten the binary structure, which simply couldn't work -- in particular with py_binary targets which have external pip deps. On the other hand, rules_docker had exactly what I needed -- they basically copy+adjust runfiles folder of a binary target in `/app` folder inside an archive that represents a Docker layer. I took what I needed from rules_docker repo and created a new rule for my purpose. In essence, I don't create a Docker layer, but reuse the layer creation logic used to package binaries correctly in archives.

My starting point was [app_layer](https://github.com/bazelbuild/rules_docker/blob/76c708fc979c1bfb65b4db300c654be08f096874/lang/image.bzl#L289) that, among other things, creates an archive (layer) given the binary. The logic inside that rule (that I copied) eventually calls [build_tar.py](https://github.com/bazelbuild/rules_docker/commits/master/container/build_tar.py) script to build an archive.

[...]

PR #754, opened 2023-09-18
Bring tar runfiles up to feature parity with pkg_files.runfiles.

aiuto:
Bring tar runfiles up to feature parity with pkg_files.runfiles.
    
The bulk of the change is to have pkg_tar use `pkg_files%add_label_list` to process `srcs`.  That brings it in conformance with pkg_files and pkg_zip. This changes the behavior of runfiles to be useful, but it is a breaking change.
 
Fixes: #153
Fixes: #579
    
Other bits:
- do not fail verify_archive_test without a min_size or max_size
- change `data` file in the sample executable from BUILD to foo.cc so that it is easier to disambiguate from other files.
- fix bug in runfiles support where files were not added to file_deps list.
- fix bug in runfiles support where the runfiles base path was not computed correctly.
- buildifier to make the linters happy.

RELNOTES: Improved runfiles support in pkg_tar. Breaking change: we now require a Bazel that supports lambdas.

### DIFF
--- a/docs/latest.md
+++ b/docs/latest.md
@@ -99,6 +99,17 @@ pkg_deb(<a href="#pkg_deb-name">name</a>, <a href="#pkg_deb-architecture">archit
 </pre>
 
 
+    Create a Debian package.
+
+    This rule produces 2 artifacts: a .deb and a .changes file. The DefaultInfo will
+    include both. If you need downstream rule to specifically depend on only the .deb or
+    .changes file then you can use `filegroup` to select distinct output groups.
+
+    **OutputGroupInfo**
+    - `out` the Debian package or a symlink to the actual package.
+    - `deb` the package with any precise file name created with `package_file_name`.
+    - `changes` the .changes file.
+    
 
 **ATTRIBUTES**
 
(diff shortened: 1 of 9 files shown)


alexeagle:
@aiuto what's needed to make this draft Ready for review? We're eager to get this in to fix python images.

aiuto:
The refactoring PR #756 has to go in first. Then I need to rebase on that and add a legacy option.
It would also help if #749 were in too, then I can import to Google and test the legacy option.
About a day's work, when I get back from vacation.

dieortin:
@aiuto both PRs are merged, is there anything we could do to help with this?

aiuto:
I'm waiting on  https://github.com/bazelbuild/rules_pkg/pull/784 so CI can
be green again.

ConnorJLatham:
will this change be pulled into an official release sometime soon? looks like the latest was from may

aiuto:
Now that we are past the bazel 7 release and the holidays it might be time
for a new one.
It needs a little CI work first to make sure we are testing with bazel 6
(wihout bzlmod)
and 7 with both bzlmod and without.
Just a matter of an afternoon of grunt work.

PR #760, opened 2023-10-06
fix(pkg_tar): properly normalize paths for empty files

### LINKED ISSUE #758: pkg_tar: package_dir_file is not applied to pkg_mkdirs entries
When a directory is created using `pkg_mkdirs`, it seems that `pkg_tar` does not apply the prefix to it.

Application of the prefix happens through `normalize_path`:
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L65-L75
at least for files
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L89
and symlinks
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L178
as well as for tree artifacts (though without using reusing the function)
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L229-L236
however it doesn't happen for empty files
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L106-L139
or directories
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/tar/build_tar.py#L141-L152

While I can imagine that there might be someone out there who wants to be able to impose a prefix on the files and symlinks in their tarball without also applying it to empty files or directories, that behavior is certainly surprising, and moreover if one needs to put an empty file or directory in a location with the prefix, and you're prefix depends on a file input (as in `package_dir_file`), you are out of luck.

adam-azarchs:
The directory prefix attribute for a `pkg_tar` was not being honored for empty files or empty directories.

Closes #758

### DIFF
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -316,11 +316,11 @@ class TarFile(object):
     if entry.type == manifest.ENTRY_IS_LINK:
       self.add_link(entry.dest, entry.src, **attrs)
     elif entry.type == manifest.ENTRY_IS_DIR:
-      self.add_empty_dir(entry.dest, **attrs)
+      self.add_empty_dir(self.normalize_path(entry.dest), **attrs)
     elif entry.type == manifest.ENTRY_IS_TREE:
       self.add_tree(entry.src, entry.dest, **attrs)
     elif entry.type == manifest.ENTRY_IS_EMPTY_FILE:
-      self.add_empty_file(entry.dest, **attrs)
+      self.add_empty_file(self.normalize_path(entry.dest), **attrs)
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
thanks!

PR #761, opened 2023-10-06
pkg_deb: allow data.tar.zst

### LINKED ISSUE #759: pkg_deb: permit data.tar.zst
`dpkg` has supported zstd-compressed data [since version 1.21.18](https://launchpad.net/debian/+source/dpkg/1.21.18) (and at least in ubuntu 22.04, dpkg 1.21.1 seems to also support it, though the feature is undocumented; I think the support is implicit through [tar 1.31+](https://savannah.gnu.org/news/?id=9308)).  While `pkg_tar` does [not yet directly support making zstd-compressed tarballs](https://github.com/bazelbuild/rules_pkg/issues/481), aside from the possibility of constructing such tarballs by other means, one can still build them by supplying a custom `compressor`.

`pkg_deb` does not currently allow using a `data.tar.zst`, in two places:
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/deb/deb.bzl#L19
disallows supplying a `tar.zst`.  If one fixes that, there's also
https://github.com/bazelbuild/rules_pkg/blob/ff60b34e037313927b9067dd883dd82e1fbc211e/pkg/private/deb/make_deb.py#L240-L241
which will strip off the `.zst` suffix, rendering the resulting `.deb` unusable.

Worth noting, every `.deb` archive that I've spot-checked in the Ubuntu jammy (22.04) repository has used `data.tar.zst`, which I'd say makes this a pretty main-stream feature.

adam-azarchs:
dpkg has officially supported zstandard compression for the data tarball since version 1.21.18, and despite shipping with dpkg version 1.21.1, most .deb packages for Ubuntu 22.04 are also in zst format.

While this PR does not do anything to make it easier to create zstd-compressed tarballs, it does make it possible to use them when creating .deb packages.

Closes #759

### DIFF
--- a/pkg/private/deb/deb.bzl
+++ b/pkg/private/deb/deb.bzl
@@ -16,7 +16,7 @@
 load("//pkg:providers.bzl", "PackageVariablesInfo")
 load("//pkg/private:util.bzl", "setup_output_files")
 
-_tar_filetype = [".tar", ".tar.gz", ".tgz", ".tar.bz2", "tar.xz"]
+_tar_filetype = [".tar", ".tar.gz", ".tgz", ".tar.bz2", "tar.xz", "tar.zst"]
 
 def _pkg_deb_impl(ctx):
     """The implementation for the pkg_deb rule."""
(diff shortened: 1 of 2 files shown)


aiuto:
I'll be OOO until Oct. 14. I can look then.

aiuto (approved):

PR #764, opened 2023-10-24
Use rules_cc to try to clean up CI runs which are failing on gcc

aiuto:


aiuto:
This does not fix all the problems, but it does fix one example. I want to merge in spite of CI not being clean.
The other CI failures look like repository caching problems.

cgrindel (approved):

PR #766, opened 2023-10-26
maybe get bzlmod working

aiuto:


aiuto:
Review note.  CI is broken for entirely different reasons. I've spent a day looking at it, but not quite figured it out.

.bazelci/tests.yml
```diff
@@ -89,15 +89,20 @@ tasks:
...
+    platform: ubuntu1804
+    build_flags:
+      - "--enable_bzlmod"
+    <<: *common
+    <<: *rolling
+    <<: *default_tests
   rolling_macos:
-    name: rolling_macos
+    name: last_green_macos
     # It seems there is no rolling Bazel for macos.
```
alexeagle, line 102: Double check this maybe? I'm pretty sure I got a rolling macOS last week.

MODULE.bazel
```diff
@@ -7,10 +7,15 @@ module(
...
 bazel_dep(name = "bazel_skylib", version = "1.2.0")
-bazel_dep(name = "rules_python", version = "0.10.2")
 
 # Only for development
 bazel_dep(name = "platforms", version = "0.0.5", dev_dependency = True)
 bazel_dep(name = "stardoc", version = "0.5.3", dev_dependency = True)
 bazel_dep(name = "rules_cc", version = "0.0.9", dev_dependency = True)
+
+# Find the system rpmbuild if one is available.
+find_rpm = use_extension("//toolchains/rpm:rpmbuild_configure.bzl", "find_system_rpmbuild_bzlmod")
```
alexeagle, line 19: Should this be a dev dependency?

toolchains/rpm/rpmbuild_configure.bzl
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 """Repository rule to autoconfigure a toolchain using the system rpmbuild."""
 
+# NOTE:
```
alexeagle, line 16: What's the note?

alexeagle (commented):
You could rebase after the runfiles change landed to make the delta smaller?

MODULE.bazel
aiuto: Yes. Changed.  But perhaps this exposes a bzlmod bug.
I can't change use_repo to dev_dependency.
So perhaps this won't work for downstream dependencies. I can look into that in a second pass. Right now I only want to get a working CI run.

Wyverald: >I can't change use_repo to dev_dependency.

`use_repo` on a `use_extension(dev_dependency=True)` behaves the way you'd expect it to.

README.md
```diff
@@ -23,6 +23,49 @@ As of Bazel 4.x, Bazel uses this rule set for packaging its distribution. Bazel
...
+    verbose = False,
+)
+```
+
+## MODULE.bazel setup
+
+```
+bazel_dep(name = "rules_pkg", version = "0.0.10")
+```
+To use `pkg_rpm()`, you must provide a copy of `rpmbuild`. You can use the
```
Wyverald, line 61: just a note, this is the perfect candidate to be split into a separate module (`rules_pkg_rpm` or some such).

tests/mappings/manifest_test_lib.py
```diff
@@ -38,14 +38,17 @@ def assertManifestsMatch(self, expected_path, got_path):
...
     got_dict = {x['dest']: x for x in got}
-    # self.assertEqual(expected_dict, got_dict)
 
     ok = True
     expected_dests = set(expected_dict.keys())
     got_dests = set(got_dict.keys())
     for dest, what in expected_dict.items():
       got = got_dict.get(dest)
       if got:
+        # bazlmod mode changes root to @@//, but history says @//
```
Wyverald, line 48: 1. nit -- "Bzlmod" instead of "bazlmod"
2. just for my information, what does "history says @//" mean?

.bazelci/tests.yml
aiuto: It does work now.

README.md
aiuto: Yes. I may end up doing that. It's going to be an interesting packaging setup because I need to depend on the module for the CI, but I don't want it in the MODULE.bazel file we ship for rules_pkg. But, that's a problem for next week. Right now I just want to get a bzlmod CI run working.

tests/mappings/manifest_test_lib.py
aiuto: changed

Wyverald (approved):

MODULE.bazel
aiuto: I'll take that to mean that it won't fail, but it's not what I expect.
My expectation is that  `name = use_extension(..., dev_dependency=True)` returns a callable if you are in dev mode and nothing in non-dev mode. That is, I expect it to blow up unless `use_repo` is also dev only, rather than return a callable that does nothing. 

But, that is irrelevant to this PR.

PR #768, opened 2023-10-27
Use runfiles from rules_python

aiuto:
Based on comments in https://github.com/bazelbuild/bazel/issues/18128

This does not solve some of the bzlmod strangeness, but it is a first step.


alexeagle (approved):

aiuto:
The checks that are failing are known.

PR #763, opened 2023-10-11
pkg_rpm: Specify _topdir for install stanza

kellyma2:
If we don't specify the _topdir as part of the install-from path we can't seem to find the correct files to install when using a set of rules like:
```
pkg_filegroup(
    name = "my_filegroup",
    srcs = [
      "//some/label:foo",
    ],
    prefixs = "/usr/bin",
)

pkg_rpm(
    name = "rpm",
    ...
    srcs = [
        ":my_filegroup",
    ],
)
```
Adding `%{{_topdir}/BUILD/` as a prefix to the source path seems to solve this problem.

Issue #662, opened 2023-01-18
feature request: build rpms without rpmbuild


adam-azarchs:
It would be nice to be able to build rpms on systems which don't have rpmbuild installed (e.g. a debian-based build machine).  At the very least, supporting `rpmbuild` as a label would be nice, so one could download it or build it from source.  But for many simple cases, [`rpmpack`](https://github.com/google/rpmpack#usage-in-the-bazel-build-system-pkg_tar2rpm) could be used instead (their readme includes instructions for doing this with `tar2rpm`, but that is arguably inefficient compared to a more direct approach, and it would be nice to have a more unified API for building `.deb` and `.rpm` together, including for example having support for `version_file`).

nacl:
I'm hesitant to support this (and @aiuto can answer more authoritatively), but to answer some of your points:

> At the very least, supporting rpmbuild as a label would be nice, so one could download it or build it from source.

`rpmbuild` can be specified as a toolchain today.  We use this functionality frequently (https://github.com/bazelbuild/rules_pkg/pull/254) to use a custom-built `rpmbuild` setup independent of the host OS.  You can find more information about toolchains [here](https://bazel.build/extending/toolchains) if you didn't already know.

Also FWIW, there are Debian packages for the RPM tooling available, [at least for Debian itself](https://packages.debian.org/search?keywords=rpm).

> But for many simple cases, [rpmpack](https://github.com/google/rpmpack#usage-in-the-bazel-build-system-pkg_tar2rpm) could be used instead

While there plenty of reasons not to use `rpmbuild` (your points and the inconveniences mentioned in https://github.com/bazelbuild/rules_pkg/issues/661 among them), the main thing that `rpmbuild` does provide is compatibility.  Empirically, RPM packages don't really conform to a specification other than what `rpmbuild` happens to produce.

> ...and it would be nice to have a more unified API for building .deb and .rpm together, including for example having support for version_file

This is a separate feature, IMO to be tracked separately.  FWIW, it should be relatively easy to implement this if you are interested.

aiuto:
>Empirically, RPM packages don't really conform to a specification other than what rpmbuild happens to produce.

Bingo! That's what makes this hard to support. Any replacement is subject to the problem of tracking rpmbuild changes.
[...]

PR #770, opened 2023-10-31
testing #763

aiuto:
@kellyma2

I'm baffled.  I took the example you had and added a python version of the binary.

python alone works fine. C++ fails with a permission denied error, not file not found.

I added some debugging "ls -l"s into the install stanza and the files seem to be in the source.  I can't spend a lot of time on it this week, but it seems that the permission problem is while calling /usr/bin/strip.  It might be that the file is mode o555 rather than o755. But.. how has this been working for others?

### DIFF
--- a/examples/MODULE.bazel
+++ b/examples/MODULE.bazel
@@ -1,0 +1,22 @@
+module(
+    name = "rules_pkg_examples",
+    version = "0.0.0",  # never ships
+    compatibility_level = 1,
+)
+
+local_path_override(
+    module_name = "rules_pkg",
+    path = "..",
+)
+
+bazel_dep(name = "rules_python", version = "0.24.0")
+bazel_dep(name = "rules_cc", version = "0.0.2")
+
+# Find the system rpmbuild if one is available.
+find_rpm = use_extension(
+    "//toolchains/rpm:rpmbuild_configure.bzl",
+    "find_system_rpmbuild_bzlmod",
+    dev_dependency = True
+)
+use_repo(find_rpm, "rules_pkg_rpmbuild")
+register_toolchains("@rules_pkg_rpmbuild//:all", dev_dependency = True)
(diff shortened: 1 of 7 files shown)


kellyma2:
I'll take a bit closer look at this either tomorrow or Thursday to see if I can narrow down what's going on for me a bit.  Are you able to share a bit more detail about the environment in which you're testing this?

aiuto:
Debian 6.5.3, bazel 6.4.0
RPM version 4.18.0

And, now that I list the rpmbuild version, I remember that rpmbuild has some breaking changes between versions. That may be a clue.  What do you have?

kellyma2:
> Debian 6.5.3, bazel 6.4.0 RPM version 4.18.0
> 
> And, now that I list the rpmbuild version, I remember that rpmbuild has some breaking changes between versions. That may be a clue. What do you have?

bazel 6.3.2, rpm 4.14.3, custom-ish Linux based on CentOS.  I'll test against vanilla Fedora/CentOS tomorrow with newer rpmbuild and see what I get.

aiuto:
@nacl 
IIRC, there were some changes between recent rpmbuild versions which we had to work around to achieve backwards compatibility. One was around [SOURCE_DATE_EPOCH at 4.14.0](https://github.com/bazelbuild/rules_pkg/blob/8e5570cf4423c097847172d3e32cd7d89d4d23bb/tests/rpm/source_date_epoch/rpm_contents_vs_manifest_test.py#L45)

[This line](https://github.com/bazelbuild/rules_pkg/blob/8e5570cf4423c097847172d3e32cd7d89d4d23bb/toolchains/rpm/rpmbuild_configure.bzl#L46) suggests that something on either side of 4.16 required putting the version in the toolchain declaration.  And lo, you are 4.14 and I am 4.18, so that's a lead.

Sometimes I wonder if we should just have done #662.

PR #774, opened 2023-11-06
chore: add a pre-commit hook config

alexeagle:
This matches the Bazel recommendation from https://github.com/bazel-contrib/rules-template for configuring an optional, client-side automation for formatting BUILD/bzl files before sending a PR.

It has been found to save a bunch of time for contributors who otherwise trip over the server-side CI check and have to do an extra iteration to get their PR green.

I imagine Googlers are already relying on some Piper feature to do the equivalent when they work on this code, so they haven't noticed the deficiency.


aiuto (commented):
LGTM

PR #777, opened 2023-11-06
doc: Fixup external manual references

kellyma2:
The original links for the RPM manual that were being referred to appear to have moved.  This update the links to point at the correct place.


alexeagle (approved):
thanks!

PR #778, opened 2023-11-06
Add support for `Obsoletes` tag in RPM definition

kellyma2:
Currently there doesn't appear to be any support `Obsoletes` in our RPM structure.  This change introduces the support for the `Obsoletes` tag in the RPM specfile.  Unlikely `Requires`, there doesn't appear to be a context-specific `Obsoletes`, so this is not added here.


aiuto (approved):
( will merge in a few days.

PR #779, opened 2023-11-07
Don't load cc toolchain from rules_cc

meteorcloudy:
The cc toolchain autoconfig in rules_cc is out of sync from Bazel and doesn't work with VS2022.

Related: https://github.com/bazelbuild/continuous-integration/pull/1770

### DIFF
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -66,7 +66,7 @@ ubuntu1804: &ubuntu
     - "//distro:relnotes"
     - "//doc_build:*"
 
-centos7: &centos
+centos7_java11_devtoolset10: &centos
   platform: centos7
   <<: *common
   <<: *default_tests_with_rpm
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -66,8 +66,8 @@ ubuntu1804: &ubuntu
     - "//distro:relnotes"
     - "//doc_build:*"
 
-centos7_java11_devtoolset10: &centos
-  platform: centos7
+centos7: &centos
+  platform: centos7_java11_devtoolset10
   <<: *common
   <<: *default_tests_with_rpm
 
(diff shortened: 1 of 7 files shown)


WORKSPACE
```diff
@@ -82,5 +82,3 @@ http_archive(
 load("@rules_cc//cc:repositories.bzl", "rules_cc_dependencies", "rules_cc_toolchains")
```
aiuto, line 82: delete from this line too.

aiuto (commented):
examples/rich_structure/WORKSPACE needs the same change.

But the centos build fails now. with
  | gcc: error: unrecognized command line option '-std=c++14'

Did rules CC change in a way to believe it is always above 14?

WORKSPACE
meteorcloudy: Done

meteorcloudy:
> Did rules CC change in a way to believe it is always above 14?

Yes, with Bazel@HEAD, we already switch the default C++ standard to 14: https://github.com/bazelbuild/bazel/pull/18280

We should switch to use centos7_java11_devtoolset10 or just drop this platform.

.bazelci/tests.yml
```diff
@@ -66,7 +66,7 @@ ubuntu1804: &ubuntu
     - "//distro:relnotes"
     - "//doc_build:*"
 
-centos7: &centos
+centos7_java11_devtoolset10: &centos
```
aiuto, line 69: Tests are still failing with the C++ std error.
Should the change be to the `platform`?

aiuto:
> We should switch to use centos7_java11_devtoolset10 or just drop this platform.
We can't drop centos.  We need it for the RPM builder tests.

.bazelci/tests.yml
meteorcloudy: You are absolute correct!

meteorcloudy:
> We can't drop centos. We need it for the RPM builder tests.

Got it, added in back in `.bcr/presubmit.yml`

alexeagle (approved):
Thanks Yun!

aiuto (approved):

PR #782, opened 2023-11-11
fix the imports on the filter directory test

aiuto:
The old code was tricky and brittle.


aiuto:
The failing tests for unrelated reasons.

alexeagle (approved):
Nice simplification

PR #784, opened 2023-11-11
CI fixes

aiuto:
These are just to stop the spurious CI failures at head.
The right solution is to go back and have everything build uniformly with bzlmod or without it.
- Get experimental git toolchain working enough with bzlmod so we can use it. But yeech. It really needs a bzlmod conversion. I would rather someone else build it.
- Do not run the doc_build with bzlmod. It is impossible to have the sources work for both bzlmod and non_bzlmod because the module name changed. Or maybe it is, but I don't want to solve that right now.

### DIFF
--- a/WORKSPACE.bzlmod
+++ b/WORKSPACE.bzlmod
@@ -14,7 +14,7 @@ local_repository(
 #    verbose = False,
 #)
 
- Needed for making our release notes
+# Needed for making our release notes
 load("//toolchains/git:git_configure.bzl", "experimental_find_system_git")
 
 experimental_find_system_git(
--- a/WORKSPACE.bzlmod
+++ b/WORKSPACE.bzlmod
@@ -14,10 +14,10 @@ local_repository(
 #    verbose = False,
 #)
 
-# Needed for making our release notes
-#load("//toolchains/git:git_configure.bzl", "experimental_find_system_git")
-#
-#experimental_find_system_git(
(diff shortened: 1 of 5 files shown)


WORKSPACE.bzlmod
```diff
@@ -15,9 +15,13 @@ local_repository(
...
 # Needed for making our release notes
-#load("//toolchains/git:git_configure.bzl", "experimental_find_system_git")
-#
-#experimental_find_system_git(
-#    name = "rules_pkg_git",
-#    verbose = False,
-#)
+load("//toolchains/git:git_configure.bzl", "experimental_find_system_git_bzlmod")
+
+experimental_find_system_git_bzlmod(
```
alexeagle, line 20: Is it possible to stop depending on git instead?

aiuto: I'm trying build tooling that helps create release notes. I have not abandoned that idea yet. I may eventually, or move it to another project.  OTOH, keeping it here as an optional module might be the right thing because people so often want to get git-hash into the names of their release artifacts.

meteorcloudy (approved):

WORKSPACE.bzlmod
meteorcloudy: Do we really have to do this through a toolchain? I would also prefer a simpler solution if possible.

PR #787, opened 2023-11-22
Allow substitution of user-defined variables in RPM preamble

kellyma2:
It's desirable to be able to parameterize some variables in the preamble such as architecture when RPM packages.  This change enables variable substitution in the preamble section so that the values may be injected in this fashion in lieu of only using statically defined values.

### DIFF
--- a/pkg/private/util.bzl
+++ b/pkg/private/util.bzl
@@ -82,5 +82,14 @@ def substitute_package_variables(ctx, attribute_value):
 
     # Map $(var) to {x} and then use format for substitution.
     # This is brittle and I hate it. We should have template substitution
-    # in the Starlark runtime.
-    return attribute_value.replace("$(", "{").replace(")", "}").format(**vars)
+    # in the Starlark runtime.  This loop compensates for mismatched counts
+    # of $(foo) so that we don't try replace things like (bar) because we
+    # have no regex matching
+    for _ in range(attribute_value.count("$(")):
+        if attribute_value.find(")") == -1:
+            fail("mismatched variable declaration")
+
+        attribute_value = attribute_value.replace("$(", "{", 1)
+        attribute_value = attribute_value.replace(")", "}", 1)
+
+    return attribute_value.format(**vars)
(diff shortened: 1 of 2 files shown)


cgrindel (commented):
Overall, the changes look reasonable to me. Could you ping me when CI is green?

kellyma2:
Will do!

pkg/private/util.bzl
```diff
@@ -82,5 +82,14 @@ def substitute_package_variables(ctx, attribute_value):
 
     # Map $(var) to {x} and then use format for substitution.
     # This is brittle and I hate it. We should have template substitution
-    # in the Starlark runtime.
-    return attribute_value.replace("$(", "{").replace(")", "}").format(**vars)
+    # in the Starlark runtime.  This loop compensates for mismatched counts
+    # of $(foo) so that we don't try replace things like (bar) because we
+    # have no regex matching
+    for _ in range(attribute_value.count("$(")):
+        if attribute_value.find(")") == -1:
```
aiuto, line 89: This looks like it forces you to use the $(foo) format rather than ${var}.

aiuto (commented):
The PR description doesn't seem to match the code at all.  It looks like the code just adds some safety against someone mis-typing  a variable like  $(foo} something (bar) .

kellyma2:
> The PR description doesn't seem to match the code at all. It looks like the code just adds some safety against someone mis-typing a variable like $(foo} something (bar) .

The initial PR + first commit match up.  When the CI job ran to test the initial changes, it hit the former case that the second commit fixes.

pkg/private/util.bzl
kellyma2: No, it's doing something slightly different.  The original line:

```
attribute_value.replace("$(", "{").replace(")", "}").format(**vars)
```

replaces every instance of "$(" with "{" and ")" with "}" in a single shot.  The problem with this is that we'll do things like transform `foo)` into `foo}` or `(foo)` into `{foo}`.  The former is probably (mostly) harmless, but the latter results in `str.format(**vars)` thinking that there's an additional format var in there while there won't be a matching one in `vars`, generating a bit of a weird error.  Normally I'd have dealt with this using regex matching, but that isn't an option here :)

aiuto (approved):

PR #788, opened 2023-11-22
Only inject pre and post scriptlets when provided

kellyma2:
Currently if the consumer doesn't provide a value for the pre or post scriplets we inject an empty scriptlet which has the effect of constructing an RPM that appears to have these scriplets (although they're empty).  Instead, we likely want to not provide them at all in these cases.


aiuto (approved):
I'm going to presume this is reasonable behavior.

PR #789, opened 2023-11-28
pkg_tar: load attributes for TreeArtifacts

novas0x2a:
This change teaches pkg_tar to reinvoke file_attributes for each individual file in a TreeArtifact; this function is used to load the attributes passed via pkg_tar arguments (like `mode`). Before this change, there's no way to set individual attributes for files in a TreeArtifact.


pkg/private/tar/build_tar.py
```diff
@@ -273,31 +273,48 @@ def add_tree(self, tree_top, destpath, mode=None, ids=None, names=None):
           to_write[dest_dir + file] = content_path
 
     for path in sorted(to_write.keys()):
+      if file_attributes:
```
aiuto, line 276: Isn't this always true?

pkg/private/tar/build_tar.py
```diff
@@ -434,11 +451,11 @@ def main():
       compressor = options.compressor,
       default_mtime=default_mtime) as output:
 
-    def file_attributes(filename):
+    def file_attributes(filename, with_default_mode=True):
```
aiuto, line 454: Why is with_default_mode needed?  Having it, and only for mode rather than the rest makes this method a little more complex to understand.

pkg/private/tar/build_tar.py
```diff
@@ -273,31 +273,48 @@ def add_tree(self, tree_top, destpath, mode=None, ids=None, names=None):
...
+          use_names = names
+          use_mode = mode
+
       content_path = to_write[path]
       if not content_path:
         # This is an intermediate directory. Bazel has no API to specify modes
         # for this, so the least surprising thing we can do is make it the
-        # canonical rwxr-xr-x
+        # canonical rwxr-xr-x unless we've gotten a custom override
+        if use_mode is None:
```
aiuto, line 293: It seems like just having 0o755 be the default mode would eliminate the need for the test here.

pkg/private/tar/build_tar.py
novas0x2a: yeah, at the moment; I copied how file_attributes was [used elsewhere in the script](https://github.com/bazelbuild/rules_pkg/blob/ed48f1c901a290e76815ede2da85133172d62218/pkg/private/tar/build_tar.py#L322-L325) because I didn't understand the context about why the script was already checking for it, and I figured it was safer to do it the same way.

pkg/private/tar/build_tar.py
novas0x2a: Without `with_default_mode`, I can't tell the difference between a value that's intentionally set and a value that's falling back to the default. This is important because file_attributes falls back to the default mode (as passed by the end-user caller); it's pretty common to set the default mode to something like 0644, which is a fine default for files but a terrible default for directories. `with_default_mode` allows me to have different defaults for directories and files (without a breaking change)

novas0x2a: I said "without a breaking change" but I guess I should note that there's a non-zero chance of a breaking change in this; before, if someone set `modes` to a file or a directory in a TreeArtifact, it would have been silently ignored before and now it would be honored. It's somewhat unlikely that anyone would actually be bitten by this (I noticed the bug when I tried to do exactly this, and noticed it didn't work) but it'd probably need to be called out.

novas0x2a:
Any updates?

aiuto:
Sorry for the last response. I was unavailable from Jan 10 until about now. 
Catching up on things during March.

PR #792, opened 2023-12-12
Explicitly set `%{_builddir}` macro

kellyma2:
rules_pkg has baked in the assumption that the value of `%{_builddir}` is going to be `%{_topdir}/BUILD` which is where rpmbuild will `cd` to when being run.  When using the built in system rpmbuild in a situation where the value of `%{_builddir}` has been overriden with a custom local macro, this assumption may be broken which will result in internal rpmbuild failures.

Because we make this assumption about the value of the macro, we can instead be explicit about what we want the value of `%{_builddir}` to be so as to avoid this problem altogether.

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -357,6 +357,7 @@ class RpmBuilder(object):
     args += [
         '--define', '_topdir %s' % dirname,
         '--define', '_tmppath %s/TMP' % dirname,
+        '--define', '_builddir %s/BUILD' % dirname,
         '--bb',
         '--buildroot=%s' % buildroot,
     ]  # yapf: disable


cgrindel (approved):
LGTM. The CI failures don't look related to this. However, I have not dug into them.

@aiuto Is there a way to re-run the CI for this PR?

aiuto:
The failures are all stardoc.  I"m going to force merge this and then disable the doc building in CI.
I'll do that manually.

aiuto:
Sigh... It/s not just stardoc.  bzlmod at bazel 7.0.0. seems to mess with some things about local_repository.
No time to explain more on that.

PR #796, opened 2023-12-19
Fix some small issues with rpm support

manuelnaranjo:
I started trying to update rules_pkg to build some of our RPMs at booking, and noticed a few issues.

This are small changes, and even though they're not related, I didn't thought it was a good idea to
have them in separate MRs


pkg/make_rpm.py
```diff
@@ -109,7 +109,7 @@ def CopyAndRewrite(input_file, output_file, replacements=None, template_replacem
   """
 
   with open(output_file, 'w') as output:
-    for line in fileinput.input(input_file):
+    for line in fileinput.input(input_file, encoding='utf-8'):
```
cgrindel, line 112: This seems like a reasonable value for encoding. I had a quick look at [the doc for fileinput](https://docs.python.org/3/library/fileinput.html) to understand what it selected for encoding when it was unspecified. Unfortunately, it does not say.

cgrindel (approved):
LGTM. The `doc_build` CI failure does not seem to be related to the PR. I am unsure if this PR is the cause for the `examples_naming` errors.

pkg/make_rpm.py
manuelnaranjo: Yeah I dived into 3 python modules source code and gave up, I have the feeling it goes down all the way to native code to decide, because I'm using the relocatable python toolchain and getting different results on CentOS7 than CentOS Stream 8 and 9 for the same source code

aiuto: Yeah.  This is probably not going to be as portable as it should be.
Can you describe the problems you are having which lead to this change.
And a test case really needs to be provided.

aiuto (changes requested):

pkg/make_rpm.py
manuelnaranjo: Literally was getting ASCII being picked as the default encoder/decoder even with a py3.11 toolchain, I can try coming up with a test, I gave it a few minutes of research but I couldn't figure out how to influence it by passing different env variables into python

manuelnaranjo: Mhh maybe the locale of my system is corrupted, I'll debug further tomorrow https://docs.python.org/3.11/glossary.html#term-locale-encoding sheds some light, I see rules_pkg set a few env variables to utf-8 but it's not getting the entire desired effect.
Also I had to bump rules_py through bzlmod, maybe I got in a state that the bug surfaces, what's weird is that it only happens on CO7 and not our COS8/9 container images

manuelnaranjo: @aiuto so I did some research, I have kind of a repro setup, which basically has a py_binary and a genrule, all the binary does is print the locale.getlocale and compare it against UTF-8, if I run the binary with bazel run on CO7 and COS9 I get the expected UTF-8, if I go through the genrule then it works on CO7 but it fails in COS9. What I noticed is that when I print the env variables in COS9 I see `LC_CLANG=C.UTF-8` as an environmental variable, if I set LC_CLANG=en_US.UTF-8 through action_env then I get the expected results.

As part of https://github.com/bazelbuild/bazel/issues/7370 you added LC_CLANG=UTF-8 as a default env for make_rpm, I'll keep digging, but it looks like my image is badly configured, shall I revert this part of the change then?

manuelnaranjo: Ok more findings https://bugzilla.redhat.com/show_bug.cgi?id=902094 so COS8 added C.UTF-8 it started with CentOS Stream 8, I find other similar issues https://gitlab.haskell.org/ghc/ghc/-/issues/18607, and I see that bazel it self sets en_US.UTF-8 in the env variables for the javac calls https://github.com/bazelbuild/bazel/blob/c7a47846f12c8ad24b53cfc6dbcd99c701c293d9/src/main/java/com/google/devtools/build/lib/rules/java/JavaCompileActionBuilder.java#L57, does it make sense that I create a patch that sets the right env variable instead? Not sure how we would test this though, it requires a customized container image

manuelnaranjo: @aiuto I changed the way to deal with this, I figured out that rules_pkg doesn't respect LC_CLANG from action_env, I made some changes to respect this value and only if not passed by the rule invocation to use the defaults. 

What do you think?

aiuto:
Friendly ping.  Sorry for the slow ping. I was essentially away all of January and February.

PR #799, opened 2023-12-21
Add %posttrans scriptlet to RPM package

k0walik:
RPM package specification defines `%posttrans` scriptlet, which can be particularly useful when upgrading a package. We should add it to the `pkg_rpm` build rule.

Reference: https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering


cgrindel:
@k0walik #797 just merged disabling the doc builds. I believe that those are causing all/most of the failures in your CI. Could you update your branch with the latest?

cgrindel (commented):
Overall, LGTM. A couple of comments:
- I did not notice a test leveraging `postrans_scriptlet_file`. Could you add one?
- Could you update your branch with the latest to address the doc_build failures?

k0walik:
> * I did not notice a test leveraging `postrans_scriptlet_file`. Could you add one?

Sure! I modified `test/rpm/BUILD` to test both `xxx_scriptlet` and `xxx_scriptlet_file` attributes - it looks like file variant was not tested at all.

> * Could you update your branch with the latest to address the doc_build failures?

Done

cgrindel (approved):
LGTM. I won't merge it yet, so that @aiuto can take a look.

aiuto (approved):
This looks like it should work.
Thanks for improving the testing

PR #804, opened 2024-01-02
Explicitly set the FILE bit in zip external attributes.

### LINKED ISSUE #802: Zip file created with a default (file) mode causes corrupted Azure function
Disclaimer: I'm not sure whether this is a bug in rules_pkg or really a problem in Azure. However I thought I better file it regardless in case others stumble over this since the "workaround" is easy. So feel free to close.

We're using the packaging rules to package up Azure Function Apps. The mode used to create a `zipfile.ZipInfo` object in the zip file created, causes a broken Azure Function App deployment. The main problem seems to be the file mode which differs significantly when used `zipfile.ZipInfo.from_file`:

## Env details
OS: GNU/Linux - Ubuntu 22.04 running in WSL2
Python: 3.11.0

## Problem description
The `filemode` set with the current code:
```Python
> <ZipInfo filename='dist/main.js' filemode='?rw-r--r--' file_size=3408 compress_size=0>
> entry_info.external_attr
27525120
> oct(entry_info.external_attr >> 16)
'0o644'
```
The `filemode` when the info is constructed with `zipfile.ZipInfo.from_file`:
```Python
> <ZipInfo filename='dist/main.js' filemode='-rw-r--r--' file_size=3408 compress_size=0>
> oct(info.external_attr >> 16)
'0o100644'
```
You'll notice the `?` at the beginning of the zipinfo for the file if constructed with the current default mode.

## Reproducer

I can reproduce this 100% with access to deploying a function app. I see if I can file a bug somewhere in Azure too. A minimal reproducer is to create a zip file with the `make_zipinfo` method from `build_zip.py` using a 16bit? file mode (0o644).

## Workaround
The workaround I currently use is to set an explicit file mode:

```
pkg_files(
    name = "pkg_zip_contents",
    srcs = [
        ":zip_contents",
    ],
    attributes = pkg_attributes(
        mode = '100644',
    ),
)
```

It seems using a 32bit file mode (sorry only stumbled over it in a [Stackoverflow post about git file modes](https://unix.stackexchange.com/questions/450480/file-permission-with-six-octal-digits-in-git-what-does-it-mean) is not causing this confusing behaviour in deployment.

aiuto:
This should not be needed, but it seems to be for some Azure users.

Fixes #802

### DIFF
--- a/pkg/private/zip/build_zip.py
+++ b/pkg/private/zip/build_zip.py
@@ -169,7 +169,7 @@ class ZipWriter(object):
     if entry_type == manifest.ENTRY_IS_FILE:
       entry_info.compress_type = self.compression_type
       # Using utf-8 for the file names is for python <3.7 compatibility.
-      entry_info.external_attr |= UNIX_FILE_BIT << 16  #  DO WE NEED AN MSDOS_DIR_BIT?
+      entry_info.external_attr |= UNIX_FILE_BIT << 16
       with open(src.encode('utf-8'), 'rb') as src_content:
         self.writestr(entry_info, src_content.read(), compresslevel=self.compression_level)
     elif entry_type == manifest.ENTRY_IS_DIR:
--- a/pkg/private/zip/build_zip.py
+++ b/pkg/private/zip/build_zip.py
@@ -26,9 +26,10 @@ from pkg.private import manifest
 ZIP_EPOCH = 315532800
 
 # Unix dir bit and Windows dir bit. Magic from zip spec
-UNIX_DIR_BIT = 0o40000
-MSDOS_DIR_BIT = 0x10
+UNIX_FILE_BIT =    0o100000
 UNIX_SYMLINK_BIT = 0o120000
+UNIX_DIR_BIT  =    0o040000
+MSDOS_DIR_BIT = 0x10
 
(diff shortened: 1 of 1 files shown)


cgrindel (approved):
Do we have any tests that can be augmented to check this?

aiuto:
No.  It seems to be Azure only.   I need @romanofski  to verify this does the right thing.
If it does, then perhaps I can figure out a windows test.

pkg/private/zip/build_zip.py
```diff
@@ -168,6 +169,7 @@ def add_manifest_entry(self, entry):
     if entry_type == manifest.ENTRY_IS_FILE:
       entry_info.compress_type = self.compression_type
       # Using utf-8 for the file names is for python <3.7 compatibility.
+      entry_info.external_attr |= UNIX_FILE_BIT << 16  #  DO WE NEED AN MSDOS_DIR_BIT?
```
aiuto, line 172: @romanofski  
I'm not sure this is sufficient for your case on Azure. Can you patch in this change and try it?  
Mu suspiciaion is that we need another bit.

romanofski: Yeah will try out. Thanks mate for the quick patch!

PR #812, opened 2024-01-16
Apply tar remap_paths to runfiles full paths

### LINKED ISSUE #790: `remap_paths` from `pkg_tar` should apply to runfiles
The new implementation of `include_runfiles` for `pkg_tar` performs path remapping only on the binaries, and then roots the runfiles in `<remapped_binary_path>.runfiles`. The runfiles themselves are not remapped, so placing them in a different location is not possible. 

My use case for this is placing C++ .so dependencies in a different path (`/lib`) This would be easy to achieve if remapping applied to runfiles. 

My proposed change would be applying remapping to individual runfiles, instead of to the root folder only. I have it working in https://github.com/bazelbuild/rules_pkg/compare/main...dieortin:rules_pkg:tar-runfile-mapping and would be happy to work on a PR if you think this makes sense.

dieortin:
This modifies the way `remap_paths` works on `pkg_tar` when using `include_runfiles` so that remapping is performed on the full path of each runfile, instead of just in the base folder (`foo.runfiles`)

This should be backwards compatible with the current behavior, and only affect those that specify runfiles paths in their mappings. For example, for the following path `foo.runfiles/some/folder/file.txt`:
```starlark
remap_paths: {
    "/foo": "/bar", # Still yields the same result as before (bar.runfiles/some/folder/file.txt)
    "/foo.runfiles/some/folder": "/bar", # Now works and yields bar/file.txt
}
```

I have implemented some tests that demonstrate this behaviors.

Resolves #790

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -427,12 +427,11 @@ def add_from_default_info(
             # the executable, but that is not always obvious. When in doubt,
             # the first file of DefaultInfo.files should be the right target.
             base_file = the_executable or all_files[0]
-            base_file_path = mapping_context.path_mapper(
-                dest_path(base_file, data_path, data_path_without_prefix))
+            base_file_path = dest_path(base_file, data_path, data_path_without_prefix)
             base_path = base_file_path + ".runfiles"
 
             for rf in runfiles.files.to_list():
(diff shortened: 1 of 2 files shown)


cgrindel:
@aiuto This looks reasonable to me. I noticed that you started a dialogue with the author. Perhaps you should review it.

PR #816, opened 2024-02-05
Fix RPM package release and version files expansion

k0walik:
## What

While trying to migrate from version 0.9.1 to 0.10.0 I stumbled across following error:

```
Traceback (most recent call last):
        File "/home/tomasz.wojno/code/rules_pkg/pkg/rpm_pfg.bzl", line 352, column 47, in _pkg_rpm_impl
                content = substitute_package_variables(ctx, "\n".join(preamble_pieces)),
        File "/home/tomasz.wojno/code/rules_pkg/pkg/private/util.bzl", line 95, column 34, in substitute_package_variables
                return attribute_value.format(**vars)
Error in format: Missing argument 'VERSION_FROM_FILE'
```

In #787 a helper function for expanding RPM preamble with user-defined variables was introduced. The function uses Starlark format expression to replace `{var}` occurrences with defined variables. However, this does not work when we pass release or version file to the build target. The `rpm_pfg.bzl` appends `Release: ${RELEASE_FROM_FILE}` and `Version: ${VERSION_FROM_FILE}` to preamble, which are later expanded by `make_rpm.py` script. Thus, those variables are failed to be expanded as they are not defined.

## Changes

This pull request introduces following changes:
- Avoid expanding `RELEASE_FROM_FILE` and `VERSION_FROM_FILE` by wrapping them in double curly braces. This way `${{RELEASE_FROM_FILE}}` becomes expanded to `${RELEASE_FROM_FILE}`, so it can be processed later by `make_rpm.py`
- Add tests that uses `version_file` and `release_file` attributes
- Fix passing changelog file to OutputGroupInfo

### DIFF
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -274,7 +274,7 @@ def _pkg_rpm_impl(ctx):
         if ctx.attr.version:
             fail("Both version and version_file attributes were specified")
 
-        preamble_pieces.append("Version: ${VERSION_FROM_FILE}")
+        preamble_pieces.append("Version: ${{VERSION_FROM_FILE}}")
         args.append("--version=@" + ctx.file.version_file.path)
         files.append(ctx.file.version_file)
     elif ctx.attr.version:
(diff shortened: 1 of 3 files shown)


cgrindel:
@aiuto Do you want to review or can I merge this PR?

aiuto (approved):
Thanks for adding good tests along with this.

PR #825, opened 2024-02-29
fix: remove unused glob patterns

dieortin:
Some glob patterns do not match any file, which makes building with --incompatible_disallow_empty_glob produce an error. We should be able to just remove them.


aiuto (approved):
Thanks for the fix.

PR #830, opened 2024-03-13
fix: use appropriate entry type for runfiles that are directories

dieortin:
This fixes a regression at some point from 0.9.1 to 0.10.1, which caused `pkg_tar` to fail when `include_runfiles` was true and some of the sources had data dependencies that were directories.

The code that handles the `include_runfiles` attribute was always assigning ENTRY_IS_FILE as entry type of runfiles. This caused an error in the python code when packaging, as it encountered a directory in an unexpected place.

The fix is just considering ENTRY_IS_TREE types, in the same way as it is done in `_process_pkg_files`.

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -436,7 +436,7 @@ def add_from_default_info(
                 _check_dest(mapping_context.content_map, d_path, rf, src.label, mapping_context.allow_duplicates_with_different_content)
                 mapping_context.content_map[d_path] = _DestFile(
                     src = rf,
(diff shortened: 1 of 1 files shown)


aiuto (changes requested):
Can you add a test for this case.  This is something that should have been caught early. Without the test, someone is likely to break it again.

helly25:
As a side note: While this is better, it does not fully solve the problem. When the directory is linked, then is_directory will be false since links are files.

aiuto:
Another change seems to have fixed the problem, as after the merge from main there are not diffs.

Issue #832, opened 2024-03-15
Dangling archive paths are not supported by pkg_tar


eejayes:
Common tar utilities allow inclusion of file paths with some depth to them, where the resulting archive does not include records for the parent directories and their corresponding file permissions.  On the other hand, when `pkg_tar` is presented with a path where an explicit specification of a path's parent directories is not provided, records will automatically be created with a default set of file permissions.  While fine-grained control of directory permission levels is available through `pkg_mkdirs`, it is not possible to instruct `pkg_tar` to refrain from creating file permission records for parents of a dangling path.

The common use case for a tar archive with a fragmented file structure, is one where extraction is being performed on top of a pre-existing file structure where only the paths specified in the archive should be subject to change. A work around of `pkg_tar`s behavior with limited applicability could be to ensure that declared directory permissions for parent directories match the state of the pre-existing structure on the extraction target. Some more specialized use cases are:
1. Parent directory records that a user doesn't want to specify correspond to paths on the extraction target that are not mounted as writable file systems. In this case an error will occur during extraction.
2. Other expected factors on the extraction target can cause it's file system state to change that is not synchronized with the subject 'pkg_tar' specification.

aiuto:
This seems like a reasonable feature if anyone would like to implement it.

PR #834, opened 2024-03-15
Add support for no parent directory inference (#832)

eejayes:
Add feature as described in #832.

RELNOTES: Automatic creation of parent directory specifications for paths with depth are prevented in pkg_tar archives.

### DIFF
--- a/pkg/private/tar/tar_writer.py
+++ b/pkg/private/tar/tar_writer.py
@@ -46,6 +46,7 @@ class TarFileWriter(object):
                name,
                compression='',
                compressor='',
+               create_parents=False,
                default_mtime=None,
                preserve_tar_mtimes=True):
     """TarFileWriter wraps tarfile.open().
@@ -106,6 +107,7 @@ class TarFileWriter(object):
     # we can adjust that here based on the setting of root_dirctory.
     self.directories.add('/')
     self.directories.add('./')
+    self.create_parents = create_parents
 
   def __enter__(self):
     return self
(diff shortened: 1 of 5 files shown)


eejayes:
This is a start, intended to solicit early feedback. Existing tests have been merely adapted. New tests have not been added.

aiuto (changes requested):
I think `create_parents` should be an attribute of TarWriter. Then we set it once on creation instead of having to plumb the constant through on each call.

On the plus side, this makes the code cleaner.
A potential minus is that we can't have add_parents toggle per src to the tarball.
I don't think that is a needed, or even desired, ability. It would only marginally facilitate a sparse tar with a few added directories. You could simulate that easily by composing 
tars or (perhaps) by having an option on pkg_files to auto-insert needed directories.

eejayes:
I agree and would have done what you suggested to start with had it been taken into account that the change subject was a Class. And no, I don't see an obvious use case for having fine-grained control of the subject behavior within a target. 
Let me know if you think the tests suffice. Otherwise I don't intend to add more to them.

pkg/private/tar/tar_writer.py
```diff
@@ -107,6 +108,8 @@ def __init__(self,
     self.directories.add('/')
     self.directories.add('./')
 
+    self.create_parents = create_parents
```
aiuto, line 111: This does not need another blank line above it.

aiuto (commented):
I think this will do for tests.
Thanks for the fix.

pkg/private/tar/tar_writer.py
eejayes: Acknowledged.

eejayes:
> I think this will do for tests. Thanks for the fix.

My pleasure!

aiuto (approved):

PR #840, opened 2024-03-22
Update scorecard.yml to new checkout and analysis actions.

aiuto:
Update scorecard workflow as per https://github.com/ossf/scorecard-action

My earlier suggested fix had no effect.  This one seems more likely, based on the error message on the security tab
```
[Scorecard analysis](https://github.com/bazelbuild/rules_pkg/actions/runs/8391092847/job/22980629120)
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

That suggests that actions/checkout is out of date.

\<rant>I think it is inexcusable for a security scanner to not do numbered releases. You can not expect humans to audit and reason about github hashes.\</rant>

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -35,12 +35,12 @@ jobs:
         uses: actions/setup-node@v4.0.2
 
       - name: "Checkout code"
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
         with:
           persist-credentials: false
 
       - name: "Run analysis"
-        uses: ossf/scorecard-action@e38b1902ae4f44df626f11ba0734b14fb91f8f86 # v2.1.2
+        uses: ossf/scorecard-action@80e868c13c90f172d68d1f4501dee99e2479f7af # v2.1.3
         with:
           results_file: results.sarif
           results_format: sarif


meteorcloudy (approved):

aiuto:
Whump, whump, whump.   That still does not help.  
I've spent more time than I can afford on this for now.  I'm going to file a bug against scorecard and see what they say.

.github/workflows/scorecard.yml
```diff
@@ -35,12 +35,12 @@ jobs:
         uses: actions/setup-node@v4.0.2
 
       - name: "Checkout code"
-        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
```
cgrindel, line 38: We may want to update this to 4.x.x soon. GitHub is [deprecating actions that use node 16](https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/).

aiuto:
Probably yes. What I want is for scorecard to fix their documentation and
examples so they are up to date.

A potential policy based solution is that all the bazelbuild repositories
share the same scorecard setup. That way we only have to fix it for
bazelbuild/bazel and everyone else feels safe just copying that workflow to
their repository. This reduces churn and improves accountability.

PR #841, opened 2024-03-25
trying scorecard update one more time

aiuto:
Pin to version tags rather than hashes, as per https://github.com/ossf/scorecard/issues/3968

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -32,15 +32,15 @@ jobs:
 
     steps:
       - name: Setup Node.js environment
-        uses: actions/setup-node@v4.0.2
+        uses: actions/setup-node@v4
 
       - name: "Checkout code"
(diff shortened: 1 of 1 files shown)


.github/workflows/scorecard.yml
```diff
@@ -32,15 +32,15 @@ jobs:
 
     steps:
       - name: Setup Node.js environment
-        uses: actions/setup-node@v4.0.2
+        uses: actions/setup-node@v4
 
       - name: "Checkout code"
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        uses: actions/checkout@v3
```
cgrindel, line 38: You need to move to [v4](https://github.com/actions/checkout/releases/tag/v4.0.0) to get the new Node version.

cgrindel (approved):

Issue #849, opened 2024-04-11
Duplicate file paths not accepted by pkg_tar


eejayes:
When tar archives which contain duplicate file paths are assigned to deps of `pkg_tar`, the function removes all but the first occurring instance, and warns with "Duplicate file in archive picking first occurrence". This contrasts with the common GNU tar utility, where "tar allows you to have infinite number of files with the same name" according to document for `--append` under [The Five Advanced tar Operations](https://www.gnu.org/software/tar/manual/html_node/Operations.html#Operations), where unpacking of the archive will yield the last occurring instance of the file path.

My use case involves doing tar archive augmentation through a series of build steps, where one of these operations is to overwrite files of an existing archive. It is not an option to unpack the archives, one over the other, and then repack the result, because new entries for parent directories of the files in the incoming archives will be created which state their permissions. This is because the state of the directories on the extraction target must be preserved, rather than set to those which were present in the environment where the augmentation took place. So if an archive with duplicate files is presented to `pkg_tar` where the one most recently added is meant to overwrite, it is actually the original, or first file, which is retained, and thus when unpacking the overwriting behavior is not achieved.

I feel that pkg_tar's behavior in this case should be compatible with GNU tar, in order to support interoperability and leverage from common understanding that it has probably established.

aiuto:
Compatibliy with gnutar is not a goal we have been aiming for. We've done a lot of work to explicitly detect duplicate files because they are generally a sign that you didn't lay out the inputs correctly.  Can you provide a more concrete example of what you are trying to with laying down multiple tar balls?
Also, can you explain how the `allow_duplicates_with_different_content` attribute does not address your use case.

eejayes:
What I am trying to do is combine two tar files, where one of them is regarded as having overwrite priority for all path conflicts. Before going into an example, my understanding of `allow_duplicates_with_different_content` is that the duplication corresponds to that occurring across dependencies, where the use case I consider is duplication within a pre-existing (dependency) archive. Also given that it's `true` default means that this option is being exercised when the problem occurs, it does not seem to be a solution.  Also given the document, it seems like this is a deprecated interface. However I have not analyzed it's behavior, and could be missing something.

I will try to illustrate the problem here. For some diversity in tar utilities, I also exemplify `bsdtar`s behavior.

# Example "conflicting" input archives
```
$ tar tvf archive_a.tar
drwxr-xr-x s0000184/linuxusers 0 2024-04-12 08:30 ./
[...]

PR #850, opened 2024-04-11
Support duplicate file paths in tar archives

eejayes:
Duplicate path entries are made possible within tar archives, as per feature request described in #849.

RELNOTES: Duplicate path entry support within tar archives


eejayes:
I will await design feedback before updating tests.

aiuto:
I think the discussion in #849 came up with a good proposal. Whenever you're ready.

eejayes:
Hello. Yes, I would agree. I have started to work on it.

eejayes:
This provides `pkg_tar` with support for duplicate file paths. Tests added show that:
- an externally defined archive that is propagated through `pkg_tar` simply has it's structure preserved
- adding a path via `srcs` which conflicts with a path within an archive in `deps` properly results in duplicates
- adding two archives via `deps` which have conflicting paths properly results in duplicates

In the pre-existing code, `deps` are added after `srcs`. Thus for paths appearing both in `deps` and `srcs`, it is the one in `deps` which will appear later in the archive list and will thus prevail when an archive is unpacked. This conflicts with common command line _tar_ tools where:
1. users may specify which archive is subject to modification during an archive concatenation operation
2. the archive is implicitly the modification subject in an operation combining a file and archive

It would be straight-forward to adapt the code to the second case. The first case would however require a mechanism to specify which archive (i.e. `dep`) is the modification subject, i.e. the sequence in which archives should be extended. The proposed iteration does not solve these. Instead it requires that multiple sequential `pkg_tar` operations be declared when there are path conflicts between any pair of `deps` or `srcs`.

The purpose of duplicate support is to support evolution of artifacts as consistent with typical _tar_ utility behavior. Since files can only serve to specify change, they are never artifacts and thus `srcs` conflicts which could exist within the scope of a single `pkg_tar` artifact should be solved prior to it's declaration. Adding duplicates to `srcs` should be an error. 

Outside of the discussion in the issue #849, there were some unforeseen interactions with `create_parents`. Currently because duplicates are not allowed, if `create_parents` is specified for a pre-existing directory, it will not be created. However with the support of `allow_duplicates_from_deps`, an additional directory could be created with the arbitrary default permissions. Since directory permissions of a pre-existing path represent information that is either equally good to the default permissions which are applied in the absence of information, or better in the case that a user has explicitly declared them, it makes more sense to use the pre-existing path info even with `allow_duplicates_from_deps`.

Future improvement considerations are:
1. Reversal of file and tar addition ordering to address (2) above
2. Sequencing mechanism for tar construction (1) above
3. Adding duplicates file paths to `srcs` should be an error.

aiuto:
I can look at this later today.

aiuto (approved):
Sorry for the slowness. Things got busy here.

PR #851, opened 2024-04-12
rpm: Add conflicts and obsoletes to sub RPMs

kellyma2:
This change brings the sub RPM rules closer to parity with both reality and the parent RPM rules by adding the `Conflicts` and `Obsoletes` fields to them.


aiuto (approved):

aiuto:
The CI failure seem seems unrelated. Merging anyway.

Issue #808, opened 2024-01-10
Update stardoc in MODULE.bazel so we can build with bazel 7.x in default mode


aiuto:
Right now you need to build with `--noenable_bzlmod`.
But, IIRC, the name of the repo is different under bzlmod, so you have to fix the build files.
Such is life.

aiuto:
Update: April 2024.  Bazel at is pretty much breaking --noenable_bzlmod.  You tend to get an error like
```
(02:55:28) ERROR: Error computing the main repository mapping: Label '@@platforms//host:extension.bzl' is invalid because 'host' is not a package; perhaps you meant to put the colon here: '@@platforms//:host/extension.bzl'?
(02:55:28) Computing main repo mapping:
```

PR #852, opened 2024-04-16
Force stardoc integration test to run with bazel 7.0.0

aiuto:
Force the integration test (stardoc and packaging) to run with bazel 7.0.0.
Bazel at head seems to have broken --noenable_bzlmod in some cases. See #808.

This also bumps platforms to the latest, but that was a red herring from the error message in the underlying problem


aiuto:
Whoopie.  CI is green again.

meteorcloudy (approved):

PR #854, opened 2024-04-17
pkg_tar: if symlink starts with ./, keep it absolute

aiuto:
For the `symlink` attribute of `pkg_tar`, if the path begins with ./, preserve the path rather than prefixing package_dir. This allows graceful migration for some uses who rely on the previous package_dir behavior.

Hyrum's law is a strong one.


aiuto:
cc: @lberki  This was one of the things which was blocking import.

cgrindel (approved):

Issue #857, opened 2024-04-17
pkg_rpm does not support the rpm 'Epoch' field


TvdW:
The Epoch rpm field allows replacing a package with a similarly named package and 'reset' the versioning by inserting a `1:` at the start. This allows a `0.0.1` version to be considered than `2.0.0`, if the epoch is higher (eg. `1:0.0.1 > 2.0.0`).

PR #858, opened 2024-04-17
Add support for "Epoch" attributes in RPMs

TvdW:
Fixes https://github.com/bazelbuild/rules_pkg/issues/857

### DIFF
--- a/tests/rpm/BUILD
+++ b/tests/rpm/BUILD
@@ -280,6 +280,7 @@ pkg_rpm(
     architecture = "noarch",
     conflicts = ["not-a-test"],
     description = """pkg_rpm test rpm description""",
+    epoch = "1",
     license = "Apache 2.0",
     post_scriptlet_file = ":post",
     postun_scriptlet_file = ":postun",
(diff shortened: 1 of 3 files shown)


cgrindel (approved):
I will wait to merge so that @aiuto can take a look.

tests/rpm/BUILD
```diff
@@ -179,6 +179,7 @@ pkg_rpm(
     architecture = "noarch",
     conflicts = ["not-a-test"],
     description = """pkg_rpm test rpm description""",
+    epoch = "1",
```
aiuto, line 182: Since this is optional, it should not be in every test case.

tests/rpm/pkg_rpm_basic_test.py
```diff
@@ -74,6 +74,7 @@ def test_scriptlet_content(self):
 
     def test_basic_headers(self):
         common_fields = {
+            "EPOCH": b"1",
```
aiuto, line 77: Can you pass in the check for EPOCH to only the samples which add it?
I know that is not the same pattern here, but what is here was hacked up in a hurry. Having version and release the same for everything is not so bad, but we should have split out noarch and group so they can vary by test case.

aiuto (changes requested):

tests/rpm/BUILD
tvdw-mongodb: Agree. Updated 😄

TvdW: Agree! Updated :smile:

tests/rpm/pkg_rpm_basic_test.py
TvdW: Done

aiuto (approved):
Thanks for the change

PR #859, opened 2024-04-17
Enable pkg_rpm and pkg_subrpm to create empty RPMs

### LINKED ISSUE #711: pkg_rpm fails trying to create an empty rpm package
Given the following:

```starlark
pkg_files(
    name = "empty",
    srcs = [],
)

pkg_rpm(
    name = "empty_rpm",
    srcs = [
        ":empty",
    ],
    architecture = "noarch",
    description = """description""",
    license = "Apache 2.0",
    summary = "summary",
    version = "1.0",
)
```
`empty_rpm` will eventually fail to build with:

```
error: Empty %files file /tmp/tmpa07nyy51/BUILD/empty_rpm.spec.files
```

Creating an empty package should be possible.  For example, one could wish to create a package consisting entirely of metadata that requires other packages.  There are two ways users could hit this issue:

- Users could (inadvertently) create a `pkg_files` with an empty input and pass it to `pkg_rpm`.  The error presented to the user is related to `pkg_rpm` internals and is not useful to end-users.
- Users could deliberately pass in an empty `srcs` value to `pkg_rpm`.  In this case, we get a failure in the `pkg_rpm` wrapper macro due to relaxed boolean checks on incoming values (empty lists are "falsey").  Checking specifically for `None` should be enough.  When tested locally, this ends up producing the same error as above.

IMO, making `srcs` non-`mandatory` is not a good idea, as requiring its presence (even if it is an empty value) makes the intent clearer.

Other packages (debian packages?) may also have this need as well.

#209 may help to simplify the fix here, but it is not necessary to address this problem.

kellyma2:
At present we fail in two ways if we try to create empty RPMs:

- we expect srcs to be both non-empty and will fail if it is empty and we have no spec file

- we don't emit anything for the `%files` block in the RPM if there are no actual files and rpmbuild doesn't like this

This change tweaks the former condition so that srcs has to be non-None or we have to have a specfile, but will allow us to have an empty (`[]`) value for srcs.  Additionally, it injects `%defattr(-,root,root)` as a reasonable default for the `%files` blocks so as to allow rpmbuild to be happy with what we're providing.

Fixes #711

### DIFF
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -352,10 +352,14 @@ def _process_subrpm(ctx, rpm_name, rpm_info, rpm_ctx):
             "%%files %s" % rpm_info.package_name,
         ]
 
-        for dep in rpm_info.srcs:
-            _process_dep(dep, sub_rpm_ctx)
+        if rpm_info.srcs:
+            for dep in rpm_info.srcs:
+                _process_dep(dep, sub_rpm_ctx)
 
-        rpm_lines += sub_rpm_ctx.rpm_files_list
+            rpm_lines += sub_rpm_ctx.rpm_files_list
+        else:
+            rpm_lines += "%defattr(-,root,root)"
+
         rpm_lines += [""]
 
     rpm_ctx.install_script_pieces.extend(sub_rpm_ctx.install_script_pieces)
@@ -702,10 +706,13 @@ def _pkg_rpm_impl(ctx):
     rpm_files_file = ctx.actions.declare_file(
         "{}.spec.files".format(rpm_name),
     )
-    ctx.actions.write(
-        rpm_files_file,
-        "\n".join(rpm_ctx.rpm_files_list),
-    )
+    if rpm_ctx.rpm_files_list:
+        ctx.actions.write(
+            rpm_files_file,
+            "\n".join(rpm_ctx.rpm_files_list),
+        )
+    else:
+        ctx.actions.write(rpm_files_file, "%defattr(-,root,root)")
 
     # TreeArtifact processing work
     if rpm_ctx.packaged_directories:
--- a/pkg/rpm_pfg.bzl
+++ b/pkg/rpm_pfg.bzl
@@ -360,7 +360,7 @@ def _process_subrpm(ctx, rpm_name, rpm_info, rpm_ctx):
 
         # rpmbuild will be unhappy if we have no files so we stick
         # default file mode in for that scenario
-        rpm_lines += DEFAULT_FILE_MODE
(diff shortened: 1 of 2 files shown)


pkg/rpm_pfg.bzl
```diff
@@ -352,10 +352,14 @@ def _process_subrpm(ctx, rpm_name, rpm_info, rpm_ctx):
...
 
-        for dep in rpm_info.srcs:
-            _process_dep(dep, sub_rpm_ctx)
+        if rpm_info.srcs:
+            for dep in rpm_info.srcs:
+                _process_dep(dep, sub_rpm_ctx)
+
+            rpm_lines += sub_rpm_ctx.rpm_files_list
+        else:
+            rpm_lines += "%defattr(-,root,root)"
```
aiuto, line 361: Should you just write this as the first line in all cases? Would that be wrong?

pkg/rpm_pfg.bzl
```diff
@@ -702,10 +706,13 @@ def _pkg_rpm_impl(ctx):
...
-        rpm_files_file,
-        "\n".join(rpm_ctx.rpm_files_list),
-    )
+    if rpm_ctx.rpm_files_list:
+        ctx.actions.write(
+            rpm_files_file,
+            "\n".join(rpm_ctx.rpm_files_list),
+        )
+    else:
+        ctx.actions.write(rpm_files_file, "%defattr(-,root,root)")
```
aiuto, line 715: Same question here.

aiuto (commented):
I still wonder why anyone would want an empty RPM, but I'll go with the flow on that.

pkg/rpm_pfg.bzl
kellyma2: That is one possibility.  Do we want to parameterize this though so that we can specify some useful defaults or just leave it as a contant?

kellyma2:
> I still wonder why anyone would want an empty RPM, but I'll go with the flow on that.

The examples I've encountered thusfar involve having an empty / stub top level RPM with non-empty sub RPMs, although I figured it's harmless to also enable sub RPMs to do this too.

pkg/rpm_pfg.bzl
aiuto: Oh, now you are asking hard questions. I think a constant is fine because IIRC we write the mode for every file.
But, now I am less certain.
How about just adding a comment about why we emit the defattr and that perhaps we could add it. And some future person will make that choice when they want to cleanup that code.

kellyma2: Done

aiuto (approved):

Issue #803, opened 2023-12-31
py_binary and oci_image don't play nice when mediated via pkg_tar


bcsgh:
I don't know that this is a bug, of _who's_ bug it is, but this seems like a good place to start trying to make thins better. (Feel free to point me at someone else.)

tl;dr; when `pkg_tar` is wrapped in the most obvious way around a `py_binary` so that it can be stuffed in an `oci_image`, files from the local workspace end up in a different location than the wrapper script that `py_binary` generates expect them to be in. (Non local files might as well, but things started working for me before I ever observed that.) I make no claim to it being correct, but the following commit seems to resolve the problem by inserting the `workspace_name` into the a suitable spot in the paths that will used in the tar. 

https://github.com/bcsgh/rules_pkg/commit/8adcd2a83dc5a3fd642573cc46af16a774631732

I expect that would be a breaking change for more than a few users. But then again, the existing behavior is wacky enough that it's possible that nobody has anything that works depending on it. ([Or maybe not.](https://www.hyrumslaw.com/))

bcsgh:
Note: seems the relevant code is somewhat new: https://github.com/bazelbuild/rules_pkg/pull/754

ConnorJLatham:
Something interesting here is I put a git_override in my MODULE.bazel
```bazel
bazel_dep(name = "rules_pkg")
git_override(
    module_name = "rules_pkg",
    remote = "https://github.com/bazelbuild/rules_pkg",
    commit = "994a1f5b94f5baa4904d9487622c2908fe44d2d1",
)
````

so that it would pull in the changes from that pr above, but im still having issues with certain files not being found. Here is how I build the image up.

```bazel
py_binary(
    name = "sadie_bot",
    srcs = ["sadie_bot.py"],
    tags = [
        "manual",
        "no-sandbox",
    ],
    deps = [
        "@external_py//discord_py",
    ],
)

pkg_tar(
    name = "sadie_bot_pkg",
    srcs = [
        ":sadie_bot",
    ],
    include_runfiles = True,
)

oci_image(
    name = "sadie_bot_image",
    base = "@python_base",
    entrypoint = ["./sadie_bot"],
    tars = [":sadie_bot_pkg"],

)
```

And here is the output when I try to run up the container. It seems to enter the script just fine, but we don't seem to be replicating the runfiles exactly. 

```bash
(.venv) [clatham@clatham-dev sadie_bot]$ docker compose up
[+] Running 1/0
[...]

PR #861, opened 2024-04-22
Fix binary location in pkg_tar

ewianda:
Fix binary location in pkg_tar

close  https://github.com/bazelbuild/rules_pkg/issues/803


cgrindel (commented):
It looks like #860 needs to merge first. Is that correct?

Do we have an example/test that exercises this fix?

ewianda:
> It looks like #860 needs to merge first. Is that correct?
> 
> Do we have an example/test that exercises this fix?

Sorry forgot to squash previous commits

PR #863, opened 2024-04-25
Exclude release from filename when using release_file

kellyma2:
When we're using `release_file` in lieu of `release` we're just pointing rpmbuild at the file containing the `Release` string and we don't have it available to inject into the filename resulting in a strange looking filename of the form `Foo-version-.arch.rpm`.

This change extracts the RPM name generation to a single helper, `_make_rpm_filename` and tweaks it s.t. if we're missing the value for `release` we'll just exclude it from the filename format instead.


cgrindel (approved):
LGTM. I will wait for @aiuto to review before merging.

aiuto (approved):

PR #856, opened 2024-04-17
Append the workspace name to the runfiles directory name

EdSchouten:
The runfiles directory is broken down by workspace name. This means that files belonging to the main workspace are placed at "${name}.runfiles/${ctx.workspace_name}/${short_path}". Files belonging to externals have a short_path starting with "../${external_name}".

Because we currently don't append ctx.workspace_name to the computed runfiles path, files belonging to the main workspace are placed at the top level of the runfiles directory, which is incorrect. Even worse is that files belonging to external repositories end up alongside the runfiles directory, instead of being contained within.

This makes the runfiles directory layout consistent with how Bazel creates it locally:

```
$ find bazel-bin/tests/an_executable.runfiles \! -type d 
bazel-bin/tests/an_executable.runfiles/MANIFEST
bazel-bin/tests/an_executable.runfiles/_repo_mapping
bazel-bin/tests/an_executable.runfiles/_main/tests/testdata/hello.txt
bazel-bin/tests/an_executable.runfiles/_main/tests/foo.cc
bazel-bin/tests/an_executable.runfiles/_main/tests/an_executable
```

PR #864, opened 2024-04-25
Append workspace name to the runfiles directory name for `pkg_files`

sitaktif:
This is a follow-up to #856, which applies to most rules but `pkg_files` uses a different code path. This change applies the fix to `pkg_files` and adds a test to avoid future regressions.

It also fixes a (silent) bug in the test logic.

### DIFF
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -298,7 +298,7 @@ def _pkg_files_impl(ctx):
             target = file_to_target[src]
             runfiles = target[DefaultInfo].default_runfiles
             if runfiles:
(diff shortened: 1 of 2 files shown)


sitaktif:
@aiuto this should probably make it before the next release to avoid confusion.

cgrindel (approved):

sitaktif:
Thanks @cgrindel - I don't have write access so I'll defer merging to you.

cgrindel:
@aiuto Do you want to look at this before we merge?

sitaktif:
Thanks!

PR #865, opened 2024-04-30
Provide support for duplicate paths

eejayes:
Outside of the discussion in the issue #849, there were some unforeseen interactions with create_parents. Otherwise the behavioral change should be what was discussed.

It is outside of the scope to order where the paths from deps or srcs are added to the archives, within a given instance of pkg_tar. It may be worth documenting that controlling the ordering of paths whether they originate from tars which are concatenated or files, requires a sequence of `pkg_tar` instances rather than relying on a single one. In a way undefined behavior may be expected when adding paths which conflict in the same `pkg_tar` instance.


aiuto (commented):
I can not understand what this change is about from the PR description.
Can you fix that to clearly describe the behavior change and how it reduces undefined behavior? 
It does eliminate undefined behavior, right?

eejayes:
> I can not understand what this change is about from the PR description.
I understand that. I am not so acquainted with the workflow yet. Let's discontinue this PR in favor of #850 which should now be in good shape.

eejayes:
> I can not understand what this change is about from the PR description.

I understand that. I am not so acquainted with the workflow yet. Let's discontinue this PR in favor of https://github.com/bazelbuild/rules_pkg/pull/850 which should now be in good shape.

eejayes:
> I can not understand what this change is about from the PR description.

I understand that. I am not so acquainted with the workflow yet. Let's discontinue this PR in favor of https://github.com/bazelbuild/rules_pkg/pull/850 which should now be in good shape.

PR #869, opened 2024-06-03
1.0 release

aiuto:
Just bump the numbers & docs.


cgrindel (approved):

kellyma2:
@aiuto Are you planning to update latest in BCR?

tonyaiuto:
I'm sort of stuck on a merge there. I have not figured out a reasonable way
to have bzlmod ignore the compatibility level loop between ***@***.***
depending (transitively) on rules_proto, where all versions depend on
rules_pkg at level 1.

Updating rules_proto is obviously wrong in 2 ways. It can't work until
rules_pkg is already added. It puts the change burden in the wrong repo.
Doing a special MODULE.bazel that hard specified rules_proto might work.
But that feels wrong too. I don't care what version of proto you get.  User
workspace should win, followed by tools I depend directly on. It is also
tedious, because the bcr tests make sure the module.bazel inside the
tarball and given are identical. Which means a dummy release of rules_pkg
for each test.  That is ugly.

I filed bugs about the difficulty of updating a compatibilty level for any
widely used module. I just ran out of time to beat my head against
something that is probably better fixed with a bazel flag to not care about
version conflicts and a fix to the bcr submit process.

cgrindel:
@tonyaiuto If the compatibility level is causing so much pain. Perhaps, we should just put it back to the previous value and document the breaking changes. Personally, I think having a BCR entry for a release is more important than signaling breaking changes with the compatibility level. WDYT?

aiuto:
That may be the path of least resistance.  I don't have the time to figure
out a general solution to the transitive loop problem.

I'll send a PR

PR #874, opened 2024-06-28
Add AlmaLinux 9.3 debuginfo support

kellyma2:
AlmaLinux 9.3 appears to behave as a CentOS variant so adding it in the same places.  This change also cleans up the way we declare debuginfo_type to make it a bit less verbose.


cgrindel (approved):
LGTM. I will wait to merge so that @tonyaiuto can have a look.

aiuto (approved):

PR #878, opened 2024-07-25
Add support for repo mapping

### LINKED ISSUE #769: Look into repo-mapping support
Context: https://github.com/bazelbuild/bazel/issues/19937
https://github.com/bazelbuild/bazel/pull/19944

Under bzlmod, paths to files in external repos changed. It may be difficult for users to write rules to rebase files from external repos to new locations. No time to write more details today, but I want this placeholder issue to point to the problem and the fix.

lamcw:
## Context

When a binary is packaged with `pkg_*` rules, the repo_mapping manifest file is not included in the archive. For example

```starlark
java_binary(
    name = "example",
    main_class = "...",
)

pkg_tar(
    name = "example_tar",
    srcs = [":example"],
    include_runfiles = True,
)
```

The above generates an archive that does not have `_repo_mapping` under runfiles directory even when bzlmod is enabled, and therefore runfiles libraries are not able to convert apparent repo names into canonical repo names.

## Changes

Patch `rules_pkg` such that when

1. bzlmod is enabled, and
1. `include_runfiles = True`

it copies the repo mapping manifest into the archive at the correct location

i.e.

```
example
example.repo_mapping
example.runfiles/
├─ _repo_mapping
```

Closes #769

### DIFF
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -29,6 +29,7 @@ here.
 
 load("@bazel_skylib//lib:paths.bzl", "paths")
 load("//pkg:providers.bzl", "PackageDirsInfo", "PackageFilegroupInfo", "PackageFilesInfo", "PackageSymlinkInfo")
+load("//pkg/private:util.bzl", "get_repo_mapping_manifest")
 
 # TODO(#333): strip_prefix module functions should produce unique outputs.  In
 # particular, this one and `_sp_from_pkg` can overlap.
@@ -311,12 +312,10 @@ def _pkg_files_impl(ctx):
                     else:
                         src_dest_paths_map[rf] = dest_path
 
-                if (
-                    getattr(target[DefaultInfo], "files_to_run") and
-                    hasattr(target[DefaultInfo].files_to_run, "repo_mapping_manifest") and
-                    target[DefaultInfo].files_to_run.repo_mapping_manifest != None
-                ):
-                    repo_mapping_manifest = target[DefaultInfo].files_to_run.repo_mapping_manifest
+                # if repo_mapping manifest exists (for e.g. with --enable_bzlmod),
+                # create _repo_mapping under runfiles directory
+                repo_mapping_manifest = get_repo_mapping_manifest(target)
+                if repo_mapping_manifest:
                     dest_path = paths.join(src_dest_paths_map[src] + ".runfiles", "_repo_mapping")
                     src_dest_paths_map[repo_mapping_manifest] = dest_path
 
--- a/pkg/mappings.bzl
+++ b/pkg/mappings.bzl
@@ -311,6 +311,15 @@ def _pkg_files_impl(ctx):
                     else:
                         src_dest_paths_map[rf] = dest_path
 
+                if (
+                    getattr(target[DefaultInfo], "files_to_run") and
+                    hasattr(target[DefaultInfo].files_to_run, "repo_mapping_manifest") and
+                    target[DefaultInfo].files_to_run.repo_mapping_manifest != None
+                ):
+                    repo_mapping_manifest = target[DefaultInfo].files_to_run.repo_mapping_manifest
+                    dest_path = paths.join(src_dest_paths_map[src] + ".runfiles", "_repo_mapping")
+                    src_dest_paths_map[repo_mapping_manifest] = dest_path
+
     # At this point, we have a fully valid src -> dest mapping in src_dest_paths_map.
     #
     # Construct the inverse of this mapping to pass to the output providers, and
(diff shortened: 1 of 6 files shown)


pkg/private/pkg_files.bzl
```diff
@@ -449,6 +449,41 @@ def add_from_default_info(
...
+                ))
+                add_single_file(
+                    mapping_context,
+                    dest_path = d_path,
+                    src = repo_mapping_manifest,
+                    origin = src.label,
+                    mode = mapping_context.default_mode,
+                    user = mapping_context.default_user,
+                    group = mapping_context.default_group,
+                )
```
lamcw, line 474: This should really be a symlink into `.runfiles/_repo_mapping`, but I have no idea how to add a symlink that also respect `remap_paths` here.

pkg/mappings.bzl
```diff
@@ -311,6 +311,15 @@ def _pkg_files_impl(ctx):
                     else:
                         src_dest_paths_map[rf] = dest_path
 
+                if (
+                    getattr(target[DefaultInfo], "files_to_run") and
+                    hasattr(target[DefaultInfo].files_to_run, "repo_mapping_manifest") and
+                    target[DefaultInfo].files_to_run.repo_mapping_manifest != None
+                ):
+                    repo_mapping_manifest = target[DefaultInfo].files_to_run.repo_mapping_manifest
+                    dest_path = paths.join(src_dest_paths_map[src] + ".runfiles", "_repo_mapping")
```
lamcw, line 320: This only adds `example.runfiles/_repo_mapping` -- `example.repo_mapping` is still missing.

cgrindel (approved):
LGTM. I will wait to merge it so that @tonyaiuto can take a look.

lamcw:
@aiuto can you PTAL?

pkg/private/pkg_files.bzl
aiuto: Can you add that comment to the code as a TODO, so some future maintainer sees it.

pkg/mappings.bzl
```diff
@@ -311,6 +311,15 @@ def _pkg_files_impl(ctx):
                     else:
                         src_dest_paths_map[rf] = dest_path
 
+                if (
```
aiuto, line 314: This is a bit of magic, and it is repeated below.
Can you extract it out to a routine (maybe in pkg/private/util.bzl) with a good name (maybe "repo_mapping_manifest_exists"?)

Also, there is a comment about bzlmod enabled. That should go with the routine to clarify why this check is the way it is.

aiuto (changes requested):

pkg/private/pkg_files.bzl
lamcw: Done in abdf4ae

pkg/mappings.bzl
lamcw: I've split the logic into `get_repo_mapping_manifest`, please take a look. 🙏 

RE: the comment about bzlmod they are not really related to the check, but the fact that we'd need to include repo_mappings in the runfiles directory. So I am keeping them with where we add repo mappings into the mapping_context

pkg/private/util.bzl
```diff
@@ -122,5 +122,7 @@ def get_repo_mapping_manifest(src):
     """
     files_to_run_provider = get_files_to_run_provider(src)
     if files_to_run_provider:
+        # repo_mapping_manifest may not exist in older Bazel versions (<7.0.0)
```
aiuto, line 125: Now "that" is a comment which makes it clear why you are doing something specal. Thanks.

aiuto (approved):

pkg/private/util.bzl
```diff
@@ -92,4 +92,37 @@ def substitute_package_variables(ctx, attribute_value):
...
+        src: target to get repo_mapping_manifest from
+
+    Returns:
+        File or None: repo_mapping_manifest
+    """
+    files_to_run_provider = get_files_to_run_provider(src)
+    if files_to_run_provider:
+        # repo_mapping_manifest may not exist in older Bazel versions (<7.0.0)
+        # https://github.com/bazelbuild/bazel/issues/19937
+        return getattr(files_to_run_provider, "repo_mapping_manifest")
```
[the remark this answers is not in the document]
tonyaiuto: Umm, because we don't control the definition of DefaultInfo.  In bazel 6, you fail if you reference repo_mapping_manifest.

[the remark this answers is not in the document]
tonyaiuto: Do you have a reproducible failure that this corrects?

PR #879, opened 2024-08-01
rpm: Refine prefix matching for sub RPMs to reduce fragility

kellyma2:
The current prefix matching for sub RPMs can potentially fail in two ways:

- if we have two overlapping prefixes we may match the shorter prefix against the wrong RPM
- the shorter prefix will match repeatedly even after finding its target

This change keeps track of which sub RPMs have already match so avoid
double matching a shorter prefix.  Additionally, we sort the sub RPM
list by name length in reverse order.  This ensures that we prefer to
match the longest prefixes first and avoid double matching with the
same prefix.

fixes: https://github.com/bazelbuild/rules_pkg/issues/875

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -446,13 +446,20 @@ class RpmBuilder(object):
   def SaveResult(self, out_file, subrpm_out_files):
     """Save the result RPM out of the temporary working directory."""
     if self.rpm_paths:
+      subrpms_seen = set()
+      sorted_subrpm_out_files = sorted(
+         subrpm_out_files, key=lambda n: len(n[1]), reverse=True)
+
       for p in self.rpm_paths:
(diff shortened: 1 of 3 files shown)


kellyma2:
This fixes https://github.com/bazelbuild/rules_pkg/issues/875

cgrindel:
Can we add a test that verifies the behavior being fixed is fixed?

aiuto (commented):
Looks good, but it could use a test.

kellyma2:
> Can we add a test that verifies the behavior being fixed is fixed?

Yes.  I'll try to get back to that this week.

kellyma2:
> Looks good, but it could use a test.

Done

cgrindel (approved):

aiuto (approved):
Thanks

Issue #882, opened 2024-08-16
rules_python version used in rules_pkg conflict with version 0.34.0 of rules_python, if rules_pkg is loaded first and then rules_python


sjonnalagadda:
The error is:

`ERROR: BUILD.bazel:4:25: //:requirements_test: no such attribute 'src' in 'py_test' rule (did you mean 'srcs'?)`

Here is the setup

WORKSPACE file contents

`# https://github.com/uber/hermetic_cc_toolchain/blob/main/examples/rules_cc/WORKSPACE

load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "rules_pkg",
    urls = [
        "https://mirror.bazel.build/github.com/bazelbuild/rules_pkg/releases/download/1.0.1/rules_pkg-1.0.1.tar.gz",
        "https://github.com/bazelbuild/rules_pkg/releases/download/1.0.1/rules_pkg-1.0.1.tar.gz",
    ],
    sha256 = "d20c951960ed77cb7b341c2a59488534e494d5ad1d30c4818c736d57772a9fef",
)
load("@rules_pkg//:deps.bzl", "rules_pkg_dependencies")
rules_pkg_dependencies()

http_archive(
    name = "rules_python",
    sha256 = "778aaeab3e6cfd56d681c89f5c10d7ad6bf8d2f1a72de9de55b23081b2d31618",
    strip_prefix = "rules_python-0.34.0",
    url = "https://github.com/bazelbuild/rules_python/releases/download/0.34.0/rules_python-0.34.0.tar.gz",
)

# "Installation" for rules_python
load("@rules_python//python:repositories.bzl", "py_repositories", "python_register_toolchains")

py_repositories()

python_register_toolchains(
    name = "python_toolchain",
    python_version = "3.11.3",
)

load("@python_toolchain//:defs.bzl", "interpreter")

load("@rules_python//python:pip.bzl", "pip_parse")
`

BUILD file contents

`load("@rules_python//python:pip.bzl", "compile_pip_requirements")

compile_pip_requirements(
    name = "requirements",
    src = "requirements.in",
    requirements_txt = "requirements_lock.txt",
)
`
requirements.in file contents

`load("@rules_python//python:pip.bzl", "compile_pip_requirements")

compile_pip_requirements(
    name = "requirements",
    src = "requirments.in",
    requirements_txt = "requirements_lock.txt",
)
`

Run command

` bazelisk run //:requirements.update`

[...]

PR #884, opened 2024-08-17
issue: 882

sjonnalagadda:
Upgrading rules_python version and fixed a regression test issue

Fix for issue: https://github.com/bazelbuild/rules_pkg/issues/882


MODULE.bazel
```diff
@@ -7,7 +7,7 @@ module(
 
 # Do not update to newer versions until you need a specific new feature.
 bazel_dep(name = "rules_license", version = "0.0.7")
-bazel_dep(name = "rules_python", version = "0.31.0")
+bazel_dep(name = "rules_python", version = "0.34.0")
```
cgrindel, line 10: Can you explain why we need this version bump?

sjonnalagadda: I tried to keep it uniform. The consumers of rules_pkg will get rules_python 0.34.0. So, if rules_pkg also need a version of rules_python for it's build, then let the build also use same version.

cgrindel: My question is whether there is a feature in `rules_python` 0.34.0 that `rules_pkg` is now using? In general for bzlmod, rulesets should declare the minimum viable version for their dependencies. This allows clients to dictate the version of a dependency.

aiuto: +1 I am very dubious about this.  It pulls in rules_python internal things. 
That may work at this instant in time, but it feels brittle w.rt. new versions.

Issue #308, opened 2021-03-02
Create "make install" equivalent


nacl:
Open source projects are typically built in three phases:

1) Configuration (`./configure`), where the project build is customized to target specific featuresets
2) Building (`make`), where the project is actually built
3) Installation (`make install`), where the project is installed to the filesystem, or staged for future packaging using the [`DESTDIR` convention](https://www.gnu.org/prep/standards/html_node/DESTDIR.html).

Bazel combines 1) and 2) (configuring is done with build flags), and 3) doesn't really exist.  Currently, the closest bazel equivalent to have bazel create a package representing the installable outputs, most often a .zip or a .tar.  While effective, it is wasteful when packages become quite large.

Since Bazel builds have no way to interact directly with the filesystem outside of the buildroot, we need to be a little more creative.  Let's create a `pkg_install` rule that creates a (python?) script that knows how to install artifacts outside of the bazel buildroot, for example:

```starlark
load("@rules_pkg//:install.bzl", "pkg_install")

# pkg_files, pkg_filegroups, etc

pkg_install(
    name = "install",
    srcs = [
        # pkg_filegroups here
   ],
)
```

Acceptance criteria:

- `pkg_install` (or its equivalent) is created with tests
- `pkg_install` supports staging to an arbitrary destination (`DESTDIR`)

aiuto:
Is this done and should be split into distinct issues for improvements? Or do you consider pkg_install not yet ready.

jacky8hyf:
I can see that this is not yet ready. In particular, TreeArtifacts and symlinks are not supported yet.

https://github.com/bazelbuild/rules_pkg/blob/447fb8e246cb437630b9ecd09f9bf9fe857d2c3f/pkg/private/install.py.tpl#L104

In addition, docs have not been updated in https://bazelbuild.github.io/rules_pkg/ to include install.bzl.

I am going to contribute to pkg_install to support TreeArtifacts in a pull request to solve the imminent problem for our use cases.

PR #885, opened 2024-08-19
pkg_install: Support TreeArtifacts.

jacky8hyf:
Implementation notes:
For tree artifacts, when creating directories, we mostly follow the modes set for the whole TreeArtifact, but also +x to allow searching the directory. This is similar to how pkg_tar etc. handles things.

Link: https://github.com/bazelbuild/rules_pkg/issues/308

### DIFF
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -74,11 +74,11 @@ class NativeInstaller(object):
                 logging.info("CHOWN %s:%s %s", user, group, dest)
                 shutil.chown(dest, user, group)
 
-    def _do_file_copy(self, src, dest, mode, user, group):
+    def _do_file_copy(self, src, dest):
         logging.info("COPY %s <- %s", dest, src)
         shutil.copyfile(src, dest)
 
-    def _do_mkdir(self, dirname, mode, user, group):
+    def _do_mkdir(self, dirname, mode):
         logging.info("MKDIR %s %s", mode, dirname)
         os.makedirs(dirname, int(mode, 8), exist_ok=True)
 
@@ -93,12 +93,12 @@ class NativeInstaller(object):
 
     def _install_file(self, entry):
         self._maybe_make_unowned_dir(os.path.dirname(entry.dest))
-        self._do_file_copy(entry.src, entry.dest, entry.mode, entry.user, entry.group)
+        self._do_file_copy(entry.src, entry.dest)
         self._chown_chmod(entry.dest, entry.mode, entry.user, entry.group)
 
     def _install_directory(self, entry):
         self._maybe_make_unowned_dir(os.path.dirname(entry.dest))
-        self._do_mkdir(entry.dest, entry.mode, entry.user, entry.group)
+        self._do_mkdir(entry.dest, entry.mode)
         self._chown_chmod(entry.dest, entry.mode, entry.user, entry.group)
 
     def _install_treeartifact(self, entry):
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -101,17 +101,48 @@ class NativeInstaller(object):
         self._do_mkdir(entry.dest, entry.mode)
         self._chown_chmod(entry.dest, entry.mode, entry.user, entry.group)
 
+    def _install_treeartifact_file(self, entry, src, dst):
+        self._do_file_copy(src, dst)
+        self._chown_chmod(dst, entry.mode, entry.user, entry.group)
+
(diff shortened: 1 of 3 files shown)


pkg/private/install.py.tpl
```diff
@@ -103,15 +103,27 @@ class NativeInstaller(object):
 
     def _install_treeartifact(self, entry):
         logging.info("COPYTREE %s <- %s/**", entry.dest, entry.src)
-        raise NotImplementedError("treeartifact installation not yet supported")
-        for root, dirs, files in os.walk(entry.src):
-            relative_installdir = os.path.join(entry.dest, root)
+        shutil.copytree(
+            src = entry.src,
```
aiuto, line 107: python style is no space around = for arg passing.

pkg/private/install.py.tpl
```diff
@@ -103,15 +103,27 @@ class NativeInstaller(object):
...
+            dst = entry.dest,
+            symlinks = True,
+            copy_function=lambda s, d: self._do_file_copy(s, d, entry.mode, entry.user, entry.group),
+            dirs_exist_ok = True,
```
aiuto, line 111: I think you should add `ignore_dangling_symlinks=True`.

pkg/private/install.py.tpl
```diff
@@ -103,15 +103,27 @@ class NativeInstaller(object):
...
+        )
+
+        dir_mode = None
+        if entry.mode:
+            dir_mode = int(entry.mode, 8)
+            dir_mode |= 0o111  # +x, allow listing files below directories
```
aiuto, line 117: If we are going to make this choice for them, we might as well make it 555.
Tree artifacts are almost always a bundle of readable things.

pkg/private/install.py.tpl
```diff
@@ -103,15 +103,27 @@ class NativeInstaller(object):
...
     def _install_treeartifact(self, entry):
         logging.info("COPYTREE %s <- %s/**", entry.dest, entry.src)
-        raise NotImplementedError("treeartifact installation not yet supported")
-        for root, dirs, files in os.walk(entry.src):
-            relative_installdir = os.path.join(entry.dest, root)
+        shutil.copytree(
+            src = entry.src,
+            dst = entry.dest,
+            symlinks = True,
+            copy_function=lambda s, d: self._do_file_copy(s, d, entry.mode, entry.user, entry.group),
```
aiuto, line 110: Is entry.mode the right thing? That would be the top level dir of the artifact, not a typical file within it.

pkg/private/install.py.tpl
```diff
@@ -103,15 +103,27 @@ class NativeInstaller(object):
...
+            dirs_exist_ok = True,
+        )
+
+        dir_mode = None
```
aiuto, line 114: Can you add a comment that this is cleaning up the owners of the directories because copytree will create them as the owner running install.

That *is* what this is doing, right?

pkg/private/install.py.tpl
jacky8hyf: I would assume that `entry` is for the WHOLE tree artifact, so -- for all directories & files within it. I can check what pkg_tar/pkg_zip does and try aligning the behavior too.

jacky8hyf:
I did a few changes:

- Clean up some internal functions (done in a separate patch and rebased the patch on top of it)
- Added tests
- Addressed the comments.

As such, I did a force push for a fresh review. Please take a look; thanks!

I didn't handle symlinks within the tree artifact, though.

pkg/private/install.py.tpl
jacky8hyf: As far as I can see, the "intermediate directories" within the tree artifact is just set with a hard-coded 755:

https://github.com/bazelbuild/rules_pkg/blob/main/pkg/private/tar/build_tar.py#L285

So I did the same thing here.

For the files within the TreeArtifact, I did use entry.mode. But I haven't double check that's the same behavior in pkg_tar/pkg_zip, so I'll leave this open until I confirm this.

jacky8hyf: I have checked that pkg_tar uses the same behavior here (that is, the files inside the TreeArtifact has the same attribute as the TreeArtifact's `attributes` as well). As such, I think it is doing the right thing now, and I have also added tests to mandate this behavior.

jacky8hyf:
Hello, any updates? Could you please take a look at this pull request? Thanks!

cgrindel (commented):
LGTM. I will let @aiuto look before merging.

PR #888, opened 2024-08-29
Make compresslevel configurable for .tar.gz

gdh1995:
## What's the problem

The current tar packaging process is not efficient enough, as evidenced by:
- The default gzip compression level "6" can take 200% more time than compression level "1", but the size is only reduced by 10%-20%.
 - For tar packages that reach gigabyte levels and are only transmitted and used within an organization’s intranet, packing speed might be more critical than size.

## How to solve

Add `compresslevel: str` which can be `"" (auto, 6) | "0" | "1" | ... | "9"` (this is extracted from #887).

### DIFF
--- a/pkg/private/tar/build_tar.py
+++ b/pkg/private/tar/build_tar.py
@@ -42,7 +42,8 @@ class TarFile(object):
   class DebError(Exception):
     pass
 
-  def __init__(self, output, directory, compression, compressor, create_parents, allow_dups_from_deps, default_mtime):
+  def __init__(self, output, directory, compression, compressor, create_parents,
+               allow_dups_from_deps, default_mtime, compression_level):
     # Directory prefix on all output paths
     d = directory.strip('/')
     self.directory = (d + '/') if d else None
@@ -52,6 +53,7 @@ class TarFile(object):
     self.default_mtime = default_mtime
     self.create_parents = create_parents
     self.allow_dups_from_deps = allow_dups_from_deps
+    self.compression_level = compression_level
 
   def __enter__(self):
     self.tarfile = tar_writer.TarFileWriter(
@@ -60,7 +62,8 @@ class TarFile(object):
         self.compressor,
         self.create_parents,
         self.allow_dups_from_deps,
-        default_mtime=self.default_mtime)
+        default_mtime=self.default_mtime,
+        compression_level=self.compression_level)
     return self
 
   def __exit__(self, t, v, traceback):
@@ -397,6 +400,9 @@ def main():
   parser.add_argument('--allow_dups_from_deps',
                       action='store_true',
                       help='')
+  parser.add_argument(
+      '--compression_level', default=-1,
+      help='Specify the numeric compress level in gzip mode; may be 0-9 or -1 (default to 6).')
   options = parser.parse_args()
 
   # Parse modes arguments
@@ -440,6 +446,10 @@ def main():
   if options.stamp_from:
     default_mtime = build_info.get_timestamp(options.stamp_from)
 
+  compression_level = -1
+  if options.compression_level:
+    compression_level = int(options.compression_level)
+
   # Add objects to the tar file
   with TarFile(
       options.output,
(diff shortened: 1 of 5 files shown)


pkg/private/tar/tar_writer.py
```diff
@@ -86,10 +87,11 @@ def __init__(self,
     else:
       mode = 'w:'
       if compression in ['tgz', 'gz']:
+        compresslevel = int(compresslevel) if compresslevel or compresslevel == 0 else 6
```
cgrindel, line 90: Why do you need the `compresslevel == 0`?

pkg/private/tar/tar_writer.py
```diff
@@ -49,7 +49,8 @@ def __init__(self,
                create_parents=False,
                allow_dups_from_deps=True,
                default_mtime=None,
-               preserve_tar_mtimes=True):
+               preserve_tar_mtimes=True,
+               compresslevel=None):
```
cgrindel, line 53: To keep things consistent, can you rename to `compress_level`?

pkg/private/tar/tar.bzl
```diff
@@ -272,6 +274,10 @@ pkg_tar_impl = rule(
         ),
         "create_parents": attr.bool(default = True),
         "allow_duplicates_from_deps": attr.bool(default = False),
+        "compresslevel": attr.string(
```
cgrindel, line 277: Rename to `compress_level`.

pkg/private/tar/build_tar.py
```diff
@@ -42,7 +42,8 @@ class TarFile(object):
   class DebError(Exception):
     pass
 
-  def __init__(self, output, directory, compression, compressor, create_parents, allow_dups_from_deps, default_mtime):
+  def __init__(self, output, directory, compression, compressor, create_parents,
+               allow_dups_from_deps, default_mtime, compresslevel=None):
```
cgrindel, line 46: Rename to `compress_level`.

cgrindel (commented):
Is there any way that we can test that the compress level value is being passed along to the command line correctly? Alternatively, can we add a test that at least uses the new attribute?

pkg/private/tar/tar_writer.py
gdh1995: Well it's indeed unnecessary... I'll remove it.

pkg/private/tar/tar_writer.py
gdh1995: The current word `compresslevel` is from `gzip.GzipFile(compresslevel=...)`. So how do you think about the two names `compresslevel` and `compress_level`?

cgrindel: Per [the Starlark style guide](https://bazel.build/rules/bzl-style#naming-convention), the preferred name should be `compress_level`. I understand that there may be an existing API that does not conform. However, I suggest that does not mean that this non-conforming API dictate what we implement today.

@aiuto Any thoughts?

gdh1995:
Hello, I added a simple test named `test_compress_level` in tests/tar/pkg_tar_test.py, and its expected sizes are captured on my Python3.8 environment - fortunately it looks like working well in CI jobs.

cgrindel (approved):
LGTM.  I will wait to merge so that @aiuto can review.

pkg/private/tar/tar.bzl
```diff
@@ -272,6 +274,10 @@ pkg_tar_impl = rule(
         ),
         "create_parents": attr.bool(default = True),
         "allow_duplicates_from_deps": attr.bool(default = False),
+        "compress_level": attr.string(
+            doc = """Specify the numeric compress level in gzip mode; may be 0-9 or empty (6).""",
```
aiuto, line 278: Why as string rather than an int? The docstring is given in terms of an int

gdh1995: I made it string, so that its default value can be not `0` but an empty string. Do I need to make the attr to be `6` by default?

aiuto: People usually make the default -1 for ints and then figure out what do to.  So you can make the behavior work well either way.

The real question is what is easier for the user. I can make a case for either. If this were low, medium, high, then a string for sure.  zip uses 0-9, so an int seems right. Another compression type might use different level meanings.

If you had wanted to use an int, but only did it for the default, then you can make it that. But I suspect this won't matter because most people will use the default. We can even change it in the macro wrapper so that people could use either. :-)

I'll merge it either way. Just LMK

gdh1995: Hello, today I noticed the name `compression_level` in ZIP-related code, so I updated this PR and now the attribute is `compression_level: int = -1`.

aiuto (approved):

PR #891, opened 2024-09-13
pkg_install: modify log levels & print destdir by default.

### LINKED ISSUE #889: Allow pkg_install() to print destdir by default
If destdir is a relative path, and/or if destdir is provided in the BUILD file, then sometimes the user does not know where the files are installed to. For a better UI, I would like to see this:

```
$ bazel run //package:my_install
Files are installed to /path/to/installed/dir!
```

There are a few difficulties / issues that has not been settled.

1. The default log level is warning. However, the log `Files are installed to /path/to/installed/dir!` appears to be best categorized as an INFO log. 
2. If the user specified destdir in the command line, we might not want to print anything for brevity.

----

Hence, I propose the following changes:

- The default log level for install.py.tpl should be INFO. All existing logs (CHDIR, CHOWN, etc.) are moved to `logging.debug()` level. Then make the log level adjust with the following:
  ```
      if args.quiet:
        logging.getLogger().setLevel(logging.WARNING)
    elif loudness == 0:
        logging.getLogger().setLevel(logging.INFO)
    elif loudness == 1:
        logging.getLogger().setLevel(logging.DEBUG)
  ```
  That is: 
  - if -q only print warning/error (there are none of these logs now)
  - by default print info
  - if -v print debug level logs
  With these two changes, the current command-line API is preserved (-v prints the logs)

- If destdir is relative, OR if --destdir is not specified in the command line, then print `Files are installed to /path/to/installed/dir!` at the INFO level. This covers the following cases:
  - If the user did not provide `--destdir` in the command line, we fall back to the value in BUILD file, so we print the value to tell the user where the file goes.
  - If the user did provide a `--diestdir` but it is relative, we also print the value because the user might not know that it is below BUILD_WORKSPACE_DIRECTORY.

Please let me know if these changes are acceptable. If so I'll make a pull request for this. Thanks!

jacky8hyf:
After this change, log levels are defined by the following:
* -q: only print errors (not even warnings)
* (default): print errors, warnings, infos
* -v: print above & also debug logs.

There are no errors and warnings logs at this moment. The existing verbose logs about file operations are categorized as debug logs, so they are only printed when -v is set.

By default (if no flag is set), the destdir is printed as an INFO level log.

Also improve the logging format to be similar to bazel's output, e.g.

    INFO: Installing to /tmp/inst

Fixes #889

### DIFF
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -65,29 +65,29 @@ class NativeInstaller(object):
 
     def _chown_chmod(self, dest, mode, user, group):
         if mode:
-            logging.info("CHMOD %s %s", mode, dest)
+            logging.debug("CHMOD %s %s", mode, dest)
             os.chmod(dest, int(mode, 8))
         if user or group:
             # Ownership can only be changed by sufficiently
             # privileged users.
             # TODO(nacl): This does not support windows
             if hasattr(os, "getuid") and os.getuid() == 0:
-                logging.info("CHOWN %s:%s %s", user, group, dest)
+                logging.debug("CHOWN %s:%s %s", user, group, dest)
                 shutil.chown(dest, user, group)
 
     def _do_file_copy(self, src, dest):
-        logging.info("COPY %s <- %s", dest, src)
+        logging.debug("COPY %s <- %s", dest, src)
         shutil.copyfile(src, dest)
 
     def _do_mkdir(self, dirname, mode):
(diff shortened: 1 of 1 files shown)


jacky8hyf:
> There are no errors and warnings logs at this moment.

I missed that there are some critical logs. However this change does not change when critical logs are printed (they are always printed).

cgrindel (approved):
LGTM.

aiuto (approved):

PR #894, opened 2024-09-20
install: Add --wipe_destdir option

### LINKED ISSUE #893: Add pkg_install(wipe_destdir: bool)
I want to request a feature to add the following:

```
pkg_install(
    wipe_destdir = True,
)
```

Default value is false. If true, pkg_install() script will wipe the given destdir before installing the files.

This is particularly useful if we want the resulting destdir to be always strictly containing the given list of files, but not any old files. For example, I have a pkg_install() rule that generates stardoc and put it in our source tree. If the list of generated file names changes, I need to manually delete the old documentation files now. But with this attribute, I can set it to True so I don't have to manually delete old documentation. The example is illustrated here: https://r.android.com/c/kernel/build/+/3273727/2/kleaf/docs/BUILD.bazel

@aiuto if you think this is a useful feature, I can start preparing a pull request; or please let me know if this does not fit in pkg_install(). Thanks!

jacky8hyf:
If specified, wipe destination directory before installing.

Fixes #893.

### DIFF
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -53,10 +53,12 @@ RUNFILE_PREFIX = os.path.join(os.getenv("RUNFILES_DIR"), WORKSPACE_NAME) if os.g
 #
 # See also https://bugs.python.org/issue37157.
 class NativeInstaller(object):
-    def __init__(self, default_user=None, default_group=None, destdir=None):
+    def __init__(self, default_user=None, default_group=None, destdir=None,
+                 wipe_destdir=False):
         self.default_user = default_user
         self.default_group = default_group
         self.destdir = destdir
+        self.wipe_destdir = wipe_destdir
         self.entries = []
 
     # Logger helper method, may not be necessary or desired
@@ -171,6 +173,9 @@ class NativeInstaller(object):
 
     def do_the_thing(self):
         logging.info("Installing to %s", self.destdir)
+        if self.wipe_destdir:
+            logging.debug("RM %s", self.destdir)
+            shutil.rmtree(self.destdir, ignore_errors=True)
         for entry in self.entries:
             if entry.type == manifest.ENTRY_IS_FILE:
                 self._install_file(entry)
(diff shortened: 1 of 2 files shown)


pkg/private/install.py.tpl
```diff
@@ -171,6 +173,9 @@ class NativeInstaller(object):
 
     def do_the_thing(self):
         logging.info("Installing to %s", self.destdir)
+        if self.wipe_destdir:
+            logging.debug("RM %s", self.destdir)
+            shutil.rmtree(self.destdir, ignore_errors=True)
```
aiuto, line 178: This will delete the target directory, which will mean that we recreate it, so the perms will change. Unless someone has set that explicitly, this may be unexpected if people had hand created a landing spot. Not that this is wrong, but maybe it deservces some comment in the docs.

pkg/private/install.py.tpl
```diff
@@ -236,6 +241,8 @@ def main(args):
                              f"Relative paths are interpreted against "
                              f"BUILD_WORKSPACE_DIRECTORY "
                              f"({os.getenv('BUILD_WORKSPACE_DIRECTORY')})")
+    parser.add_argument("--wipe_destdir", action="store_true", default=False,
+                        help="Wipe destdir before installing.")
```
aiuto, line 245: Maybe "delete destdir tree before installing".

aiuto (approved):
Your call on the doc wording.

pkg/private/install.py.tpl
jacky8hyf: Done

pkg/private/install.py.tpl
jacky8hyf: Done

jacky8hyf:
The test error seemed unrelated to my patch. How do I re-run the test?

aiuto:
> The test error seemed unrelated to my patch. How do I re-run the test?

They are unrelated errors. Something change in Bazel at head that is breaking a few of these. It's not that Bazel is really broken either, it's that we have tests for compatibility against WORKSPACE too, and Bazel is rapidly deprecating support for that.

I will force merge.

PR #900, opened 2024-11-15
`tar_writer.py`: address duplicate dir warning regression

### LINKED ISSUE #899: `pkg_tar` incorrectly reports duplicate **file** for duplicate directories
I believe since #850 landed we get a lot of duplicate file warnings for duplicate directories

I dont think duplicate directories should emit any warnings as this is just a case of merging directories with different content

Illedran:
It looks like #850 removed the check below, resulting in warning being printed on duplicate _directories_ in addition to files:
https://github.com/bazelbuild/rules_pkg/pull/850/files#diff-9a83a2a408d04cf1937eb6d1eac00438774072bfe46a19082c7f8729d3ac4a27L133-L135

Fixes #899

### DIFF
--- a/pkg/private/tar/tar_writer.py
+++ b/pkg/private/tar/tar_writer.py
@@ -129,7 +129,7 @@ class TarFileWriter(object):
       # Enforce the ending / for directories so we correctly deduplicate.
       if not info.name.endswith('/'):
         info.name += '/'
-    if not self.allow_dups_from_deps and self._have_added(info.name):
+    elif not self.allow_dups_from_deps and self._have_added(info.name):
         print('Duplicate file in archive: %s, '
               'picking first occurrence' % info.name)
         return


pkg/private/tar/tar_writer.py
```diff
@@ -129,7 +129,7 @@ def _addfile(self, info, fileobj=None):
       # Enforce the ending / for directories so we correctly deduplicate.
       if not info.name.endswith('/'):
         info.name += '/'
-    if not self.allow_dups_from_deps and self._have_added(info.name):
+    elif not self.allow_dups_from_deps and self._have_added(info.name):
```
cgrindel, line 132: Please add a comment here explaining why it needs to be this way. It may help prevent future contributors from changing the logic inadvertently.

cgrindel (approved):

cgrindel:
@aiuto Some CI tests are failing. However, the failures do not appear to be related to this change. Do we know if `main` is green? Are there periodic tests that run on `main`?

Illedran:
Hold off for a second on this one - I've just tested it locally in my project to build tars for image layers to be used w/ [`rules_oci`](https://github.com/bazel-contrib/rules_oci) and it results in the tarball being unable to be `docker load`-ed:
```
1b7379da5 add: Loading layer [=>                                                 ]  32.77kB/837.1kB
duplicates of file paths not supported
```
I suspect that this is an issue specifically in the consumer/docker since duplicate directories should be valid - normal GNU `tar` handles archives produced from this PR just fine. While this does fix the warning, I'll take another look at it

cgrindel (approved):

PR #903, opened 2024-11-27
Delete deb_packages directory

njlr:
The notice in this folder is misleading since `rules_pkg` now has better maintained `.deb` rules than the two suggested alternatives

### DIFF
--- a/deb_packages/README.md
+++ b/deb_packages/README.md
@@ -1,5 +1,0 @@
-# This module is deprecated
-
-The work here has forked to two other places.  Please look at both to see which is better maintained and meets your needs best.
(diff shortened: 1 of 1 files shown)


tonyaiuto:
The intent might be different. deb_packages was not about producing .deb packages, Those repositories were about using them, so I think the premise of the PR is wrong.
OTOH, neither of those repositories have had an update in 5 years, so maybe it's time to drop the refs.

tonyaiuto (approved):

aiuto:
The tests are failing for unrelated reasons, so I'll merge anyway.

Issue #896, opened 2024-09-28
Test are broken at head


aiuto:
A few tests are broken at head because of a rules_python fix vs. recent Bazel changes.

The root cause might be https://buildkite.com/bazel/rules-pkg/builds/3359

The error indicates this is somewhere within the external repo tests, which build a WORKSPACE dynamically and test that we work when uses as a repo, rather than just testing behavior standalone.
```
(21:33:14) ERROR: no such package '@@[unknown repo 'mappings_test_external_repo' requested from @@]//pkg': The repository '@@[unknown repo 'mappings_test_external_repo' requested from @@]' could not be resolved: No repository visible as '@mappings_test_external_repo' from main repository. Was the repository introduced in WORKSPACE? The WORKSPACE file is disabled by default in Bazel 8 (late 2024) and will be removed in Bazel 9 (late 2025), please migrate to Bzlmod. See https://bazel.build/external/migration.
```

Thanks to anyone who wants to jump on this and fix it.  Making them pass is probalby easy.  Deciding if we should test this or just assume we work as a repository is a harder decision.

PR #911, opened 2025-01-10
fix: Fix broken head

hofbi:
Closes https://github.com/bazelbuild/rules_pkg/issues/896

Bump all dependencies and set `--enable_workspace` since this is now built with Bazel 8

I plan to come up with a few follow up PRs, for now this was my "minimal changes" attempt to make head green.


doc_build/BUILD
hofbi, line 1: These changes are required because of the empty glob flag enabled by bazel 8

tonyaiuto: ack

tonyaiuto (approved):

PR #913, opened 2025-01-13
Pre commit autoupdate and format files

hofbi:
Autoupdate pre-commit hooks and run the formatter to bring all files to a consistent autoformatted state. In a follow up PR, I would add the pre-commit check as a GH action, similar to https://github.com/bazel-contrib/rules-template/pull/130.


hofbi:
@aiuto What do you think about this one?

aiuto (approved):

PR #910, opened 2025-01-08
Fix check_must_contain_regex in verify_archive.bzl

krakeusz:
Closes https://github.com/bazelbuild/rules_pkg/issues/909

PR #915, opened 2025-01-13
Fix tests broken after https://github.com/bazelbuild/rules_pkg/pull/910.

aiuto:
The test was incorrect, but was passing becuase the test harness was broken. pr/910 fixed the harness, but did not follow up and fix the broken tests.


aiuto:
@krakeusz

PR #917, opened 2025-01-13
Allow some common rule args in verify_archive_test

### LINKED ISSUE #916: verify_archive_test doesn't allow common test attrs
`verify_archive_test` is useful but it can't be passed typical test attributes like `size`, `testonly`, or even `visibility`.
Ie. the following fails with `bazel test`:
```
verify_archive_test(
    name = "verify_product_shared_only_package_contents",
    size = "small",
    testonly = True,
    must_contain_regex = [
        ".*\\.so",
    ],
    target = ":pkg_tar_shared_only",
)
```
```
                def verify_archive_test(
Error: verify_archive_test() got unexpected keyword arguments: size, testonly, visibility
```

krakeusz:
Closes #916

This is my proposal for allowing common test arguments for `verify_archive_test`.

### DIFF
--- a/pkg/verify_archive.bzl
+++ b/pkg/verify_archive.bzl
@@ -23,6 +23,21 @@ The execution time is O(# expected patterns * size of archive).
 
 load("@rules_python//python:defs.bzl", "py_test")
 
+# Attribute names common to all build rules. See https://bazel.build/reference/be/common-definitions
+COMMON_BUILD_ATTR_NAMES = [
+    "tags",
+    "target_compatible_with",
+    "testonly",
+    "visibility",
+]
+
+# Attribute names common to all test rules. See https://bazel.build/reference/be/common-definitions#common-attributes-tests
+COMMON_TEST_ATTR_NAMES = COMMON_BUILD_ATTR_NAMES + [
+    "size",
+    "timeout",
+    "flaky",
+]
+
 def _gen_verify_archive_test_main_impl(ctx):
     ctx.actions.expand_template(
         template = ctx.file._template,
@@ -95,7 +110,6 @@ def verify_archive_test(
         must_not_contain_regex = None,
         min_size = 1,
         max_size = -1,
-        tags = None,
         verify_links = None,
         **kwargs):
     """Tests that an archive contains specific file patterns.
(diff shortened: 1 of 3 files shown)


krakeusz:
I'll keep it as a draft until CI passes :)

hofbi:
> I'll keep it as a draft until CI passes :)

passed

krakeusz:
I'm making it a draft for a moment - the `kwargs` need to be forwarded to `_gen_verify_archive_test_main` as well. I'll add a test as well.

pkg/verify_archive.bzl
```diff
@@ -23,6 +23,21 @@ The execution time is O(# expected patterns * size of archive).
...
+    "target_compatible_with",
+    "testonly",
+    "visibility",
+]
+
+# Attribute names common to all test rules. See https://bazel.build/reference/be/common-definitions#common-attributes-tests
+COMMON_TEST_ATTR_NAMES = COMMON_BUILD_ATTR_NAMES + [
+    "size",
+    "timeout",
+    "flaky",
```
aiuto, line 38: Part of my wants to never allow flaky tests. :-)

pkg/verify_archive.bzl
```diff
@@ -23,6 +23,21 @@ The execution time is O(# expected patterns * size of archive).
 
 load("@rules_python//python:defs.bzl", "py_test")
 
+# Attribute names common to all build rules. See https://bazel.build/reference/be/common-definitions
+COMMON_BUILD_ATTR_NAMES = [
+    "tags",
```
aiuto, line 28: I think tags is more useful in test args.
My reasoning:
- I don't have a need to control if I build the test harness
- I do want to be able to skip tests based on a tag pattern.

We could put it in both, but just having it in test is good enough. If we skip the test, the binary may not need to be built.

krakeusz: These attributes will be used both in the test harness `_gen_verify_archive_test_main` and the `py_test` itself, see line 35.
My reasoning is: if Bazel supports it for build targets, why not allowing the user to do the same. What's more, if I add `tags = ['manual']`, then there's no need to build anything needed to run the test, just for performance reasons. But it's not a strong opinion.

Maybe I should rename it to `COMMON_ATTR_NAMES` to avoid confusion?

aiuto (approved):

PR #923, opened 2025-01-20
fix: Avoid else after return in mappings bzl

hofbi:
Many linters suggest not to use else-after-return. This is a drive by fix for the mapping module + a few whitespace fixes. I will make a follow up PR to streamline the whitespaces in general and avoid too many unrelated autofixes by saving in an IDE.


cgrindel (approved):

cgrindel (approved):
LGTM. I will not merge it yet so that @aiuto can review.

PR #925, opened 2025-01-22
fix: symlinks in pkg_tar

jsun-splunk:
User defined symlinks was not working in pkg_tar as _check_dest() in pkg_files did not account for src=None.


aiuto:
I'm not going to be able to answer any questions until 1/27 because of a
family issue.

jsun-splunk:
Looks like this was a false positive and I was trying to generate symlinks to a non-existent path. Closing for now.

Issue #929, opened 2025-01-28
Support packaging raw symlinks from declare_symlink


rickeylev:
Over in rules_python, declare_symlink() is used to create symlinks for efficiency and correctness reasons. Such artifacts are "raw" symlinks, that is, Bazel will keep them exactly as-is and not try to transform them, as happens when using declare_file().

Unfortunately, pkg_tar gives an error when it sees these artifacts. What happens is it thinks the file is a regular file, so tries to read it. Since its a relative symlink, it either is dangling (in the context of pkg_tar), or will read the actual file (instead of creating a symlink), neither of which is desirable behavior.

Thankfully, starting in Bazel 8, File.is_symlink can be used to detect these artifacts. This allows the manifest that pkg_tar generates, and the build_tar program that uses it, to handle them correctly.

The two basic changes are:
1. When File.is_symlink is true, write a manifest entry with entry_type=raw_symlink (pkg_files.bzl)
2. Modify build_tar.py to handle entry_type=raw_symlink: read os.readlink() from the src file. Then store it as a symlink. This preserves the original symlink.

I prototyped this a bit and it seems to work. I'd reference my branch, but its in a very hacky and broken state right now. The above gets the gist though

rickeylev:
I sent out https://github.com/bazelbuild/rules_pkg/pull/930 to demonstrate the gist of fixes necessary for pkg_tar.

Something to note is that File.is_symlink is a Bazel 8+ API. For earlier versions of Bazel, I'm not aware of a way to detect these. For earlier Bazel versions, a workaround I thought of is to have rules_pkg check for OutputGroupInfo.files_that_are_symlinks. A separate provider could be used, too. The reason I suggest a more structural-based approach is to avoid the issue of language rules having to export multiple packaging-implementation-specific providers or vice versa.

PR #930, opened 2025-01-29
feat: support raw symlinks from declare_symlink()

rickeylev:
Using the `ctx.actions.declare_symlink()` API, rules can create explicit symlinks to arbitrary
paths, which Bazel will keep as symlinks. The packaging logic treats these
as regular files, so tries to read through them and store the content they point to. This
is incorrect for two reasons: they are supposed to be left as symlinks and may be dangling
symlinks from the perspective of packaging.

To fix, introduce a new type of file entry: `raw_symlink`. These entries mean the `src` is
a symlink whose stored path should be stored into the destination verbatim, also as a symlink.
Such artifacts are identified using `File.is_symlink`, a Bazel 8+ API.

Work towards https://github.com/bazelbuild/rules_pkg/issues/929

### DIFF
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -40,9 +40,9 @@ default_tests_with_rpm: &default_tests_with_rpm
 
 win_tests: &win_tests
   test_flags:
-    - "--test_tag_filters=-nowindows"
+    - "--test_tag_filters=-nowindows,only-this-test"
   build_flags:
-    - "--build_tag_filters=-nowindows"
+    - "--build_tag_filters=-nowindows,only-this-test"
   test_targets:
     - "//tests/tar:raw_symlinks_test"
     - "-//.."
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -41,8 +41,10 @@ default_tests_with_rpm: &default_tests_with_rpm
 win_tests: &win_tests
   test_flags:
     - "--test_tag_filters=-nowindows,only-this-test"
+    - "--windows_enable_symlinks"
   build_flags:
     - "--build_tag_filters=-nowindows,only-this-test"
+    - "--windows_enable_symlinks"
   test_targets:
     - "//tests/tar:raw_symlinks_test"
     ## - "//pkg/..."
--- a/.bazelci/tests.yml
+++ b/.bazelci/tests.yml
@@ -40,26 +40,25 @@ default_tests_with_rpm: &default_tests_with_rpm
 
 win_tests: &win_tests
   test_flags:
-    - "--test_tag_filters=-nowindows,only-this-test"
+    - "--test_tag_filters=-nowindows"
   build_flags:
-    - "--build_tag_filters=-nowindows,only-this-test"
+    - "--build_tag_filters=-nowindows"
   test_targets:
-    - "//tests/tar:raw_symlinks_test"
-    ## - "//pkg/..."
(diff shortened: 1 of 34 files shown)


cgrindel (commented):
Can we add a test for this?

rickeylev:
Ok, it's been awhile since I could get time to come back to this!

I've added a basic test. PTAL.

rickeylev:
OK, PTAL.

I fixed the buildifier presubmit failure.

The other failure looks unrelated -- I get the same error at clean head. It's specific to the rolling version of bazel and something about workspace support.

rickeylev:
Actually, I partially take that back -- The LTS windows test is failing. Some quick searches indicates windows symlink support should work fine with the values the test is using (relative symlinks that point nowhere).

Diagnosing and fixing this will be a bit hard, but I'll see what I can do. If it comes to it, can we simply mark this test to skip windows?

meteorcloudy:
https://bazel.build/configure/windows#symlink
Do you have `--windows_enable_symlinks` specified as a startup flag? Otherwise a junction might be created instead of a symlink, which doesn't support relative path.

See https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/windows/WindowsFileSystem.java;l=83-92

rickeylev:
I tried adding `--windows_enable_symlinks`, but that didn't fix it.

So far, all I can tell is that, when tar building occurs, and python calls `os.readlink()` to figure out the value contained in the symlink, it gets back an absolute path e.g. `\\?\C:\b\3ctu2esx\execroot\_main\bazel-out\x64_windows-fastbuild\bin\tests\tar\link1`. That is the correct resolved path (`foo_link1 -> ./link1`), but yeah, we don't want the resolved path.

I can't tell which of these cases is happening:
1. Bazel is writing an absolute path
2. Windows is silently transforming it to an absolute path at write
3. Windows is silently transforming it to an absolute path at read
4. Python is transforming it to an absolute path

I won't have direct access to a windows machine until this evening, so I'll have to wait until then to check some of these ideas.

But overall, this isn't sounding promising. I don't think we can work around (1) or (2). A work around for (3) and (4) might be possible by using e.g. ctypes and dropping into lower-level APIs.

rickeylev:
After testing on my personal windows machine, I'm thinking its (1: Bazel writing an absolute path). I can use `mklink` to create relative symlinks, and python's `os.readlink()` returns the relative path correctly. When I use an absolute path, it returns a `\\?\...` looking string, the same as what I'm observing.

rickeylev:
Aha, here we go. known issue in Bazel specific to windows: https://github.com/bazelbuild/bazel/pull/24213

meteorcloudy:
OK, I guess let's disable the test on Windows for now

rickeylev:
Tests fixed. PTAL

rickeylev:
bump -- this has been sitting for awhile; is there an ETA on when someone can review?

cgrindel:
> bump -- this has been sitting for awhile; is there an ETA on when someone can review?

In the future, if you hit the "Re-request review" button, that will let the reviewer who has left comments look at the PR again. Looking now

cgrindel (approved):
LGTM. @aiuto Did you want to look before we merge?

rickeylev:
Oh, i had not noticed that thingy before. Done, for at least my own practice :). Thanks!

tonyaiuto (approved):

PR #934, opened 2025-02-13
rpm: fix sub-RPM post_scriptlet and add %postun support

wade-arista:
The existing implementation of `%post` neglected to actually include the contents of the `%post` scriptlet and simply added a `%post` section to the spec file without the body, e.g.

    %post sub

    %files sub
    ...

With this change, I've also added support for `%postun` and included the scriptlet support in the `test_golden_sub_rpm_contents` test case.


cgrindel (approved):
LGTM.

cgrindel:
@aiuto It appears the CI failures are unrelated to the PR. How do you usually handle this?

aiuto (approved):

PR #941, opened 2025-02-25
Fix installing code-signed binaries on macOS

luispadron:
When running a `pkg_install` on a code-signed binary on macOS, that is already executing from a previous install... I was getting `Killed: 9` when attempting to run the newly installed executable.

This can be reproduced semi-consistently with something like:

```sh
# Run an install & execute the installed binary, background immediately, then do it again.
bazel run //:install --destdir bin && ./bin/foo &; \
bazel run //:install --destdir bin && ./bin/foo
```

This is apparently a long-standing "quirk" on macOS when executing code-signed binaries. 

>This code is incorrect because it modifies the command-line tool’s executable file in place. macOS caches information about the code’s signature in the kernel. It doesn’t flush that cache when you modify the file’s contents. Modifying the file in place yields a mismatch between the file’s contents and the in-kernel cache, which can cause a hard-to-reproduce code-signing crash the next time you run the tool.

> To update a file that contains signed code without risking this crash, write the updated code to a temporary file and replace the existing file with that temporary one

> Using a temporary file eliminates the risk of a code-signing crash because the in-kernel cache is associated with the old file, which remains unmodified. The new file gets its own in-kernel cache, built from the contents of that new file.

Source: https://developer.apple.com/documentation/security/updating-mac-software

Maybe we could do this for just macOS but this should be supported on all platforms.

### DIFF
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -23,6 +23,7 @@ import os
 import pathlib
 import shutil
 import sys
+import tempfile
 
 from pkg.private import manifest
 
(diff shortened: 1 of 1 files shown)


cgrindel (approved):

aiuto (approved):

aiuto:
An install tests seems to be failing .https://buildkite.com/bazel/rules-pkg/builds/3583#01953b88-df12-4579-afb0-a560668ecd50

There are also a bunch of other broken tests which I have a fix for in #943 
Let's get that in and see if this PR is actually clean.

aiuto:
Can you rebase against head and push again. I did some things to fix the baseline continuous build, so I want to see the tests stay clean.

luispadron:
@aiuto thanks! There was a legit error on linux due to calling `replace` on different file-systems. I fixed this with `dir` in the tempfile call.

luispadron:
@aiuto any plans to create a release soon?

aiuto:
1.1 came out last week.

PR #942, opened 2025-02-26
Generic `debuginfo` RPM platform support

nicbadiu:
This change splits the `debuginfo` RPM support into `fedora` and `centos` and bundles `almalinux` into the `centos` category.

This facilitates future support as more versions are being added as opposed to hardcoding every individual OS name and version combination.

The fedora and centos specific logic remains unchanged.


kellyma2:
LGTM

aiuto (approved):
Nice.  Even someone who does not know all the nuances might be able to maintain this going forward.

PR #945, opened 2025-03-12
Overridable `debuginfo` type

nicbadiu:
This change enables the override of the auto-detected `/etc/os-release`-based `debuginfo_type` attribute of the `build_repo_for_rpmbuild_toolchain` repository rule and allows the auto-detection by means of a new `default` option.

This also enables the use of` none` as an override option.

This change also includes a regeneration of the documentation, which seems to have trailed recent code changes. But there are no documentation changes related to this change as `build_repo_for_rpmbuild_toolchain` repository rule is not documented (yet).

### DIFF
--- a/toolchains/rpm/rpmbuild_configure.bzl
+++ b/toolchains/rpm/rpmbuild_configure.bzl
@@ -82,9 +82,9 @@ def _build_repo_for_rpmbuild_toolchain_impl(rctx):
         fail("debuginfo_type must be one of", DEBUGINFO_VALID_VALUES)
 
     debuginfo_type = rctx.attr.debuginfo_type
-    if debuginfo_type == DEBUGINFO_TYPE_AUTODETECT:
-        rctx.watch(RELEASE_PATH)
+    if debuginfo_type == DEBUGINFO_TYPE_AUTODETECT:   
         if rctx.path(RELEASE_PATH).exists:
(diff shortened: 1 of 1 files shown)


tonyaiuto:
@kellyma2  Do you want to take a look at it.

@nicbadiu  Can you provide a description of the API change in the PR description?
Are you able to update the docs too?  See the `How To` of doc_build/BUILD

toolchains/rpm/rpmbuild_configure.bzl
```diff
@@ -57,6 +57,9 @@ def _parse_release_info(release_info):
 
     return os_name, os_version
 
+DEBUGINFO_TYPE_DEFAULT = "default"
```
kellyma2, line 60: maybe we should call this DEBUGINFO_TYPE_AUTODETECT?

kellyma2: I'm good with it.

tonyaiuto:
It looks like you did this
```
+        rctx.watch(RELEASE_PATH)
+        if rctx.path(RELEASE_PATH).exists:
```
instead of
```
+        if rctx.path(RELEASE_PATH).exists:
+           rctx.watch(RELEASE_PATH)
```
I patch the PR.

PR #949, opened 2025-04-06
pkg_tar: Fix #948: Don't add duplicate directory members to tar file when a symlink with the same path exists

### LINKED ISSUE #948: pkg_tar: Combining tar files where directories shadow symlinks results in duplicate tar file members with the same name
I'm using `pkg_tar` to combine tars produced with `rules_distroless` and some other tars into an OCI image layer. However, upon loading the layer using Docker or Podman, I get the following error:

`Error processing tar file(duplicates of file paths not supported)`.

After analyzing the resulting tar archive, I have determined that the issue was caused by the presence of a symbolic link from `/lib` to `/usr/lib` in one of the packages and a `/lib` directory in one of the other packages. It seems that `pkg_tar` incorrectly deduplicates tar file entries when one of the the member's type is a directory and the other member's type is not (e.g. it is a symlink).

The behavour I would expect in this case is that if the symlink precedes the directory, then the member should remain a symlink. If the directory precedes the symlink, the directory should remain a directory and a warning should be displayed.

nickguletskii:
I've modified the deduplication logic to account for the fact that directories must not have the same path as other types of files. Previously, the paths `/lib` and `/lib/` would be treated differently.

Two test cases are included: one for symlinks preceding directories, and one for directories preceding symlinks. In both cases, the first member is preserved and the second member with the same path is discarded as a duplicate.

### DIFF
--- a/pkg/private/tar/tar_writer.py
+++ b/pkg/private/tar/tar_writer.py
@@ -35,6 +35,18 @@ PORTABLE_MTIME = 946684800  # 2000-01-01 00:00:00.000 UTC
 
 _DEBUG_VERBOSITY = 0
 
+TARFILE_MEMBER_TYPE_TO_STR = {
+    b"0": "REGTYPE",
+    b"\0": "AREGTYPE",
+    b"1": "LNKTYPE",
+    b"2": "SYMTYPE",
+    b"3": "CHRTYPE",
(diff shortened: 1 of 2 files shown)


cgrindel (approved):
LGTM.

Issue #947, opened 2025-04-01
Support make var expansion in pkg_tar


dmarquez-splunk:
We have a use case where we need to use variables from the toolchain's make vars in order to interpolate paths passed to remap_paths attribute in pkg_tar. Currently this is unsupported and results in the files not being found. Is this something that we'd be open to supporting?

tonyaiuto:
We don't build on feature branches. 
Clone the repo
push to your repo
then do the pull request across forks

tonyaiuto:
> We have a use case where we need to use variables from the toolchain's make vars in order to interpolate paths passed to remap_paths attribute in pkg_tar. Currently this is unsupported and results in the files not being found. Is this something that we'd be open to supporting?

It seems like something people would want to do, but how do you propose making it work. There is not "the" toolchain, there are toolchains all over the place. I may make a package that includes targets built under transitions, so I might have toolchains for both C++ and  Rust, support differint machine architectures.

Why don't you propose a design for the feature first.

dmarquez-splunk:
I didn't suspect it'd require too much effort. I've been working with [these](https://github.com/bazelbuild/rules_pkg/pull/950) changes on my local which seem to satisfy our use case. Please have a look at your earliest convenience

aiuto:
Please don't just say it would not be too much effort. Please explain how
the new feature would work so people can evaluate it.

dmarquez-splunk:
Oh of course. I by no means meant that as a reason to not write up a proposal but rather an informal estimate for the work required (which can be completely wrong). I will spend some time on this early next week.

PR #950, opened 2025-04-11
Feature: Enable variable expansion for remap_paths in pkg_tar (#947)

dmarquez-splunk:
These changes allow for variable expansion in remap_paths in the pkg_tar rule

### DIFF
--- a/pkg/private/tar/tar.bzl
+++ b/pkg/private/tar/tar.bzl
@@ -25,6 +25,7 @@ load(
     "write_manifest",
 )
 load("//pkg/private:util.bzl", "setup_output_files", "substitute_package_variables")
+load("@aspect_bazel_lib//lib:expand_make_vars.bzl", "expand_variables")
 
 # TODO(aiuto): Figure  out how to get this from the python toolchain.
 # See check for lzma in archive.py for a hint at a method.
@@ -40,9 +41,12 @@ SUPPORTED_TAR_COMPRESSIONS = (
 _DEFAULT_MTIME = -1
 _stamp_condition = Label("//pkg/private:private_stamp_detect")
 
-def _remap(remap_paths, path):
+def _remap(ctx, path):
     """If path starts with a key in remap_paths, rewrite it."""
-    for prefix, replacement in remap_paths.items():
+    for prefix, replacement in ctx.attr.remap_paths.items():
+        prefix = expand_variables(ctx, prefix)
+        replacement = expand_variables(ctx, replacement)
+
         if path.startswith(prefix):
             return replacement + path[len(prefix):]
     return path
@@ -120,7 +124,7 @@ def _pkg_tar_impl(ctx):
     # Now we begin processing the files.
     path_mapper = None
     if ctx.attr.remap_paths:
-        path_mapper = lambda path: _remap(ctx.attr.remap_paths, path)
+        path_mapper = lambda path: _remap(ctx, path)
 
     mapping_context = create_mapping_context_from_ctx(
         ctx,
--- a/tests/tar/BUILD
+++ b/tests/tar/BUILD
@@ -423,6 +423,16 @@ pkg_tar(
     },
 )
 
+pkg_tar(
+    name = "test-remap-paths-variables",
+    srcs = [
+        ":tree_artifact_to_rename",
+    ],
+    remap_paths = {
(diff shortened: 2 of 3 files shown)


pkg/private/tar/tar.bzl
```diff
@@ -40,9 +41,12 @@ SUPPORTED_TAR_COMPRESSIONS = (
 _DEFAULT_MTIME = -1
 _stamp_condition = Label("//pkg/private:private_stamp_detect")
 
-def _remap(remap_paths, path):
+def _remap(ctx, path):
```
cgrindel, line 44: Please revert this change and pass the `remap_paths` to this function. We should try to limit the number of locations where `ctx` is passed.

cgrindel (changes requested):

dmarquez-splunk:
Running into a doc build issue that I'm not certain how to resolve:

```
/workdir/doc_build/BUILD:65:12: in deps attribute of starlark_doc_extract rule //doc_build:pkg_tar_gen.extract: missing
bzl_library targets for Starlark module(s) @aspect_bazel_lib//lib:expand_make_vars.bzl. Since this rule was created by the 
macro 'stardoc', the error might have been caused by the macro implementation
```

Any guidance would be greatly appreciated! cc: @cgrindel

aiuto:
You have to pull the dependency on the expand vars stuff into doc_build/BUILD.

But, I really don't want to take dependencies on other packages from here.
What particular value is the other expand_variables doing over pkg/private/tar/util.bzl%substitute_pkg_variables?

dmarquez-splunk:
> You have to pull the dependency on the expand vars stuff into doc_build/BUILD.
> 
> But, I really don't want to take dependencies on other packages from here. What particular value is the other expand_variables doing over pkg/private/tar/util.bzl%substitute_pkg_variables?

I haven't experimented with the substitute_pkg_variables functions yet. But expand_make_variables in aspect bazel supports expanding pre-defined make variables, pre-defined gen rule variables, and custom make variables. 

In our use case we have MSVC and other windows SDK variables defined within our windows toolchain that need to be interpolated in order for our builds to package proper versions of windows SDK libraries (like UCRT and MSVCRT). expand_variables seems to support it out of the box which at first glance seemed like the best path. Open to hearing alternative solutions if you think substitute package variables is a viable alternative!

dmarquez-splunk:
Can I get another review on this MR please? @aiuto @cgrindel

tests/mappings/external_repo/MODULE.bazel
```diff
@@ -26,6 +26,7 @@ local_path_override(
 bazel_dep(name = "rules_license", version = "1.0.0")
 bazel_dep(name = "rules_python", version = "1.0.0")
 bazel_dep(name = "bazel_skylib", version = "1.7.1")
+bazel_dep(name = "aspect_bazel_lib", version = "2.14.0")
```
cgrindel, line 29: Why did this need to be added to this test workspace? It is already included in `rules_pkg`.

cgrindel (commented):
I am ambivalent about pulling in aspect_bazel_lib. I will leave that decision to @aiuto. If we do keep it, please see my PR comment. Something does not seem right.

tests/mappings/external_repo/MODULE.bazel
dmarquez-splunk: Let me try and build the tests without this. I assumed it was required

dmarquez-splunk: Removed

aiuto:
Sorry for slow responses. I am in the middle of my last few weeks at Google. Not leaving Bazel community though. Will have time to look at this starting May 5.

dmarquez-splunk:
> I am ambivalent about pulling in aspect_bazel_lib. I will leave that decision to @aiuto. If we do keep it, please see my PR comment. Something does not seem right.

@aiuto how do you feel about this? I am interested to hear your thoughts before I proceed writing tests & fixing the build

dmarquez-splunk:
> I am ambivalent about pulling in aspect_bazel_lib. I will leave that decision to @aiuto. If we do keep it, please see my PR comment. Something does not seem right.

I am opening to writing a new implementation similar to expand_variables in the aspect_bazel_lib. My first inclination was to just use a pre-existing implementation but can look into writing an alternative if thats what we think is best. Just looking for a little direction here

PR #952, opened 2025-04-28
pkg_rpm: propagate make_rpm PATH to rpmbuild

wade-arista:
Depending on the specific rpmbuild macros in use, we might need to have a valid PATH set.

I saw a specific issue where one of the macros calls a "find-debuginfo" script that uses "which" (/usr/bin/which) to locate gdb. Because "which" is not a shell command, it doesn't know about the default PATH provided by bash, and so it doesn't see _any_ path.

If we just propagate PATH we can avoid this confusing class of errors. It seems reasonable that the PATH used by `rpmbuild` is inherited from bazel in this case as it is more generally during rule execution.

### DIFF
--- a/pkg/make_rpm.py
+++ b/pkg/make_rpm.py
@@ -420,6 +420,7 @@ class RpmBuilder(object):
     env = {
         'LANG': 'C',
         'RPM_BUILD_ROOT': buildroot,
+        'PATH': os.getenv( 'PATH' ),
     }
 
     if self.source_date_epoch is not None:


pkg/make_rpm.py
```diff
@@ -420,6 +420,7 @@ def CallRpmBuild(self, dirname, rpmbuild_args, debuginfo_type):
     env = {
         'LANG': 'C',
         'RPM_BUILD_ROOT': buildroot,
+        'PATH': os.getenv( 'PATH' ),
```
cgrindel, line 423: Inheriting the PATH is not a good idea. Can you point out where you saw `which` being called? I see [a Which() function](https://github.com/bazelbuild/rules_pkg/blob/106429fdbc6bc83b2f1da0c481c6bde8f456fa4f/pkg/make_rpm.py#L127-L142) that is implemented in Python.

wade-arista: It was, I thought at least, from `find-debuginfo.sh` via [`debugedit`](https://sourceware.org/debugedit/), but I looked at the upstream version and they're using the shell builtins, instead (`type`).

I suppose that since we're relying on non-hermetic tools, we are best to leave it up to bash to pick its own PATH defaults and just ensure we don't need PATH by only using builtins since the distro will (should) have consistent bash PATH default and installed tools locations, presumably.

I'll close the PR.

PR #957, opened 2025-06-11
Use rules_shell in tests

keith:
This is required for testing with bazel @ HEAD


cgrindel (approved):

tonyaiuto (approved):
Centos is still failing on RPM, but that is a different problem.

PR #958, opened 2025-06-11
Use DefaultInfo directly

keith:
Required for compatibility with https://github.com/bazelbuild/bazel/issues/20183


keith:
CI failing until https://github.com/bazelbuild/rules_pkg/pull/957

cgrindel (approved):

PR #960, opened 2025-06-17
pkg_deb: add support for md5sums control file

### LINKED ISSUE #959: pkg_deb: support md5sums file
The [md5sums control file](https://manpages.debian.org/testing/dpkg-dev/deb-md5sums.5.en.html) is optional, but recommended and missing it will result in a [lintian](https://lintian.debian.org/tags/no-md5sums-control-file.html) error.  There are also some tools like older versions of aptly which required it.  It's also handy for efficiently comparing the content of two packages without fully extracting them, e.g.
```
diff <(dpkg-deb -I pkg_1_all.deb md5sums) <(dpkg-deb -I pkg_2_all.deb md5sums)
```

Yes, I know it's 2025 and we shouldn't be using md5 for anything any more (it's not just insecure, it's also _slow_ compared to better hashing functions).  But that's what's in the standard.

I'm not proposing that `rules_pkg` should necessarily be responsible for _creating_ the file, but it should be allowable to supply one to the `pkg_deb` rule to support cases where one is desired or required.

adam-azarchs:
This adds a new, optional attribute to the pkg_deb rule allowing a user to supply an md5sums file.

Closes #959

### DIFF
--- a/pkg/private/deb/deb.bzl
+++ b/pkg/private/deb/deb.bzl
@@ -76,6 +76,9 @@ def _pkg_deb_impl(ctx):
     if ctx.attr.triggers:
         args.append("--triggers=@" + ctx.file.triggers.path)
         files.append(ctx.file.triggers)
+    if ctx.attr.md5sums:
+        args.append("--md5sums=@" + ctx.file.md5sums.path)
+        files.append(ctx.file.md5sums)
 
     # Conffiles can be specified by a file or a string list
     if ctx.attr.conffiles_file:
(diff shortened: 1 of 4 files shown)


tonyaiuto (approved):

adam-azarchs:
Looks like I'll need some changes to make the test not fail on windows.

aiuto (approved):

tonyaiuto:
Can you rebase and push again to whack the tests.

PR #973, opened 2025-08-08
Document pkg_tar.extension's compression options.

### LINKED ISSUE #972: Document compression support for pkg_tar
I didn't know this, but `pkg_tar` supports some compressions natively:

https://github.com/bazelbuild/rules_pkg/blob/8c7c2cf7f1742a1fb199d0164b800d2939fa9fef/pkg/private/tar/tar.bzl#L81-L95

by setting the `extension` attribute:

```starlark
pkg_tar(
    name = "package_tester_bin_tar",
    srcs = [":package_tester_bin"],
    include_runfiles = True,
    extension="gz",  # this bit!
)
```

Let's get that documented.

dougthor42:
Fixes #972

### DIFF
--- a/docs/latest.md
+++ b/docs/latest.md
@@ -370,7 +370,7 @@ pkg_tar(<a href="#pkg_tar-name">name</a>, <a href="#pkg_tar-deps">deps</a>, <a h
 | <a id="pkg_tar-create_parents"></a>create_parents |  -   | Boolean | optional |  `True`  |
 | <a id="pkg_tar-empty_dirs"></a>empty_dirs |  -   | List of strings | optional |  `[]`  |
 | <a id="pkg_tar-empty_files"></a>empty_files |  -   | List of strings | optional |  `[]`  |
-| <a id="pkg_tar-extension"></a>extension |  -   | String | optional |  `"tar"`  |
+| <a id="pkg_tar-extension"></a>extension | The extension of the generated file. If `"gz"`, `"bz2"`, or `"xz"`, the tarball will also be compressed using that tool, and is mutually exclusive with `compressor`. Note that `xz` may not be supported based on the Python toolchain. | String | optional |  `"tar"`  |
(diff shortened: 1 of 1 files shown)


docs/latest.md
```diff
@@ -370,7 +370,7 @@ pkg_tar(<a href="#pkg_tar-name">name</a>, <a href="#pkg_tar-deps">deps</a>, <a h
 | <a id="pkg_tar-create_parents"></a>create_parents |  -   | Boolean | optional |  `True`  |
 | <a id="pkg_tar-empty_dirs"></a>empty_dirs |  -   | List of strings | optional |  `[]`  |
 | <a id="pkg_tar-empty_files"></a>empty_files |  -   | List of strings | optional |  `[]`  |
-| <a id="pkg_tar-extension"></a>extension |  -   | String | optional |  `"tar"`  |
+| <a id="pkg_tar-extension"></a>extension | The extension of the generated file. If `"gz"`, `"bz2"`, or `"xz"`, the tarball will also be compressed using that tool, and is mutually exclusive with `compressor`. Note that `xz` may not be supported based on the Python toolchain. | String | optional |  `"tar"`  |
```
dougthor42, line 373: > Note that `xz` may not be supported based on the Python toolchain.

I'm not 100% sure this is true. The source has a TODO to confirm `xz` availabilty, but it's been there for > 3 years now and `HAS_XZ_SUPPORT` is True.

https://github.com/bazelbuild/rules_pkg/blob/8c7c2cf7f1742a1fb199d0164b800d2939fa9fef/pkg/private/tar/tar.bzl#L29-L31

Maybe I just shouldn't mention that `xz` might not be supported?

aiuto: This is fine.

aiuto: But the problem is that it won't work if you change it here. 
It will get lost on the next release update when the file is generated. You need to change it in the rule souce.

dougthor42: Ah, thanks for letting me know. I've updated the starlark source instead, please LMK if there's something else I'm supposed to do.

PR #974, opened 2025-08-14
📦 Conditionally preserve file permissions when archiving through pkg_tar

ronodnn:
If the new `preserve_mtime` flag is enabled, set mtime by extracting it from the file's mtime attribute. The `preserve_mtime` flag is disabled by default.

This is related to Issue https://github.com/bazelbuild/bazel/issues/1299


cgrindel (approved):
Small formatting comment. Otherwise, LGTM.

cgrindel (approved):
LGTM. I will not merge so that @aiuto can review.

PR #979, opened 2025-10-06
Implement pkg_install(symlink)

aiuto:
- A first attempt at pgk_install of symlinks. It was pretty obvious code once I realized I needed os.lchmod()
- Minor lint fixes

### DIFF
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -102,8 +102,11 @@ class NativeInstaller(object):
         logging.debug("SYMLINK %s <- %s %s %s %s", target, link_name, mode, user, group)
         os.symlink(target, link_name)
         if mode:
-            logging.debug("CHMOD %s %s", mode, link_name)
-            os.lchmod(link_name, int(mode, 8))
+            if hasattr(os, "lchmod"):
+                 logging.debug("CHMOD %s %s", mode, link_name)
+                 os.lchmod(link_name, int(mode, 8))
+            else:
+                 logging.debug("CHMOD-NOT AVAILABLE %s %s", mode, link_name)
         if user or group:
             # Ownership can only be changed by sufficiently
             # privileged users.
--- a/pkg/private/install.py.tpl
+++ b/pkg/private/install.py.tpl
@@ -99,7 +99,18 @@ class NativeInstaller(object):
         os.makedirs(dirname, int(mode, 8), exist_ok=True)
 
     def _do_symlink(self, target, link_name, mode, user, group):
-        raise NotImplementedError("symlinking not yet supported")
+        logging.debug("SYMLINK %s <- %s %s %s %s", target, link_name, mode, user, group)
+        os.symlink(target, link_name)
+        if mode:
+            logging.debug("CHMOD %s %s", mode, link_name)
+            os.lchmod(link_name, int(mode, 8))
+        if user or group:
+            # Ownership can only be changed by sufficiently
+            # privileged users.
+            # TODO(nacl): This does not support windows
+            if hasattr(os, "lchown") and os.getuid() == 0:
+                logging.debug("CHOWN %s:%s %s", user, group, link_name)
+                os.lchown(link_name, user, group)
 
     def _maybe_make_unowned_dir(self, path):
         logging.debug("MKDIR (unowned) %s", path)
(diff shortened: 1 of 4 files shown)


pkg/private/install.py.tpl
```diff
@@ -99,7 +99,21 @@ class NativeInstaller(object):
         os.makedirs(dirname, int(mode, 8), exist_ok=True)
 
     def _do_symlink(self, target, link_name, mode, user, group):
-        raise NotImplementedError("symlinking not yet supported")
+        logging.debug("SYMLINK %s <- %s %s %s %s", target, link_name, mode, user, group)
```
apsaltis-ddog, line 102: Issue: This is not like the others.  The (l)chmod and (l)chown are covered below, and the thing created is on the right side of the arrow, unlike with the copy operation.

Suggestion: Simplify to:

```suggestion
        logging.debug("SYMLINK %s <- %s", link_name, target)
```

pkg/private/install.py.tpl
```diff
@@ -99,7 +99,21 @@ class NativeInstaller(object):
...
+                 os.lchmod(link_name, int(mode, 8))
+            else:
+                 logging.debug("CHMOD-NOT AVAILABLE %s %s", mode, link_name)
+        if user or group:
+            # Ownership can only be changed by sufficiently
+            # privileged users.
+            # TODO(nacl): This does not support windows
+            if hasattr(os, "lchown") and os.getuid() == 0:
+                logging.debug("CHOWN %s:%s %s", user, group, link_name)
+                os.lchown(link_name, user, group)
```
apsaltis-ddog, line 116: Thought: is there a way that we could reuse `_chown_chmod` here?  Calling out `lchown` and `lchmod` is equivalent to passing `follow_symlinks=False` ([docs](https://docs.python.org/3/library/os.html#os.lchmod).

Suggestion: Add the argument to `_chown_chmod`, and catch `NotImplementedError` if `follow_symlinks` isn't supported ([docs](https://docs.python.org/3/library/os.html#files-and-directories))?  

I don't feel particularly strongly about this.

tests/install/test.py
```diff
@@ -29,9 +29,11 @@ class PkgInstallTestBase(unittest.TestCase):
     def setUpClass(cls):
         cls.runfiles = runfiles.Create()
         # Somewhat of an implementation detail, but it works.  I think.
-        manifest_file = cls.runfiles.Rlocation("rules_pkg/tests/install/test_installer_install_script-install-manifest.json")
+        manifest_file = cls.runfiles.Rlocation(
+            "rules_pkg/tests/install/test_installer_install_script-install-manifest.json"
+        )
 
-        with open(manifest_file, 'r') as fh:
+        with open(manifest_file, "r") as fh:
```
apsaltis-ddog, line 36: Nit: mention reformatting in PR message.

tests/install/BUILD
```diff
@@ -97,3 +99,19 @@ pkg_files(
...
+    name = "fake_so_in_lib",
+    srcs = [":fake.so.1.2.3"],
+    prefix = "/lib",
+)
+
+pkg_mklink(
+    name = "fake_lib_link",
+    link_name = "/lib/fake.so.1",
+    target = "fake.so.1.2.3",
+)
```
apsaltis-ddog, line 117: Thought: `ldconfig -n` ([docs](https://man7.org/linux/man-pages/man8/ldconfig.8.html)) would accomplish the same thing, at least on Linux.

apsaltis-ddog (commented):
(I am @nacl) Some suggestions below, otherwise LGTM.

Nit: Issue for this?  Probably not needed in this case.

tests/install/test.py
```diff
@@ -214,12 +230,15 @@ def test_wipe(self):
...
-            self.runfiles.Rlocation("rules_pkg/tests/install/test_installer"),
-            "--destdir", self.installdir,
-            "--wipe_destdir",
-        ],
-                              env=self.runfiles.EnvVars())
+        subprocess.check_call(
+            [
+                self.runfiles.Rlocation("rules_pkg/tests/install/test_installer"),
+                "--destdir",
+                self.installdir,
```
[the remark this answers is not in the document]
aiuto: SIgh... I'll put it back.   I agree with you, but the precommit hook I am using is using the Datadog agent standards.
I'll revert that.

tests/install/BUILD
aiuto: I need macos too and then a strange install.

Issue #918, opened 2025-01-15
Improving BCR release automation


hofbi:
Since I a few fixes were merged, I would be interested in a new release on BCR.

To make this even easier for maintainers in the future, I would like to advertise the latest updates in https://github.com/bazel-contrib/rules-template. Specifically https://github.com/bazel-contrib/rules-template/blob/ec5d715ad584cc8a36cff3d7c9aaae6545e0bebd/.github/workflows/tag.yaml#L13-L18 now allows to automatically create tags on a regular basis to trigger an new BCR release. I can already see https://github.com/bazelbuild/rules_pkg/tree/main/.bcr, hence there might be only a few updates to the most recent versions required.

hofbi:
@aiuto Who might be the best person to drive a new release here?

hofbi:
Updated the title since https://github.com/bazelbuild/rules_pkg/releases/tag/1.1.0 was created

PR #991, opened 2025-12-02
Create pre-release tests that run as a github workflow.

aiuto:
Create a linux only subset of the tests that run as a gitlab workflow.

- This will become the smoke test before the release workflow.
- The buildkite workflows are still needed for the cross platform testing.
- There are very strange failures with Bazel latest (10.0.0). They are related to JDK runtimes, which we do not use for any of the tests. So I did not try with the latest release yet. I'll come back to that in regular CI.
- Testing with bazel 7 showed some features are not backwards compatible, I adjusted test cases to allow bazel 7 jobs to skip those tests.

#918

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,1 @@
-# .bazelrc
+# Intentionally blank


.bazelrc
```diff
@@ -1 +1 @@
-build --enable_workspace
+# .bazelrc
```
cgrindel: Just double checking that this shoould be empty. If so, we might want to change the comment to `# Intentionally blank` or just delete the file.

aiuto: I went with intentionally blank.
That way if we add things temporarily and remove things, we will have a history of that file, rather than disjoint appearancs and deletions.

PR #1015, opened 2026-02-04
[contrib] tools for inspecting package images

tonyaiuto:
First cut at toolkit to do package inspection. This is for contrib right now, because I want to see more flight time with it before supporting it as a core feature.

I started talking about this in https://github.com/bazelbuild/rules_pkg/discussions/1001
The goal is to create a set of tools that can read through various archive types so that it is easy to build task specific validators.  Examples are
- compare the .deb file from the nightly to the one in my PR.  What files changed size or got added
- compare a .rpm and .deb of the same product, verify that the only diff is in the /etc/init scripts
- validate that the results of pkg_install produces the same tree of files you get from building a .deb and installing for the same input sources (the original motivation for this tool set)

The rpm and cpio decoder are clean room implementations, that I wrote from the spec.
After that, I used a coding agent to generate most of the readers. It's grunt work and I was happy to farm it out. 

I started a test suite in https://github.com/bazelbuild/rules_pkg/pull/1019.

We can do review the API and general utility here. When that is shaken out and landed, I'll update the tests to account for changes here.  In the interim, I'm using this daily at work, so that gives me a pretty good set of real world tests.

README.md has more details.


tonyaiuto:
The tests seem to be failing on unrelated issues, unable to load rules_java.

PR #1025, opened 2026-02-12
Ensure runfiles.{,root_}symlinks makes it into tarballs

EdSchouten:
Right now we only add files. This is problematic when creating a tarball of an sh_binary(). Starting with Bazel 9, sh_binary() is no longer provided by Bazel itself. In that case rules_shell will declare its own copy of runfiles.bash, putting a symlink at the old location in the runfiles directory.


cgrindel (approved):
LGTM. @aiuto I will not merge this PR so that you can take a look as well.

PR #1026, opened 2026-02-13
Unwrap pkg_deb rule to enable inherit_attrs

mxpv:
I was trying to wrap `pkg_deb` into a macro to express some common functionality. One convenient thing that Bazel has is `inherit_attrs = pkg_deb`, which lets you redefine certain attributes and reuse everything else. However, in the current implementation it's not possible to use, because `pkg_deb` is being wrapped in a legacy macro.

```
Error in macro: in call to macro(), parameter 'inherit_attrs' got value of type 'function', want 'rule, macro, string, or NoneType'
```

This PR exposes the rule directly and moves the logic to default out into the rule implementation.

### DIFF
--- a/pkg/private/deb/deb.bzl
+++ b/pkg/private/deb/deb.bzl
@@ -29,9 +29,12 @@ def _pkg_deb_impl(ctx):
             ctx.attr.architecture,
         )
 
+    out_file = ctx.outputs.out or ctx.actions.declare_file(ctx.attr.name + ".deb")
+
     outputs, output_file, output_name = setup_output_files(
         ctx,
         package_file_name = package_file_name,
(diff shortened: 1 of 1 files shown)


mxpv:
> Or, update this to a symbolic macro. Although that may break bazel 7 compatibilty. I'ld have to check.

@tonyaiuto Symbolic macros were [introduced in Bazel 8](https://bazel.build/rules/macro-tutorial), so that won't work on 7.

> What if we just exposed the attribute list as something you can use yourself, and wrap the implementaion so you can reuse it.

Exposing the attribute list is another option. Any particular reason not to expose the rule directly? That seems more natural and can be used directly with `inherit_attrs`.

Issue #996, opened 2025-12-12
Move providers out of rules_pkg


malt3:
`PackageFilesInfo` is a very useful construct that many codebases use to define how exactly files should be packaged.
I would like for `rules_img` to consume this provider as well to build layers (in the `image_layer` rule).
However, I don't want to depend on the whole transitive closure of rules_pkg (including rules_python).
Could we agree to move providers to a neutral Bazel module that doesn't have any dependencies?
This way, producers and consumers can easily depend on the provider module.

Tagging @aiuto for comment.

PR #1030, opened 2026-02-19
Expose api for extendability

FaBrand:
This alloes wo construct the mapping structures of rules_pkg internal representation and to extend based on pkg_* rules

### DIFF
--- a/doc_build/BUILD
+++ b/doc_build/BUILD
@@ -55,6 +55,7 @@ ORDER = [
     ("PackageFilesInfo", "//providers:providers.bzl"),
     ("PackageSymlinkInfo", "//providers:providers.bzl"),
     ("pkg_install", None),
+    ("package_content", None),
     ("legacy_pkg_rpm", None),
 ]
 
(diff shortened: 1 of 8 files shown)


cgrindel (commented):
I am not sure if feel qualified to approve the API change. @tonyaiuto do you want to elevate these to public scope?

cgrindel:
I am deferring to @tonyaiuto on how to proceed.

aiuto:
Sorry for the slow response.
I don't quite understand why it duplicates so much of process_src in package_content.bzl.
And create_mapping_context_from_ctx seems to have a behavior that should be documented make usage clear.

It looks like it does things like
strip_prefix is used as is, if provided, otherwise it defaults to the value from the original ctx.
But it feels like that should be up to the person writing their own rule wrapper.

Do you have a real world example of the need for that particular API?

FaBrand:
@aiuto i moved the functions into a public package after the suggestion in: https://github.com/bazelbuild/rules_pkg/issues/996#issuecomment-3934994777

I can't share the code directly (internal policy...) but the gist is this:

1) Have a custom rule that uses pkg_* as srcs
2) Do something with it
3) Consume it in e.g. pkg_install 

Basically the "pkg-rules sandwich"

My concrete use case is:

Define a rich hirarchy of pkg_files/filegroups/etc, have a pkg_jfrog_* ruleset that generates a jf-cli upload specification from this defined structure.

Afterwards use `pkg_install` to install it into a defined location instead of having to search and upload "something"

I found myself trying to read the different Providers of rules_pkg in my rule which ended up effectively duplicating e.g. [process_src](https://github.com/bazelbuild/rules_pkg/pull/1030/changes#diff-b661e773fcd638620364d2d0a2278d133830d2df094a1b62b48e06cbf418bdd3R105-R149)

It works well when buildifiers private visibility check, but of course is always a risk and not advisable to depend on private details

Hence my suggestion to expose these functions as a publicly usable interface to enable interfacing or injecting custom rules in the pkg-dependency graph


Sry for the lengthy text^^

FaBrand:
Kind ping @aiuto

PR #1032, opened 2026-02-22
Depend on the providers via the bcr module

aiuto:
Blocked on
- having the providers in bcr
- a release notes/ readme, or something to explain how this works if you depend on the package with http_archive.

# Annoucement (draft)

We are about to introduce a structural change to rules_pkg where the providers are in a distinct repository.  This will allow those to be shared across several projects in a way that facilitates interoperability without having to build every extension inside rules_pkg.

- `rules_pkg_providers` will be a new release name and a BCR module. It will start at 0.0.1 and quickly bump to 1.0.0 after an initial shake-out period.
- `rules_pkg` will start to depend on `rules_pkg_providers` in `MODULE.bazel`.

**Note that this is a breaking change for WORKSPACE users**

For that reason, we will bump the major version number of rules_pkg.
- `rules_pkg` will have a final 1.x release this week.
- `rules_pkg` 2.0.0 will depend on `rules_pkg_providers`.

#996


peakschris:
I'm trying to update rules_img to support rules_pkg_providers, but I'm blocked on rules_pkg not depending on rules_pkg_providers. Could this change be made and released?

PR #1053, opened 2026-04-16
rpm_pfg: allow multiple tree artifacts at the same destination

chouquette:
When two pkg_files targets both map a configure_make tree artifact (e.g. an include/ directory) to the same destination prefix, pkg_rpm was raising a hard error even though the file contents don't overlap. The duplicate-destination check was firing unconditionally before the src.is_directory branch, so two packages installing headers to embedded/include/ (from different subdirectories) would conflict.

Move the dest_check_map conflict check inside the non-directory branch so that tree artifacts sharing a destination are treated as a merge rather than a conflict. Actual file-level collisions within those trees will be caught at install time by the underlying tooling.

### DIFF
--- a/pkg/private/pkg_files.bzl
+++ b/pkg/private/pkg_files.bzl
@@ -144,6 +144,12 @@ def _check_dest(content_map, dest, src, origin, allow_duplicates_with_different_
     if old_entry.src == src or old_entry.origin == origin:
         return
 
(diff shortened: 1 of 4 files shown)


cgrindel (commented):
Looks reasonable. Is there a way to add a test for this?

aiuto (approved):

PR #1068, opened 2026-05-27
pkg_deb: Add `installed_size` attribute

### LINKED ISSUE #1067: pkg_deb missing `installed_size` attribute
The `Installed-Size` Debian control field is already supported by `make_deb.py` (the `--installed_size` flag is auto-generated from `DEBIAN_FIELDS`), but there is no attribute on the `pkg_deb` Bazel rule to pass a value through.

dandersson:
Closes #1067 

The `Installed-Size` Debian control field was already supported by `make_deb.py` (the `--installed_size` flag is auto-generated from `DEBIAN_FIELDS`), but there was no attribute on the `pkg_deb` Bazel rule to pass a value through.

## Changes

- Add an `installed_size` `attr.string` to `pkg_deb_impl` (`pkg/private/deb/deb.bzl`), following the same pattern as other optional string control-field attributes (`homepage`, `license`, `priority`, etc.). For simplicity I opted to make it a string and not an integer, even though technically it always should be integer-valued, to be in line with other existing attributes.
- Emit `--installed_size` to the `make_deb` action args when the attribute is set.

This PR also includes a preparatory commit that fixes the stale version number in `docs/latest.md` (still said `1.2.0` after the bump to `1.3.0` in `c62b199`).

## Testing

```
bazel test //tests/deb/...
```

All `deb` tests pass.

(I get a test failure for `//tests/tar:pkg_tar_test`, but I get that on current `main` (22b54a3) as well, and I can't find a revision where this test passes for me at all.)

## Thoughts on auto-calculating size

One could imagine automatically calculating this value from the `data` parameter using the Python `tarfile` module, but apart from being a more invasive change it could be backwards-incompatible for broken archives, and in particular would be a practical problem for zstd compression which doesn't get Python standard library support until Python 3.14.

Implementing it as just another named argument seems safe. Calculating the size can be done at the call site instead.


docs/latest.md
cgrindel, line 1: @aiuto I presume this updated to 1.30 due to the pre-release?

cgrindel (approved):
LGTM. I'll leave it open for @aiuto to take a look as well.

aiuto (approved):

docs/latest.md
aiuto: Whoops. Lost this.

PR #1073, opened 2026-07-11
[TarFile]: normalize also concatenated paths (#1072)

### LINKED ISSUE #1072: [TarFile]: normalize_path does not normalize when --directory is present
### Problem description

When using `--directory` (`self.directory`) with `TarFile`, the following segment produces non-normalized path components:
```python
# pkg/private/tar/build_tar.py
    # This prevents a potential problem for users with both a prefix_dir and
    # symlinks that also repeat the prefix_dir. The old behavior was that we
    # would get just the symlink path. Now we are prefixing with the prefix,
    # so you get the file in the wrong place.
    # We silently de-dup that. If people come up with a real use case for
    # the /a/b/a/b/rest... output we can start an issue and come up with a
    # solution at that time.
    if self.directory and not dest.startswith(self.directory):
      dest = self.directory + dest
```

We are using `--directory`, and so I was seeing paths like this:
```console
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/__init__.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/_cmp.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/_typing_compat.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/_version_info.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/converters.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/exceptions.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/filters.pyi
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/py.typed
runtime/_main/../_main~ps~av_pip_attrs_3_10/pypi__attrs/attr/setters.pyi
...
```
Note that these collapse into `runtime/_main~ps~av_pip_attrs_3_10/pypi__attrs/...`.

This may normally not be a problem, but becomes an issue when piping the `tar` file into `mksquashfs`:
```bash
Parallel mksquashfs: Using 32 processors
Creating 4.0 filesystem on bazel-out/k8-fastbuild/<redacted>.squashfs, block size 131072.
[=======================|                                                                                                                                                                       ]  1911/15662  12%
FATAL ERROR: Error: Tar pathname can't have '.' or '..' in it
```

### What to do

Call the existing `normpath()` on the concatenated path.

grrtrr:
The result of `normalize_path()` may be a non-normalized path, containing e.g. `../`, as a result of prepending `self.directory`.

Avoid the problem by using `normpath()` on the result. Resolves #1072.


cgrindel (approved):

aiuto:
This makes one of the tar tests fail `AssertionError: 'loremipsum.txt' != './loremipsum.txt'`
So there is some subtle different behavior.  We need to understand how that might change existing workflows.

What if you just normalized --directory?

grrtrr:
> This makes one of the tar tests fail `AssertionError: 'loremipsum.txt' != './loremipsum.txt'` So there is some subtle different behavior. We need to understand how that might change existing workflows.
> 
> What if you just normalized --directory?

I don't think the test failures are related. Normalizing `--directory` might work, doing this inside the function seemed to make the most sense, consistent with the earlier call to `normpath()`.

PR #1079, opened 2026-09-16
pkg_deb: compute Installed-Size automatically when unset

chouquette:
The `Installed-Size` control field can be set through the `installed_size` attribute, but it requires knowing the package size ahead of time and hardcoding it, which is unpractical. 

This is likely to cause the attribute to be omitted and the installed size metadata to be missing.

This PR now automatically computes the installed size when the attribute is not explicitly provided.


cgrindel (approved):
LGTM. I'll leave it open so that @aiuto can take a look.

aiuto:
I think what I would rather see is the size (and checksums?) being a side effect output the tarball maker.
That eliminates a second pass over the file.  Let's prototype that in datadog-agent, and then upstream what works.

aiuto:
This is what I have been thinking about w.r.t. improving the performance.  I should have wrote this a half a year ago, when I first realized I needed to go there.
https://github.com/bazelbuild/rules_pkg/issues/1080

